PR #12, opened 2020-10-15
Add dart textobjects.scm

GenesisTMS:
Hi, here is my try to add text-object imp for `dart`.
Would very much appreciate any feedback.

**problem with sequence**
Is it just me but sequence with multiple nodes does not work - only last count:
```
(
  (method_signature)
  (function_body)
) @function.outer
```

Thanks
P.S.: I am new to tree-sitter family and open-source contributing.

### DIFF
--- a/queries/dart/textobjects.scm
+++ b/queries/dart/textobjects.scm
@@ -1,0 +1,70 @@
+; class
+(class_definition
+  body: (class_body) @class.inner) @class.outer
+(mixin_declaration (class_body) @class.inner) @class.outer
+(enum_declaration
+  body: (enum_body) @class.inner) @class.outer
+(extension_declaration
+  body: (extension_body) @class.inner) @class.outer
+
+; function/method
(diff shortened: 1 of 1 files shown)


theHamsta:
Thanks for your contribution!

Matching multiple nodes is indeed an open issue for this repo (also needed for Lua inner functions and grouping of muliple line comments). I hope I can implement it this weekend. It only requires a small change of the code.

theHamsta (approved):

theHamsta:
First draft: https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/14 it has your suggestion for dart functions.

theHamsta:
I guess it's better when we would have a `(make-range! "function.outer" function.outer.start function.outer.end)` in nvim-treesitter instead. Then, this feature would not require a code change and other plugins could create virtual nodes by start and end, too.

PR #14, opened 2020-10-15
Fix #13: Support textobjects defined by start/end node

### LINKED ISSUE #13: Support textobjects defined by start and end node
#12 

We could define texobjects like in #12 with by their start and end nodes. E.g.:
```
(
  (method_signature) @function.outer.start
  (function_body) @function.outer.end
) 
```
or in Lua
```
(local_function . (_) . (_) . (_) @function.inner.start  (_) @function.inner.end .)
```

theHamsta:
Requires https://github.com/nvim-treesitter/nvim-treesitter/pull/553

This does not fix the "requires refactor" problem of that function.

### DIFF
--- a/lua/nvim-treesitter/textobjects/shared.lua
+++ b/lua/nvim-treesitter/textobjects/shared.lua
@@ -7,6 +7,31 @@ local ts_utils = require'nvim-treesitter.ts_utils'
 
 local M = {}
 
+local function is_in_node_range(node1, node2, line, col)
+  if not node2 then return end
+  local start_line, start_col, end_line, end_col
+  if node1 and node2 then
+   start_line, start_col = node1:start()
+   end_line, end_col = node2:end_()
+  else
+   start_line, start_col, end_line, end_col = node2:range()
+  end
+
+  if line >= start_line and line <= end_line then
+    if line == start_line and line == end_line then
+      return col >= start_col and col < end_col
+    elseif line == start_line then
+      return col >= start_col
+    elseif line == end_line then
+      return col < end_col
+    else
+      return true
+    end
+  else
+    return false
+  end
+end
+
 function M.textobject_at_point(query_string, pos, bufnr)
   bufnr =  bufnr or vim.api.nvim_get_current_buf()
   local lang = parsers.get_buf_lang(bufnr)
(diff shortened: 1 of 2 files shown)


lua/nvim-treesitter/textobjects/shared.lua
```diff
@@ -7,6 +7,31 @@ local ts_utils = require'nvim-treesitter.ts_utils'
...
+  if not node2 then return end
+  local start_line, start_col, end_line, end_col
+  if node1 and node2 then
+   start_line, start_col = node1:start()
+   end_line, end_col = node2:end_()
+  else
+   start_line, start_col, end_line, end_col = node2:range()
+  end
+
+  if line >= start_line and line <= end_line then
```
theHamsta, line 20: should upstream that `ts.is_in_node_range` could also support ranges.

lua/nvim-treesitter/textobjects/shared.lua
```diff
@@ -40,8 +65,8 @@ function M.textobject_at_point(query_string, pos, bufnr)
   local earliest_start
 
   for _, m in pairs(matches) do
-    if m.node and ts_utils.is_in_node_range(m.node, row, col) then
-      local length = ts_utils.node_length(m.node)
+    if is_in_node_range(m.start and m.start.node, m.node or (m['end'] and m['end'].node), row, col) then
+      local length = ts_utils.node_length(m['end'] and m['end'].node or m.node)
```
theHamsta, line 69: TODO: length should end-start in the start end case.

theHamsta:
Maybe we can incorporate the refactor and also select to next textobject if cursor is between textobject.

theHamsta:
I guess it's better when we would have a (make-range! "function.outer" function.outer.start function.outer.end) in nvim-treesitter instead. Then, this feature would not require a code change and other plugins could create virtual nodes by start and end, too.

GenesisTMS:
> I guess it's better when we would have a `(make-range! "function.outer" function.outer.start function.outer.end)` in nvim-treesitter instead.

I am voting for vanila approach with sequence.
ofc i don't know how much work it will cost but with basic sequence you can describe any scenario (or am i wrong?)

```
(
  (method_signature)
  (function_body)
) @function.outer
```

or am i missed the point? :D trying to catch up

theHamsta:
@GenesisTMS sorry, I posted to the wrong issue. My solution for this problem involves to much code changes and it would be better to express those node ranges in the queries.

Btw.

```scheme
((
  (method_signature) @start
   .  ; the dot is necessary since you don't want to match with the signature of *another*  function `.` means the nodes are adjacent
  (function_body) @end
) 
(make-range! "function.outer" @start @stop)) ; create a new virtual node by the range defined by start/end
```

theHamsta:
Superseded by #16

PR #17, opened 2020-10-31
queries/go: add @parameter.inner

fsouza:
This way we can use swap with Go.

There are some limitations in the query though, opening as a draft to see if I can get some input from others. With the current parser, the following function has only one parameter_declaration:

```go
func DoThing(i, j int) {}
```

This seems to be a problem with the Go parser? I'll keep playing around with it in the playground.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ EOF
 <tr>
 <td>fennel</td><td> </td> <td> </td> <td> </td> <td> </td> <td> </td> <td> </td> <td> </td> <td> </td> <td> </td> <td> </td> <td> </td> <td> </td> <td> </td> <td> </td> <td> </td> </tr>
 <tr>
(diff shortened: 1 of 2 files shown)


theHamsta:
Adding this seems to solve the problem:
```
(parameter_declaration
  (identifier)
  (identifier) @parameter.inner)
(parameter_declaration
  (identifier) @parameter.inner
  (identifier))
```

This also works for this case
```
func DoThing(j, i int, k float) {}
```
parameters are only swapped with something with same parent. So either `i, j ` is swapped or `i, j int` with `k float`

fsouza:
@theHamsta thank you very much, pushed an updated commit and this is now ready! 😁

theHamsta:
Thanks for the contribution!

PR #19, opened 2020-11-04
Add @parameter.inner for js and ts

fsouza:
Wanna do some swap with js and ts :)

### DIFF
--- a/queries/javascript/textobjects.scm
+++ b/queries/javascript/textobjects.scm
@@ -43,6 +43,18 @@
 (formal_parameters
   (identifier) @parameter.inner)
 
+; function (v = default_value)
+(formal_parameters
+  (assignment_pattern) @parameter.inner)
+
+; function ({ x }) ...
+(formal_parameters
+  (object_pattern) @paremeter.inner)
+
+; function ([ x ]) ...
+(formal_parameters
+  (array_pattern) @paremeter.inner)
+
 ;; arguments
 (arguments
   (_) @parameter.inner)
--- a/queries/javascript/textobjects.scm
+++ b/queries/javascript/textobjects.scm
@@ -45,4 +45,4 @@
 
 ;; arguments
 (arguments
-  (identifier) @parameter.inner)
+  (_) @parameter.inner)
--- a/queries/javascript/textobjects.scm
+++ b/queries/javascript/textobjects.scm
@@ -38,3 +38,11 @@
 
 ;; blocks
 (_ (statement_block) @block.inner) @block.outer
+
+;; parameters
+(formal_parameters
+  (identifier) @parameter.inner)
+
+;; arguments
+(arguments
+  (identifier) @parameter.inner)
--- a/queries/typescript/textobjects.scm
+++ b/queries/typescript/textobjects.scm
@@ -1,1 +1,5 @@
 ; inherits: javascript
+
+;; parameters
+(formal_parameters
+  (required_parameter) @parameter.inner)


theHamsta:
Thanks a lot for all these improvements!

theHamsta (approved):

fsouza:
@theHamsta thanks for the quick reviews!

queries/javascript/textobjects.scm
```diff
@@ -38,3 +38,11 @@
...
 ;; blocks
 (_ (statement_block) @block.inner) @block.outer
+
+;; parameters
+(formal_parameters
+  (identifier) @parameter.inner)
+
+;; arguments
+(arguments
+  (identifier) @parameter.inner)
```
theHamsta, line 48: Maybe also lambda parameters (from highlights):
```
;; Lambda parameter
(inferred_parameters (identifier) @parameter) ; (x,y) -> ...
(lambda_expression
    parameters: (identifier) @parameter) ; x -> ...
```

queries/typescript/textobjects.scm
```diff
@@ -1 +1,5 @@
 ; inherits: javascript
+
+;; parameters
+(formal_parameters
```
theHamsta, line 4: Is there also optional parameters when there is  required_parameter?

theHamsta:
```
    "children": {
      "multiple": true,
      "required": false,
      "types": [
        {
          "type": "optional_parameter",
          "named": true
        },
        {
          "type": "required_parameter",
          "named": true
        },
        {
          "type": "rest_parameter",
          "named": true
        }
```

theHamsta: Maybe just use any child `(_)`?

queries/javascript/textobjects.scm
fsouza: Hmm interesting, I used the playground with some anonymous functions and the current queries worked for those. I'll add the lambda parameters just in case.

While I have your attention, I noticed the current query for arguments doesn't really work with all cases, because an argument isn't always an `identifier`. Example that fails:

```typescript
    return new Promise<RespType>((resolve, reject) => {
      this.client.makeUnaryRequest<ReqType, RespType>(
        `/company.Service/${method}`,
        (req) => Buffer.from(encoder.encode(req).finish()),
        (buf) => (decoder.decode(buf) as unknown) as RespType,
        data,
        this.grpcMeta,
        { deadline },
        (err: grpc.ServiceError | null, resp: RespType | undefined) => {
          if (err) {
            reject(err)
          } else {
            resolve(resp)
          }
        }
      )
    })
```

The first parameter to makeUnaryRequest is actually `template_string`, while the second and third are `arrow_function`. Doing `(arguments (_) @parameter.inner)` fixes it (and every other case I tested), but is that too open?

theHamsta: Would be ok for me. How does that handle comments? It's also possible that the JS highlights are outdated/incomplete.

queries/typescript/textobjects.scm
fsouza: Yeah I thought so too, but in my examples I always got required_parameter. For example, this gives me two `required_parameter`s

```typescript
  doThing(p1: number, p2: number | undefined = undefined): number {
    return p2 ?? p1
  }
```

I wonder if it's an issue in the parser?

queries/javascript/textobjects.scm
theHamsta: Apprently, arguments can have any expression as child.
```
      "types": [
        {
          "type": "_expression",
          "named": true
        },
        {
          "type": "spread_element",
          "named": true
        }
      ]
    }
```
https://github.com/nvim-treesitter/nvim-treesitter/pull/453 would bring us `(not-has-type? @parameter.inner "comment")`

theHamsta: @fsouza sorry, those queries were from Java (was fuzzy searching Java script)
these are the ones from JS:
```
(formal_parameters (identifier) @parameter)

(formal_parameters
  (rest_parameter
    (identifier) @parameter))

; ({ a }) => null
(formal_parameters
  (object_pattern
    (shorthand_property_identifier) @parameter))

; ({ a: b }) => null
(formal_parameters
  (object_pattern
    (pair
      value: (identifier) @parameter)))

; ([ a ]) => null
(formal_parameters
  (array_pattern
    (identifier) @parameter))

; a => null
(variable_declarator                                                                                                                                                                                        
    value: (arrow_function                                                                                                                                                                                    
      parameter: (identifier) @parameter))
```

theHamsta: So you could also have optional parameters in JS I guess for the highlights.

fsouza: Oh wait it looks like lambda_expression is a java thing, not javascript?

theHamsta: Right, I was grepping the wrong language.

fsouza: Made the change to the arguments query: https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/19/commits/0ad8731bf7ccda7cb58a3eaa1616d72357966103

Let me know if it looks good, then I can squash it :D

theHamsta: Yeah, looks good. Maybe we can tune this not-a-comment behavior later when we have predicate for it.

theHamsta: Should be fine :+1:

queries/javascript/textobjects.scm
```diff
@@ -38,3 +38,11 @@
 
 ;; blocks
 (_ (statement_block) @block.inner) @block.outer
+
+;; parameters
+(formal_parameters
+  (identifier) @parameter.inner)
```
theHamsta, line 44: this can also be an assignment_pattern (optional parameter) or object_pattern array_pattern ...

queries/javascript/textobjects.scm
fsouza: I double check the arguments appear to be fine!

queries/javascript/textobjects.scm
fsouza: Ohh, right. Thanks for double checking!

Should we be explicit here or just go with `(_)`?

```
(formal_parameters
  (_) @parameter.inner)
```

vs

```
;; parameters
(formal_parameters
  (identifier) @parameter.inner)

; function (v = default_value)
(formal_parameters
  (assignment_pattern) @parameter.inner)

; function ({ x }) ...
(formal_parameters
  (object_pattern) @paremeter.inner)

; function ([ x ]) ...
(formal_parameters
  (array_pattern) @paremeter.inner)

;; arguments
(arguments
  (_) @parameter.inner)
```

Thoughts?

fsouza: (pushed the latter, but let me know if you prefer the former)

theHamsta: Should be fine. Thanks.

PR #21, opened 2020-11-12
Fix #20: Add TSTextobjectSelect

### LINKED ISSUE #20: Easier option to define mapping manually
Hey guys,

thank you very much for your great work on all the treesitter features! I highly appreciate it! :pray: 

I must admit that I was never a friend of letting a plugin doing the key mappings for me. Also if it allows me to define a configuration option to define the key-combo to use. I feel like having less control, it becomes unstructured and ugly. At least that is my view. Unfortunately it seems to become a pattern for the more recent Lua'ish plugins to push for this kind of mappings.
So I denied to use the map definitions within the initialization step of this plugin. I'm adding them manually by myself where they belong to (from my perspective). Unfortunately this results in incredible long lines which look like this for example:

```vim
vnoremap <silent> af :<C-u>lua require'nvim-treesitter.textobjects.select'.select_textobject('@function.outer')<CR>
```

Talking about code quality this results into a fail that is hard to read, because 80% is equal boilerplate code. It is hard to see what is the actual relevant configuration here.

I wanted to ask if you guys are willed to add a simple command for each of those exposed functions. Just something as simple as `api.nvim_command('command! TSTextObjectSelect lua require....`. This would make it slightly easier to add such mappings by hand, removing a lot of the boilerplate.

---

I'm also open to get suggestions how I could improve my mappings else.
Thank you very much in advanced!

theHamsta:
Add a command for selecting a textobject. More commands can be added by filling the command dictionaries of the submodules (this should be upstreamed to do on registration of every nvim-treesitter module).

@weilbith If you want more commands you can create a PR. You need to do the remap for select and operator pending mode (their was a comment why select is better than visual mode for this plugin).

image

I guess it's more comfortable to set the mappings with a Lua plugin like
https://github.com/svermeulen/vimpeccable
```
vimp.nnoremap('<leader>hw', function()
  print('hello')
  print('world')
end)
```

or a own abbreviation
https://github.com/kyazdani42/dotfiles/blob/7a75655e387c0b4ec021f91d631027219b05b902/config/common/nvim/lua/init.lua#L69

### DIFF
--- a/lua/nvim-treesitter-textobjects.lua
+++ b/lua/nvim-treesitter-textobjects.lua
@@ -1,5 +1,6 @@
 local queries = require "nvim-treesitter.query"
 local configs = require "nvim-treesitter.configs"
+local utils = require "nvim-treesitter.utils"
 
 local M = {}
 
@@ -55,6 +56,9 @@ function M.init()
       }
     }
   }
+  for _, m in ipairs({'select', 'move', 'swap', 'lsp_interop'}) do
+    utils.setup_commands('textobjects.'..m, require ('nvim-treesitter.textobjects.'..m).commands)
+  end
 end
 
 return M
--- a/autoload/nvim_treesitter_textobjects.vim
+++ b/autoload/nvim_treesitter_textobjects.vim
@@ -1,0 +1,3 @@
+function! nvim_treesitter_textobjects#available_textobjects(arglead, cmdline, cursorpos) abort
+    return join(luaeval("vim.tbl_map(function(o) return '@'..o end, require'nvim-treesitter.textobjects.shared'.available_textobjects())"), "\n")
+endfunction
--- a/lua/nvim-treesitter/textobjects/attach.lua
+++ b/lua/nvim-treesitter/textobjects/attach.lua
@@ -1,4 +1,5 @@
 local configs = require'nvim-treesitter.configs'
+local utils = require'nvim-treesitter.utils'
 local parsers = require'nvim-treesitter.parsers'
 local queries = require'nvim-treesitter.query'
 local api = vim.api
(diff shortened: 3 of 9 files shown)


lua/nvim-treesitter-textobjects.lua
```diff
@@ -55,6 +56,9 @@ function M.init()
       }
     }
   }
+  for _, m in ipairs({'select', 'move', 'swap', 'lsp_interop'}) do
```
theHamsta, line 59: do this upstream in nvim-treesitter.define_module for every module?

theHamsta: will also import every module

weilbith:
That looks pretty cool! Thanks for the quick implementation. :pray: 
Adding the other commands looks so easy/smallish, do you think it makes much sense to split this into two PRs? :thinking:

theHamsta:
@weilbith can you test the PR? You should now be able to set the mappings in your `init.vim` with the `TSTextobjects*` commands.

weilbith:
Hmm. I'm confused. Switching to your fork and feature branch provides me with the new commands, but they don't work. Also the lua functions do not work. But for example my statusline segement using treesitter is still fine. Moreover I don't get any parameter suggestion. Just mentioning that because you have implemented it.
I'm not sure whats the issue here. Obviously my plugin manager has changed the plugin "source", else the commands would not be available. But the rest...

theHamsta:
I accidentally forked this branch from my outdated master instead of the current master. This the extension strings were missing e.g. for C++. Should be fixed now.

Can you try again? Also did you use the function with the leading `@` like `@function.outer`?

weilbith:
Alright. Now everything is working for me (okay, I did not test all commands and parameters). :tada: 

Besides the missing documentation, do you also wanna add a command for the peek definition? :thinking:

theHamsta:
PeakDefinition had not the `TSTextobjects` prefix like the other commands. Thanks for noting.

kyazdani42 (approved):
that looks good :)

PR #24, opened 2020-11-18
[Dart] Complete function.outter with annotation

GenesisTMS:


theHamsta (approved):
Thanks! Could there be annotations be also on other textobjects like classes?

GenesisTMS:
I looked at it more deeply. Fixed some conditional and parameter stuff. Also add annotations to the class you pointed out.
(is it okay to have it in this PR? i can make another since it does more than title says)

Readme updated...
Thanks!

queries/dart/textobjects.scm
```diff
@@ -10,20 +14,21 @@
...
-(function_body) @function.inner
-(type_alias (function_type) @function.inner) @function.outer
+(type_alias (function_type)? @function.inner) @function.outer
 
 ; parameter
-(formal_parameter) @parameter.inner
-(arguments (_) @parameter.inner)
-(normal_parameter_type) @parameter.inner
-(type_parameter) @parameter.inner
+(formal_parameter_list) @parameter.outer
```
theHamsta, line 24: parameter.outer is the parameter plus the following comma.

queries/dart/textobjects.scm
```diff
@@ -1,6 +1,10 @@
 ; class
-(class_definition
-  body: (class_body) @class.inner) @class.outer
+((
+  [(marker_annotation)? (annotation)?] @class.outer.start .
+  (class_definition 
+    body: (class_body) @_end @class.inner) @_start
+)
+(make-range! "class.outer" @_start @_end))
```
theHamsta, line 7: Thanks for adding this.

queries/dart/textobjects.scm
GenesisTMS: Ah, so unless there is a way how to select specific character after parameter (i don't know) this cannot be done with current dart treesitter parser.
Removing outer ones...

GenesisTMS: Ok i found implementation i `c` textobjects. So i added `parameter.outer`.

theHamsta (approved):

PR #30, opened 2020-12-07
feat: Added queries for php

Aerex:
I took the liberty to add some queries for php. I still new on the query language. I was not able to get `if_statement`
to work correctly at the moment. Also, I wanted the ability to select the body of a function, excluding the `{` and `}`.

I reviewed [Add @parameter.outer for Python/C/C++](https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/23) PR but was not able the result that I am looking for. I am not sure if excluding the braces is feasibly in treesitter at the moment.

### DIFF
--- a/queries/php/textobjects.scm
+++ b/queries/php/textobjects.scm
@@ -1,0 +1,27 @@
+(function_definition 
+ body: (compound_statement) @function.inner) @function.outer
+
+(method_declaration
+  body: (compound_statement) @function.inner) @function.outer
+
+(class_declaration
+  body: (declaration_list) @class.inner) @class.outer
+
+(foreach_statement
+  body: (_)? @loop.inner) @loop.outer
+
+(while_statement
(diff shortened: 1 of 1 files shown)


theHamsta:
Thanks for your contribution. Sorry for forgetting to review!

Regarding `{` `}`, at the moment tree-sitter makes it difficult to select the last named node of a node. In theory, this is possible

```scheme
(some_node .
         (node_type) @function.inner.start
         (node_type) @function.inner.end .        ; the dot enforces this to be the last named node in this block "}" is not named
          )
```
However, some bug prevents to capture the last node when another nodes is captured at the same time. The above example would work with out `last?` directive which we removed in favor of the built-in variant.

What could also work would be this `offset!` directive: (will be extended to `strip!` directive at some point): https://github.com/neovim/neovim/pull/13367

PR #37, opened 2021-02-06
Allow linewise selection

### LINKED ISSUE #36: Allow linewise selection
When attempting to make linewise selection of a textobject (for example, with `Vaf` for `function.outer`) I expect it to become linewise selection. Instead it is forced to be a regular selection (like an outcome of `vaf`). This is a consequence of how `nvim-treesitter.ts_utils.update_selection()` operates: it selects exactly the textobject forcing regular visual selection.

Linewise selection is useful, for example, when you want to delete lines with indented function definition.

Making explicit linewise visual selection possible.

Technically, adding extra `V` keystroke solves the issue: `vafV` forces linewise selection. However, it seems to present a little unnecessary cognitive load compared to regular vim experience.

**Additional context**
I will make a PR which utilizes the equivalence of expected `Vaf` and `vafV`.

echasnovski:
Fixes #36.

Example Python code:
```python
def a():
    print("Hello")
    def b():
        print("World")
```

Now with cursor inside `print("World")` these are now possible:
- `Vaf` selects lines 3-4.
- `dVaf` deletes lines 3-4 (utilizing operator-pending linewise mode).

There is one minor corner case: with cursor inside `def b()` using `Vaf` selects all lines 1-4. This seems to be because of how `nvim-treesitter.ts_utils.update_selection()` works: it updates selection to the "nearest" textobject containing the _whole_ current selection. After using `V` on line `def b()`, selection goes outside of scope of `b()` and the whole enclosing function is selected (if there is no enclosing function and indent is due to outer class, for example, nothing is selected). Currently I don't see a way of fixing this without updating `nvim-treesitter.ts_utils.update_selection()` directly.

### DIFF
--- a/lua/nvim-treesitter/textobjects/select.lua
+++ b/lua/nvim-treesitter/textobjects/select.lua
@@ -11,17 +11,22 @@ local M = {}
 function M.select_textobject(query_string, keymap_mode)
   local bufnr, textobject = shared.textobject_at_point(query_string)
   if textobject then
-    -- Detect if selection should be made linewise depending keymaping mode.
-    -- This should be done before calling `ts_utils.update_selection()` because
-    -- it forces 'v' mode.
-    local force_linewise = (keymap_mode == "x" and vim.fn.visualmode() == "V") or
-      (keymap_mode == "o" and vim.fn.mode(1) == "noV")
-    -- Make regular selection
-    ts_utils.update_selection(bufnr, textobject)
-    -- Possibly make selection linewise
-    if force_linewise then
-      vim.fn.nvim_exec("normal! V", false)
+    local selection_mode = "charwise"
+
+    -- Update selection mode based on keymapping mode
+    local ctrl_v = vim.api.nvim_replace_termcodes("<C-v>", true, true, true)
+    if keymap_mode == "x" then
+      local t = { V = "linewise"}
+      t[ctrl_v] = "blockwise"
+      selection_mode = t[vim.fn.visualmode()]
+    elseif keymap_mode == "o" then
+      local t = { noV = "linewise" }
+      t["no" .. ctrl_v] = "blockwise"
+      selection_mode = t[vim.fn.mode(1)]
     end
+
+    -- Make selection
+    ts_utils.update_selection(bufnr, textobject, selection_mode)
   end
 end
 
--- a/lua/nvim-treesitter/textobjects/select.lua
+++ b/lua/nvim-treesitter/textobjects/select.lua
@@ -26,9 +26,7 @@ function M.detect_selection_mode(keymap_mode)
     local method = keymap_to_method[keymap_mode]
 
     if method == "visual" then
-      local t = { V = "linewise"}
-      t[ctrl_v] = "blockwise"
-      selection_mode = t[vim.fn.visualmode()]
+      selection_mode = vim.fn.visualmode()
     elseif method == "operator-pending" then
       local t = { noV = "linewise" }
       t["no" .. ctrl_v] = "blockwise"
--- a/lua/nvim-treesitter/textobjects/select.lua
+++ b/lua/nvim-treesitter/textobjects/select.lua
@@ -16,24 +16,24 @@ function M.select_textobject(query_string, keymap_mode)
 end
 
 function M.detect_selection_mode(keymap_mode)
-    local selection_mode = "charwise"
-    local ctrl_v = vim.api.nvim_replace_termcodes("<C-v>", true, true, true)
+  local selection_mode = "charwise"
+  local ctrl_v = vim.api.nvim_replace_termcodes("<C-v>", true, true, true)
 
-    -- Update selection mode with different methods based on keymapping mode
-    local keymap_to_method = {
-      o = "operator-pending", s = "visual", v = "visual", x = "visual"
-    }
-    local method = keymap_to_method[keymap_mode]
+  -- Update selection mode with different methods based on keymapping mode
+  local keymap_to_method = {
+    o = "operator-pending", s = "visual", v = "visual", x = "visual"
+  }
+  local method = keymap_to_method[keymap_mode]
 
-    if method == "visual" then
-      selection_mode = vim.fn.visualmode()
-    elseif method == "operator-pending" then
-      local t = { noV = "linewise" }
-      t["no" .. ctrl_v] = "blockwise"
-      selection_mode = t[vim.fn.mode(1)]
-    end
+  if method == "visual" then
+    selection_mode = vim.fn.visualmode()
+  elseif method == "operator-pending" then
+    local t = { noV = "linewise" }
+    t["no" .. ctrl_v] = "blockwise"
+    selection_mode = t[vim.fn.mode(1)]
+  end
 
-    return selection_mode
+  return selection_mode
 end
 
 function M.attach(bufnr, lang)
--- a/lua/nvim-treesitter/textobjects/select.lua
+++ b/lua/nvim-treesitter/textobjects/select.lua
@@ -8,10 +8,20 @@ local ts_utils = require'nvim-treesitter.ts_utils'
 
 local M = {}
 
-function M.select_textobject(query_string)
+function M.select_textobject(query_string, keymap_mode)
   local bufnr, textobject = shared.textobject_at_point(query_string)
   if textobject then
+    -- Detect if selection should be made linewise depending keymaping mode.
+    -- This should be done before calling `ts_utils.update_selection()` because
+    -- it forces 'v' mode.
+    local force_linewise = (keymap_mode == "x" and vim.fn.visualmode() == "V") or
+      (keymap_mode == "o" and vim.fn.mode(1) == "noV")
+    -- Make regular selection
     ts_utils.update_selection(bufnr, textobject)
+    -- Possibly make selection linewise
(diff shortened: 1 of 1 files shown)


theHamsta (approved):

theHamsta:
Maybe you could have a look later on how `update_selection` can be fixed. The code for this plugin should be fine as it is right now.

Thanks for the contribution.

theHamsta:
Me: let's use git blame to see who wrote `update_selection`. This person surely will understand better how it works. Please don't let it be me.
Git blame: theHamsta (damned!)

echasnovski:
I had two problems with updating `update_selection`:

1. Not sure if it actually needs to respect linewise selection: I couldn't describe to myself what should be a general relation between "update selection" and "linewise". I ended up adding an extra argument but it seemed clunky.
2. I couldn't see a way to distinguish inside `update_selection` between normal, visual, and operator-pending linewise modes. Need to go deeper.

echasnovski:
> The code for this plugin should be fine as it is right now.

Sure, no problem. Of course, feel free to close this PR without merging if you think these changes don't belong here. I am totally fine with this :)

theHamsta:
> I couldn't see a way to distinguish inside update_selection between normal, visual, and operator-pending linewise modes. Need to go deeper.

couldn't that just be an optional parameter? `is_linewise_selection`

theHamsta:
> >  The code for this plugin should be fine as it is right now.

> Sure, no problem. Of course, feel free to close this PR without merging if you think these changes don't belong here. I am totally fine with this :)

I meant to say: the code is fine as it is with your PR.

echasnovski:
> Maybe you could have a look later on how update_selection can be fixed.

> couldn't that just be an optional parameter? is_linewise_selection

OK, so if optional parameter is an appropriate solution, then I can suggest this approach:
1. Update `ts_utils.update_selection()` to have new argument `selection_mode` which can be one of "charwise" (default), "linewise", and "blockwise". I am not sure why someone would need blockwise selection of node (which currently will select a block spanning from start position to end position of a node), but it is doable for a little cost and might be useful in the future. Maybe later blockwise selection will need to create a block from left-most to right-most columns and top-most to bottom-most rows?
    The actual update will replace [`vim.fn.nvim_exec("normal! v", false)`](https://github.com/nvim-treesitter/nvim-treesitter/blob/7055ba9c89831c8cddb4d3ed1e40352770c6d11d/lua/nvim-treesitter/ts_utils.lua#L148) with these lines:
```lua
-- Start visual selection in appropriate mode
  local v_table = {charwise = "v", linewise = "V", blockwise = "<C-v>"}
  ---- Call to `nvim_replace_termcodes()` is needed for sending appropriate
  ---- command to enter blockwise mode
  command = vim.api.nvim_replace_termcodes("normal! " .. v_table[selection_mode], true, true, true)
  vim.cmd(command)
```
2. Update this PR to utilize the new argument.

Does this sound good? If yes, I'll create PR into `nvim-treesiter` with update to `update_selection()`.

theHamsta:
Sounds good! I imagine that the linewise selection could be useful in more occasions!

echasnovski:
Made update that utilized proposed functionality of `selection_mode` in `ts_utils.update_selection()`. So this depends on merge of nvim-treesitter/nvim-treesitter#932. As a side note, this also means that users must update 'nvim-treesitter' in order to use this functionality.

Now not only linewise selection is possible, but also a blockwise. With example from the first issue:

- <C-v>af selects block from start to end position (the same as `vaf<C-v>`).
- d<C-v>af deletes that block (the same as `vaf<C-v>d`).

lua/nvim-treesitter/textobjects/select.lua
```diff
@@ -11,17 +11,22 @@ local M = {}
...
+    -- Update selection mode based on keymapping mode
+    local ctrl_v = vim.api.nvim_replace_termcodes("<C-v>", true, true, true)
+    if keymap_mode == "x" then
+      local t = { V = "linewise"}
+      t[ctrl_v] = "blockwise"
+      selection_mode = t[vim.fn.visualmode()]
+    elseif keymap_mode == "o" then
+      local t = { noV = "linewise" }
+      t["no" .. ctrl_v] = "blockwise"
+      selection_mode = t[vim.fn.mode(1)]
```
theHamsta, line 25: Couldn't the util function do all this complicated mapping? The util function could also accept "<c-v>" and all these cryptic abbreviations for all the modes.

theHamsta:
We should also replace our version of `get_visual_selection` with https://github.com/neovim/neovim/pull/13896/files as soon as it is available.

lua/nvim-treesitter/textobjects/select.lua
echasnovski: I can wrap these lines in separate function, like `M.detect_selection_mode(keymap_mode)`.

But I don't really understand the second part. Those abbreviations are coming from Vim's (well Neovim's) functionality. There are only three types of selections (charwise, linewise, blockwise) which can be detected in certain ways depending on keymapping mode. I can extend that utility function to also acknowledge "v" (Visual + Selection) and "s" (Selection) modes (with the same treatment as "x"), but this seems to be it. General mapping with "" will produce charwise selection.

echasnovski: Just pushed what I see the reasonable utility function. It defaults to "charwise" selection mode which might be updated on certain supported keymapping modes.

theHamsta: This can now be simplified with the updated upstream function, right?

echasnovski: Sure. I like this "symmetric" treatment of visual and operator-pending mode more, but I guess it has minor execution speed issues.

theHamsta: ok, so this is ready to be merged?

theHamsta: Are you familiar with this issue? https://github.com/nvim-treesitter/nvim-treesitter-textobjects/issues/39

echasnovski: > ok, so this is ready to be merged?

Just now spotted the 4-space indent instead of 2-space and pushed fix. Otherwise, this does what is discussed: adds possibility to make linewise and blockwise selection (during visual selection and in operator-pending mode) in the best way I can currently suggest.

> Are you familiar with this issue? #39

I've looked at it. Not an expert in intricacies of Vim's binding mechanisms, but the core of a problem seems to lie in logic of both 'ts_utils.update_selection()` and `TSTextobjectSelect`. `ts_utils.update_selection()` currently makes selection by sequentially moving to start of textobject, executing `v`, `V`, or `C-v`, and moving to end of textobject. At the same time `TSTextobjectSelect` accepts only one argument (which is query), so selection will always be done in with `v` command (default). So with enabled `<cmd>` mapping, typing `vaf` executes: `v`, go to start, `v` (which cancels first v), to the end. This looks like cursor just moves to the end of textobject.

However, I don't think this has something to do with this PR (which works with suggested method of mapping through `keymaps` option of `'nvim-treesitter.configs'.setup`).

theHamsta:
Nice work! Thank you for your persistence!

AckslD:
@echasnovski @theHamsta I also just noticed that when I eg yank a function this is charwise. I don't completely understand how this PR fixes that to use linewise instead. Should I make use of a different binding in the setup? I currently have:
```lua
...
    textobjects = {
        select = {
            enable = true,
            keymaps = {
                ["ac"] = "@class.outer",
                ["ic"] = "@class.inner",
                ["af"] = "@function.outer",
                ["if"] = "@function.inner",
            },
        },
...
```

echasnovski:
@AckslD, you can first visually linewise select and then yank: `Vafy`.

AckslD:
Thanks @echasnovski! So at the moment it's not possible to always have linewise? I'd prefer `yaf` over `Vafy` :)

echasnovski:
Oh, you are asking about some sort of setting to do it permanently. As far as I know, no, currently there is none.

AckslD:
Yeah! To me that's seems like the natural behaviour and what I expected it to do :)

AckslD:
@echasnovski @theHamsta What I ended up doing was a hack to setup a keybind that updates the register to be linewise afterwards:
```lua
local function omap(lhs, rhs)
    local opts = {silent = true, noremap = true}
    vim.api.nvim_set_keymap('o', lhs, rhs, opts)
end

local function setup_linewise_motion(motion, object, reg)
    omap(motion, ":lua require'nvim-treesitter.textobjects.select'.select_textobject('"..object.."', 'o')<CR>:call setreg('"..reg.."', getreg('"..reg.."'), 'V')<CR>")
end

local motions = {
    {'af', '@function.outer', '"'},
    {'if', '@function.outer', '"'},
    {'ac', '@function.outer', '"'},
    {'ic', '@function.outer', '"'},
}

for _, m in ipairs(motions) do
    setup_linewise_motion(unpack(m))
end
```

Note: This doesn't actually do the correct thing since `d`, `c` should update `"1` and `y` should update `"0` but good enough for me right now to get linewise motion. Also if someone uses the above, avoid having treesitter-textobjects also set the keybindings.

AckslD:
Btw I opened #70 which makes class and function objects linewise, let me know what you think

PR #41, opened 2021-03-13
Add first Julia queries

savq:
Add some basic text-object queries for Julia.


theHamsta (approved):
Nice! Thank you a lot!

PR #44, opened 2021-03-15
Fix typescript textobjects for nvim-treesitter/nvim-treesitter#1000

theHamsta:
@fsouza Some parameter queries will stop to work for typescript

### DIFF
--- a/queries/typescript/textobjects.scm
+++ b/queries/typescript/textobjects.scm
@@ -1,4 +1,47 @@
-; inherits: javascript
+; inherits: (jsx)
+(function_declaration
+  body: (statement_block) @function.inner) @function.outer
+
+(export_statement
+  (function_declaration) @function.outer) @function.outer.start
+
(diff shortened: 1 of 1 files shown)


fsouza:
@theHamsta thanks for taking care of this!

theHamsta:
If you have time you could look for the equivalent typscript nodes for the javascript nodes that stopped working.

fsouza:
@theHamsta I can take a look later today. Will let you know!

theHamsta:
We have the same problem in core. I out-commented all parameter queries until someone sorts out (actually kyazdani42 or steelsojka wanted to that)

fsouza:
@theHamsta I may have to look into this later after merging. Can send a PR at some point :D

fsouza (approved):
Just tested it locally and it works with all cases that we had for javascript before, so we should be good. Will review core stuff.

PR #52, opened 2021-04-14
Allow to use "." for swapping

### LINKED ISSUE #35: Swap doesn't repeat with . operator
If I have a piece of code `f(a, b, c, d)` and I put my cursor on `a` and perform swap_next for `@parameter.inner`, it does not compose with `.`; after the initial swap I have `f(b, a, c, d)` with the cursor still on `a`, and I would expect that `.` would once more do another swap (leaving `f(b, c, a, d)`). Instead `.` ignores the swap does whatever happened before that (eg an insertion).

Thanks for this plugin, by the way; enabling this kind of thing is to me one of the most exciting outcomes of treesitter.

theHamsta:
Fix #35

### DIFF
--- a/lua/nvim-treesitter/textobjects/swap.lua
+++ b/lua/nvim-treesitter/textobjects/swap.lua
@@ -4,6 +4,8 @@ local attach = require'nvim-treesitter.textobjects.attach'
 
 local M = {}
 
+local CR = vim.api.nvim_replace_termcodes("<cr>", true, true, true)
+
 local function swap_textobject(query_string, direction)
   local bufnr, textobject_range, node = shared.textobject_at_point(query_string)
   if not node then return end
@@ -20,10 +22,12 @@ end
 
 function M.swap_next(query_string)
   swap_textobject(query_string, 1)
+  pcall(vim.fn["repeat#set"], ":TSTextobjectSwapNext "..query_string..CR)
 end
 
 function M.swap_previous(query_string)
   swap_textobject(query_string, -1)
+  pcall(vim.fn["repeat#set"], ":TSTextobjectSwapPrevious "..query_string..CR)
 end
 
 local normal_mode_functions = {"swap_next",


lua/nvim-treesitter/textobjects/swap.lua
```diff
@@ -20,10 +20,12 @@ end
 
 function M.swap_next(query_string)
   swap_textobject(query_string, 1)
+  vim.fn["repeat#set"](":TSTextobjectSwapNext "..query_string..vim.api.nvim_replace_termcodes("<cr>", true, true, true))
```
theHamsta, line 23: Is there a way to do this less vim-scripty :vomiting_face:

theHamsta: Also this is not silent...

theHamsta: And I really should make `find_best_match` faster... So much to do!

theHamsta: Does this require `tpope/vim-repeat`?

stsewd (approved):

andymass:
If possible I would recommend avoiding vim-repeat.  Instead you can use vim's native `g@` like in this example: 
 https://gist.github.com/andymass/d465970cc178452eac006840e5e58b2b

Specifically, although `g@` is usually used as the start of normal operators you can also use it for entire normal commands with the sequence `g@l`.

theHamsta:
@andymass can you create a PR for it? I can't think of a elegant way to implement this. As far as I understood, we must set operatorfunc. But we can't refer to our Lua functions with `v:lua` because we have a `-` in our folder name. Also we would have to create a outer function that sets the operatorfunc and an inner that does the actual action.

 Can't you use the `g@l` hack already now in your config?

andymass:
> Can't you use the g@l hack already now in your config?

Yes, it's possible via a wrapper function;

```vim
function! s:repeatable(cmd)
  function! s:inner(...) closure abort
    execute a:cmd
  endfunction
  let &opfunc=get(funcref('s:inner'), 'name')
  return 'g@l'
endfunction

nnoremap <silent><expr> <left>
      \ <sid>repeatable('TSTextobjectSwapPrevious @parameter.inner')
nnoremap <silent><expr> <right>
      \ <sid>repeatable('TSTextobjectSwapNext @parameter.inner')
```

> But we can't refer to our Lua functions with v:lua because we have a - in our folder name.

I think that could be fixed fairly easily, but a problem remains that there would have to be some state held somewhere and I'm not sure how to refer to lua inner functions in operatorfunc.

> can you create a PR for it?

Here is a first shot at it using the same technique as above: #55 .  I'm not sure how to do it with pure lua.

stellarhoof:
Ping

kuator:
Bump

theHamsta:
Super-seded by #343 or #55

PR #55, opened 2021-04-23
Allow repeating swap operation

andymass:
An attempt at #52; allows repeating swap with `.`

### DIFF
--- a/autoload/nvim_treesitter_textobjects.vim
+++ b/autoload/nvim_treesitter_textobjects.vim
@@ -1,3 +1,11 @@
 function! nvim_treesitter_textobjects#available_textobjects(arglead, cmdline, cursorpos) abort
     return join(luaeval("vim.tbl_map(function(o) return '@'..o end, require'nvim-treesitter.textobjects.shared'.available_textobjects())"), "\n")
 endfunction
+
+function! nvim_treesitter_textobjects#repeatable(cmd) abort
+    function! s:inner(...) closure abort
+        execute a:cmd
+    endfunction
+    let &operatorfunc=get(funcref('s:inner'), 'name')
+    return 'g@l'
+endfunction
--- a/lua/nvim-treesitter/textobjects/attach.lua
+++ b/lua/nvim-treesitter/textobjects/attach.lua
@@ -4,7 +4,7 @@ local queries = require'nvim-treesitter.query'
 local api = vim.api
 local M = {}
 
-function M.make_attach(normal_mode_functions, submodule)
+function M.make_attach(normal_mode_functions, submodule, repeatable)
   return function(bufnr, lang)
     local config = configs.get_module("textobjects."..submodule)
     local lang = lang or parsers.get_buf_lang(bufnr)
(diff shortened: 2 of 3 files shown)


autoload/nvim_treesitter_textobjects.vim
```diff
@@ -1,3 +1,11 @@
...
     return join(luaeval("vim.tbl_map(function(o) return '@'..o end, require'nvim-treesitter.textobjects.shared'.available_textobjects())"), "\n")
 endfunction
+
+function! nvim_treesitter_textobjects#repeatable(cmd) abort
+    function! s:inner(...) closure abort
+        execute a:cmd
+    endfunction
+    let &operatorfunc=get(funcref('s:inner'), 'name')
+    return 'g@l'
+endfunction
```
theHamsta, line 11: This could as well been written in Arabic for me.

lua/nvim-treesitter/textobjects/attach.lua
```diff
@@ -18,7 +18,13 @@ function M.make_attach(normal_mode_functions, submodule)
...
         if query then
           local cmd = ":lua require'nvim-treesitter.textobjects."..submodule.."'."..function_call.."('"..query.."')<CR>"
-          api.nvim_buf_set_keymap(bufnr, "n", mapping, cmd, {silent = true, noremap = true })
+
+          if repeatable then
+            local fn = 'nvim_treesitter_textobjects#repeatable("'..cmd..'")'
+            api.nvim_buf_set_keymap(bufnr, "n", mapping, fn, {silent = true, noremap = true, expr = true })
+          else
+            api.nvim_buf_set_keymap(bufnr, "n", mapping, cmd, {silent = true, noremap = true })
+          end
```
theHamsta, line 27: This looks very nice to do it for all possible functions! But it will not work when calling the Lua function directly or when using the command the command :TSTextobjectSwapNext. It's also a bit strange that you can only make all functions of a module repeatable (which would still work for our current use case.

stellarhoof:
Is this stale?

theHamsta:
@clason it does not seem to be possible to set `operatorfunc` to a Lua function. I would prefer to wait for this feature or a general way to make a Lua function repeatable (without

The best thing that I could get working was
```lua
        local function fn()
          if opts.repeatable then
            _G._nvim_treesitter_textobject_last_function = function()
              require("nvim-treesitter.textobjects." .. submodule)[function_call](query)
            end
            vim.o.opfunc = "v:lua._nvim_treesitter_textobject_last_function"
            vim.api.nvim_feedkeys("g@l", "n", false)
          else
            require("nvim-treesitter.textobjects." .. submodule)[function_call](query)
          end
        end
        vim.keymap.set(
          keymap_modes,
          mapping,
          fn,
          { buffer = bufnr, silent = true, remap = false, desc = mapping_description }
        )
```
since `opfunc` seems to insist on a string

clason:
> it does not seem to be possible to set operatorfunc to a Lua function. 

That's right; this is tracked here: https://github.com/neovim/neovim/issues/18132

theHamsta:
Supersede by #343

PR #56, opened 2021-04-23
Update CodeQL textobjects

pwntester:


queries/ql/textobjects.scm
```diff
@@ -1,7 +1,13 @@
...
-;; function textobject
-(classMember) @function.outer
-(memberPredicate (body) @function.inner)
+;; predicate textobject
+(charpred (className) (_) @function.inner) @function.outer
+(memberPredicate (body) @function.inner) @function.outer
 (classlessPredicate (body) @function.inner) @function.outer
+
+;; identifier textobject
+(dataclass name: (className) @identifier.outer)
```
theHamsta, line 10: How is a identifier textobject different from `iw`?

pwntester: My goal is to use these identifiers with movements (eg: `[i`) rather than text objects, so I want to be able to jump to the name of the function from within the body of the function for example

theHamsta: so it's not every identifier but rather function/class names. Maybe just `@scopename.inner`? `@functionname.inner`?

pwntester: Thanks, will change it to `@scopename.inner`

theHamsta: Could you add this with a description to CONTRIBUTING.md? Then it will also appear on the README.md automatically

pwntester: Screwed up with my fork and delete it accidentally 🤦🏼 so recreated it in a different PR
https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/57

PR #58, opened 2021-04-23
Update Julia queries

savq:
* Update queries for `@parameter_outer` in:
    - call_expression / argument_list
    - function_definition / parameter_list
* Add queries for:
    - structs
    - math functions
    - macros

### DIFF
--- a/queries/julia/textobjects.scm
+++ b/queries/julia/textobjects.scm
@@ -1,16 +1,24 @@
-(compound_expression) @block.outer    ;; begin blocks
+(compound_expression) @block.outer    ; begin blocks
 (let_statement)       @block.outer
 
-(call_expression) @call.outer
+(call_expression
+  (identifier)
+  (argument_list) @parameter.outer) @call.outer
+
+(struct_definition) @class.outer    ; not classes, but close enough
 
 (comment) @comment.outer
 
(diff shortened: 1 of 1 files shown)


queries/julia/textobjects.scm
```diff
@@ -1,16 +1,24 @@
...
+(struct_definition) @class.outer    ; not classes, but close enough
 
 (comment) @comment.outer
 
 (if_statement) @conditional.outer
 
-(function_definition) @function.outer
-(function_expression) @function.outer    ;; lambdas
+(function_definition
+  name: (identifier)
```
theHamsta, line 15: why do we need the `name` in the query? Can't the function as well be anonymous and if not then `name: (identifier)` does not really add a benefit.-

theHamsta (approved):
Nice! Really glad to receive this addition!

queries/julia/textobjects.scm
theHamsta: Ok. It's a function_expression without the name.

savq: I actually forgot to check that case 😬

Good thing it's covered by the other query, although it does seem a bit confusing. Not sure why the grammar would be defined that way.

PR #60, opened 2021-04-27
Allow other query groups for "textobject.select"

theHamsta:
Fixes https://github.com/nvim-treesitter/nvim-treesitter/issues/1258

TODO: other modules

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ require'nvim-treesitter.configs'.setup {
         ["ic"] = "@class.inner",
 
         -- You can also take textobjects from other query groups
-        ["aS"] = { "@scope", "locals" }, -- selects `@scope` from locals.scm
+        ["aS"] = { "@scope", "locals" }, -- selects `@scope` from locals.scm (you could also use injection.scm or highlights.scm)
 
         -- Or you can define your own textobjects like this
         ["iF"] = {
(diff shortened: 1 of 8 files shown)


lua/nvim-treesitter/textobjects/select.lua
```diff
@@ -1,15 +1,15 @@
 local api = vim.api
-local configs = require'nvim-treesitter.configs'
-local parsers = require'nvim-treesitter.parsers'
-local queries = require'nvim-treesitter.query'
+local configs = require "nvim-treesitter.configs"
```
theHamsta, line 2: This change caused some lines in this file to get longer so it was time for a auto-format.

lua/nvim-treesitter/textobjects/lsp_interop.lua
```diff
@@ -38,7 +38,7 @@ function M.preview_location(location, context)
   return vim.lsp.util.open_floating_preview(contents, filetype)
 end
 
-function M.make_preview_location_callback(textobject, context)
+function M.make_preview_location_callback(textobject, query_group, context)
```
theHamsta, line 41: we could also limit the "other query groups" feature to `select` only.

lua/nvim-treesitter/textobjects/move.lua
```diff
@@ -42,20 +42,31 @@ local function move(query_string, forward, start, bufnr)
     end
   end
 
-  local match = queries.find_best_match(bufnr, query_string, 'textobjects', filter_function, scoring_function)
-  local config = configs.get_module('textobjects.move')
+  local match =
+    queries.find_best_match(bufnr, query_string, query_group or "textobjects", filter_function, scoring_function)
```
theHamsta, line 46: @synek @tom-anders I would just recommend to do the same for `folds`. Just have `"folds"` and `@fold` be the default but make it and option to change it to something else.

README.md
```diff
@@ -105,6 +108,27 @@ require'nvim-treesitter.configs'.setup {
...
+lua <<EOF
+require'nvim-treesitter.configs'.setup {
+  textobjects = {
+     move = {
+      enable = true,
+      goto_next_start = {
+        ["]F"] = { "@definition.function", "locals" },
+      },
+      goto_previous_start = {
+        ["[F"] = { "@definition.function", "locals" },
```
theHamsta, line 125: @synek @tom-anders this config hijacks the queries in `locals.scm` to be used as textobject. It would also be possible to fold on textobjects. We already have classes, functions, etc. Next, step would be to allow `@function.outer` *and* `@class.outer` at the same time.

theHamsta:
Super-Seeded by #360

PR #64, opened 2021-05-27
Add textobjects.select.lookahead option to select the next textobject

RRethy:
Haven't added docs, if you think this functionality is in scope of the plugin let me know and I can add docs.

This allows `select` to look ahead in the syntax tree if it cannot find a match with the cursor position in it. E.g.

```lua
local x = 'hello' -- <- your cursor is here and you press `vaf`
function foo()  -- ----
  print('foo')  --     | <- this would be selected
end             -- ----
```

Sample Config:

```lua
treesitter.setup {
    textobjects = {
        select = {
            enable = true,
            lookahead = true,
            keymaps = {
                ['af'] = '@function.outer',
            }
        },
    }
}
```

### DIFF
--- a/lua/nvim-treesitter/textobjects/shared.lua
+++ b/lua/nvim-treesitter/textobjects/shared.lua
@@ -13,7 +13,8 @@ function M.available_textobjects(lang)
   return found_textobjects
 end
 
-function M.textobject_at_point(query_string, pos, bufnr, lookahead)
+function M.textobject_at_point(query_string, pos, bufnr, opts)
+  opts = opts or {}
   bufnr =  bufnr or vim.api.nvim_get_current_buf()
   local lang = parsers.get_buf_lang(bufnr)
   if not lang then return end
@@ -69,7 +70,7 @@ function M.textobject_at_point(query_string, pos, bufnr, lookahead)
           end
         end
       end
-    elseif lookahead then
+    elseif opts.lookahead then
       local start_line, start_col, start_byte = m.node:start()
       if start_line > row or start_line == row and start_col > col then
         local length = ts_utils.node_length(m.node)
--- a/lua/nvim-treesitter/textobjects/shared.lua
+++ b/lua/nvim-treesitter/textobjects/shared.lua
@@ -13,7 +13,7 @@ function M.available_textobjects(lang)
   return found_textobjects
 end
 
-function M.textobject_at_point(query_string, pos, bufnr)
+function M.textobject_at_point(query_string, pos, bufnr, lookahead)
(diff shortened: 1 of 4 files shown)


theHamsta:
I haven't tested this yet but the functionality is nice. The reason why this hasn't been merged yet is that I originally wanted to change the logic of that function or finding textobjects in general. The function and logic of `get_textobject_at_point` is way too complicated for something that is essentially not much different from `next/previous_textobject`.

In particular I want to get rid of `get_capture_matches_recursively` because it used the slow nvim-treesitter query aggregation instead of directly using the API provided by nvim. I wanted that all functions would use a version of `find_best_matches` or better `find_n_best_matches` where you can steer the behavior of the function by a scoring function and filter function, This makes it much easier to implement a feature of counts and also look ahead / look behind much easier (you can just score size and position of the textobject) and I can try to make this function `find_n_best_matches` fast as it can be reused in many places of nvim-treesitter. 

But the code looks nice so I can review it on the weekend and we have a nice intermediate solution that I can probably copy some parts over to the refactor.

theHamsta (approved):

theHamsta:
Feels very cool! Thanks for this nice work. But a refactor for this plugin is overdue. I have the feeling that this would also simplify the implementation of this feature.

lua/nvim-treesitter/textobjects/shared.lua
```diff
@@ -13,7 +13,7 @@ function M.available_textobjects(lang)
   return found_textobjects
 end
 
-function M.textobject_at_point(query_string, pos, bufnr)
+function M.textobject_at_point(query_string, pos, bufnr, lookahead)
```
theHamsta, line 16: Just in case that there will be more customizations in the futures... Could you put lookahead in a options/opts parameter?

RRethy:
If you do refactor would you also be changing the plugin interface? I'm wondering because it might be worthwhile to change the `textobjects.select.keymaps` semantics for table values, right now a table value for keymaps indicates individual queries for specific languages, but changing the table to be a list of opts (e.g. `langs`, `lookahead`) would allow for some interesting customization that doesn't need to apply to all the keymaps (like `lookahead` does right now in this PR). For example, a single keymap might want to be able to change the operator-pending behaviour to select with `V` (by line) instead of with `v` (current behaviour). Using `V` instead of `v` would be useful for `@function.outer` but less useful for `@call.outer`. I know breaking backwards compatibility can be rough, but since the language specific queries doesn't work (#65) right now anyways, I think it would be worthwhile if there is a refactor.

Really enjoy the plugin btw.

stsewd (approved):
Thanks!

PR #67, opened 2021-06-02
Add initializer_list to @parameter for cpp queries

alexander-born:
For cpp files `@parameter` does not support `initializer_list`. This adds support for `initializer_list`.


theHamsta:
Thanks!

PR #75, opened 2021-06-28
Add lookahead readme

singlexyz:
Add `lookahead` option summary.
```
...
  -- Auto look forward to the textobj. Vim's default textobj works in this way. 
  lookahead = true,
...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,10 @@ require'nvim-treesitter.configs'.setup {
   textobjects = {
     select = {
       enable = true,
+
+      -- Auto look forward to the textobj. Vim's default textobj works in this way. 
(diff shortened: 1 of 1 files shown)


theHamsta:
I don't think Vim's default work this way. It's what targets.vim does

README.md
```diff
@@ -139,148 +143,131 @@ EOF
 7. @comment.outer
 8. @conditional.inner
 9. @conditional.outer
-10. @frame.inner
```
theHamsta, line 142: This part of the README is auro-generated

theHamsta:
I didn't mean that you have to close. Just mention that it's similar to targets.vim

PR #82, opened 2021-07-07
Fix lsp_interop linting

ranjithshegde:
Apologies about the snafu with linting and multi commits.
This should fix it


stsewd (approved):
Thanks!

PR #92, opened 2021-07-25
Added Julia inner text objects

Oliver-Leete:
This should add inner text objects for julia language. I'm still learning treesitter queries, so some of them might not be the most elegant solution, but I have tested them all and they seem to work with every edge case I could think of.

The one thing I couldn't figure out is how to get the let_statement inner.block to not also include the variable declarations at the start of the block. Best I could manage was to make it not include the first variable declaration, but I thought that it was better to have it consistent but wrong so I left it as including all of them.

### DIFF
--- a/queries/julia/textobjects.scm
+++ b/queries/julia/textobjects.scm
@@ -1,21 +1,104 @@
-(compound_expression) @block.outer    ; begin blocks
-(let_statement)       @block.outer
+; Blocks
+((compound_expression
+  . (_)? @_start
+  (_) @_end .)
+(#make-range! "block.inner" @_start @_end)) @block.outer
+((let_statement
+  . (_)? @_start
+  (_) @_end .)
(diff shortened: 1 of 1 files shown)


theHamsta:
Thanks for your contribution :heart: !

Oliver-Leete:
It looks like the docgen for the readme file hasn't picked up any of the text objects that are only defined by make-range. Looking at other languages it seems to be the same, as the table shows nothing supporting outer parameter. I can't quite tell how the table is being made (my lua isn't great), but would adding a comment including```@parameter.outer``` ect for the missing text objects fix it?

theHamsta:
No, not really. It doesn't have access to the strings of the queries at the moment and looks whether the queries have the captures `@parameter.outer`. Probably, they should also do a string search or analyze the `make-range!` directive.

PR #93, opened 2021-07-25
Expand coverage of parameter.outer

IndianBoy42:
This is based on taking all the existing parameter.inner captures
and just adding the outer part according to the template.
This was done to go, java, javascript, lua

Also:
- typescript now is just inherited directly from javascript. this seems to work okay
- Julia and Python have added subscript expressions as parameters, that seemed appropriate
  - Especially because they have and use multidimensional indexing a lot more than most other languages (atleast in data science)


theHamsta (approved):
Thanks!

PR #97, opened 2021-08-04
Add hooks support

glyh:
There are some good builtin commands in the nvim-treesitter-textobjects, however, it's hard for user to add their own command. 

Adding hooks in the modules will enable users to do what ever they want.

### DIFF
--- a/doc/nvim-treesitter-textobjects.txt
+++ b/doc/nvim-treesitter-textobjects.txt
@@ -150,4 +150,25 @@ Supported options:
   }
   EOF
 <
+                                                *nvim-treesitter-textobjects-hooks*
+Hooks~
+- Define your own hooks, so that every time a function of this plugin is invoked,
+  The hook function will be called with the argument passed to that function.
+
+  For example:
+>
+  lua <<EOF
+  require'nvim-treesitter.configs'.setup{
+    textobjects = {
+      shared = {
+        hook = {
+          textobject_at_point = function(query_string, pos, bufnr, opts)
+            print(query_string)
+          end
+        }
+      }
+    }
+  }
+  EOF
+<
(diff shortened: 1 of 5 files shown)


theHamsta:
I'm not sure if most users including me would understand how and in which situation one could use this feature. Could you share a use case? What's the advantage over manually patching the function?

```lua
require 'module'.fun = function (...) before(...) require 'module'.fun(...) end
```

glyh:
Well, the actually use case is that I want to know the query string given to this module, so that I can do different things on after different selection(for example) ... 

I think you're right that we can manually do this.....

glyh:
I don't know why but this won't work(Called after setup treesitter)
```lua
local select =
  require "nvim-treesitter.configs".get_module("textobjects.select")

local selectfn = select.select_textobject
```

theHamsta:
get_module is not named well. It will give you the config of module and whether it's enabled. Use require'nvim-treesitter.textobjects.share` to get the actual lua file

glyh:
> equire'nvim-treesitter.textobjects.share

This is not working as well, I've tried it after require("nvim-treesitter.configs").setup :
```lua
require'nvim-treesitter.textobjects.share'
```

theHamsta:
Are you using the exact path of the lua file? In this case the file was called `shared` https://github.com/nvim-treesitter/nvim-treesitter-textobjects/tree/master/lua/nvim-treesitter/textobjects


I'm on mobile so it's difficult to avoid mistakes

glyh:
Sorry but I'm testing another module, the shared one works, but `select` don't seem so:
```lua
local select = require'nvim-treesitter.textobjects.select' -- won't work
local shared = require'nvim-treesitter.textobjects.shared'
```

PR #103, opened 2021-08-21
Adds zig textobjects

Swoogan:
Adds all the basic textobjects for Zig. Might be missing some more advanced things like lambdas. I'm not 100% sure about the class object implementations where I manually select "{" and "}". They work, but they appear to be a non-standard implementation.


Swoogan:
I'm not sure what to do with these errors. I downloaded the appimage on my machine and ran the same command without error. I don't know what else I can do to reproduce this, and the error is not helpful enough for me to deduce the problem.

PR #104, opened 2021-08-21
Adds zig textobjects

Swoogan:
Adds the basic zig text objects. (we'll see if this one builds)


Swoogan:
I tried running this build in my fork with the rust textobjects.scm file copied into the zig folder and no other changes, it still failed. Also, I could still not recreate the issue. It's possible there is something wrong with the build environment, or maybe an issue with the combination of the zig grammar and treesitter. I really don't know at this point.

theHamsta:
@Swoogan the rust queries won't work for zig. Did you try out nvim-treesitter/playground to see the node names that the zig parser is using? We also switched the Tree-Sitter parser for zig which uses different node names compared to the previous one

Swoogan:
I kind of realized that after making the comment. However, before that I tried many builds with a massively reduced file and nothing would work. Thanks to your info about the zig parser changing, I realized that I had installed the zig parser via the `0.5-compat` branch of treesitter. That's unfortunate.

PR #109, opened 2021-08-28
Update zig textobjects

maxxnino:
- I updated, and tested with the new parser on my machine

- I don't know the proper way to use the textobject query. Should i separate them in multiple query. Example

```
(_
  (IfPrefix) @conditional.inner
  (_) @conditional.inner
  (ElseTail)? @conditional.inner) @conditional.outer
```

### DIFF
--- a/queries/zig/textobjects.scm
+++ b/queries/zig/textobjects.scm
@@ -1,6 +1,6 @@
 ;; "Classes"
 (VarDecl 
-  (_ (ContainerDecl (ContainerMembers) @class.inner))) @class.outer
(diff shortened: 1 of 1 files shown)


theHamsta:
@maxxnino probably multiple queries for the inner-textobject would be good as nvim-treesitter is using the old nvim-treesitter Lua abstraction over the queries where you can have each query only once. You could also have a `[` `]` alternation around the inner queries to ensure the `@inner` captures only appear once per query.

I will merge right a away as the current state in master is clearly broken and this would be a good hot-fix.

PR #107, opened 2021-08-23
Add more ruby text objects

jeetelongname:
I have added all the inner variants of the supplied text objects as well
as 
- adding modifier variants of conditionals and loops, 
- comments and
- parameters support

I have tried keep these query's efficient but there may be room for improvement
[](https://github.com/hlissner/doom-emacs/pull/5401)

PR #113, opened 2021-09-05
Fix some ruby text objects

luathn:
I saw this [PR](https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/107) added some text objects for Ruby, but it has some bugs and unhandled cases.
So I created this PR to fix it. Hope you can take a look.
I'm Ruby dev, so I'm very glad if you can mention to me if have any changes in the Ruby text object, I can check it to be sure it's correct.

## Commit
Fix class, function and block text objects, tested with many cases from:
https://github.com/tree-sitter/tree-sitter-ruby

Add some cases for parameter text object.

### DIFF
--- a/queries/ruby/textobjects.scm
+++ b/queries/ruby/textobjects.scm
@@ -1,27 +1,139 @@
-;; blocks
-(call
-  block: (_) @block.inner) @block.outer
+;; @functions
+(
+  [
+    (method
+      . name: (identifier) 
+      !parameters 
+      . (_) @_start (_) @_end .)
+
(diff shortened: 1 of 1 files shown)


theHamsta:
Thanks @luathn! I will try to remember you when we have changes to the Ruby textobjects.

theHamsta (approved):
Looks good to me. I personally don't have experience with Ruby so good that you're offering help :+1:

luathn:
Thanks @theHamsta.

PR #114, opened 2021-09-11
Add elixir queries

natdm:
Adding Elixir text objects.

Unfortunately, pretty much everything in the elixir grammar is a function or a call, so it's pretty hard to match some of the requested captures, as a lot of them would just be redundant. I followed the javascript queries to get as close as I can to matching what that matches and it's pretty spot on.

Adding:
- `@comment.outer` to match any `#comment`, though there was already a query to get module docs
- `@conditional.outer/inner` for all `if/cond/case`
- `@parameter.inner/outer` which matches arguments and parameters.

Altered:
- `@class.inner/outer` to remove the unneeded brackets.

Updating readme to match what was already there and what's new, as well.


theHamsta:
The README is auto-generated.

theHamsta (approved):
Thanks for your contribution!

natdm:
> The README is auto-generated.

Apologies. That might be something to call out in the contribution doc.

theHamsta:
The check fails on all of nvim-treesitter repo. It's not our fault but due to nvim AppImage not being available today. I would wait until it works again before I merge.

PR #115, opened 2021-09-11
Update README.md

luathn:
I don't see Github action update the ruby ​​row in Readme so I created this PR to update it.
Please take a look.


theHamsta:
Thanks, but it's just because all our Linux CI has stopped because the nvim AppImage is not downloadable. It should run again in future.

PR #118, opened 2021-09-12
Fix tsx support

insidewhy:
The current tsx support is totally non-functional, with this change many things work. As far as I can tell it supports everything typescript supports.

It might not be the ideal future direction of the plugin, but at least some things work for now so maybe it's worth considering this as a pragmatic change for the time being?

### DIFF
--- a/queries/typescript/textobjects.scm
+++ b/queries/typescript/textobjects.scm
@@ -1,1 +1,1 @@
-; inherits: (javascript)
+; inherits: javascript


queries/tsx/textobjects.scm
```diff
@@ -1 +1 @@
-; inherits: typescript,jsx
```
theHamsta, line 1: But typescript should include JS. Can you explain what is not working with the current tsx queries?

theHamsta (changes requested):

queries/tsx/textobjects.scm
insidewhy: @theHamsta Absolutely nothing is working with the current `tsx` queries, I don't understand at all why.

I tried ` ; inherits (typescript)` also and that didn't work.

However `; inherits (javascript)` works great.

When you look at `typescript/textobjects.scm` it's doing `; inherits (javascript)` also, and in turn, `javascript` is already inheriting `jsx`.

I thought maybe `; inherits typescript,jsx` might be breaking because `typescript` already imports `javascript` which imports `jsx` leading to a double import of `jsx`, but then why does `; inherits (typescript)` also not work.

I'm thinking, maybe something is fundamentally broken with the inheritance system? As if transitive inheritance isn't working?

insidewhy: @theHamsta I guess it's also important to note that `tsx` is the *only* `textobjects.scm` that uses transitive inheritance so it might well make sense that transitive inheritance doesn't work.

In that case maybe this is a pragmatic change to make until transitive inheritance can be fixed (maybe you don't even need to support transitive inheritance?, in this case it's pretty much redundant given `; inherits: (javascript)` is already completely equivalent)

theHamsta: can you remove the parenthesis in `typescript/textobjects.scm` it's working fine for me. There have been some problems with JS/TS having different parameter nodes but this problems does not seem to be triggered with our current queries. The parenthesis mean that this language is not automatically transitioned to indirect dependencies.

insidewhy: Cool I'll try that.

insidewhy: @theHamsta Great, thanks for the advice, that works! Updated this PR and everything is looking good.

theHamsta (approved):

insidewhy:
@theHamsta Thanks so much for helping me out so quickly and merging this quickly. One thing I just realised, I should have updated the table in the README to add the same thumbs ups as typescript.

theHamsta:
This should actually be done by a CI job in merge to master.. I'll try to figure out what's wrong there

PR #119, opened 2021-09-15
Update new signature handler

luathn:
After this [PR](https://github.com/neovim/neovim/pull/15504) is merged, I can't use `peek_definition_code`.
So this PR to check and use a correct callback.


theHamsta (approved):
Thanks for this update!

PR #122, opened 2021-09-25
Fennel textobjects

sevanteri:
First time doing Treesitter queries, please be gentle. :D

Couldn't get `#offset!` working for `@parameter.outer`, not sure what's up. I left them there anyways as they don't seem to cause any issues either.

Fennel doesn't have statements, but I defined `@statement.outer` to be anything surrounded by parenthesis. I found it useful for at least selecting, moving, and swapping.

### DIFF
--- a/queries/fennel/textobjects.scm
+++ b/queries/fennel/textobjects.scm
@@ -1,0 +1,63 @@
+
+; https://fennel-lang.org/reference
+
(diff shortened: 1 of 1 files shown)


theHamsta:
The offset! Predicate is a hack. It only works for injections.

theHamsta (approved):

PR #123, opened 2021-09-27
Python query: fix tuple and list @parameter.inner

dvzubarev:
Consider following code:

```pyhton
l = [1,2, |"str"]
```
@parameter.inner does not capture any node.  
The syntax tree of this list is

```
      list:
        integer:
        integer:
        string:
```


theHamsta (approved):

theHamsta:
Thanks 😊!

PR #126, opened 2021-10-03
Update Ruby block selections

npezza93:
Outer was selecting the entire top line instead of just the do

Example:
```ruby
foo.tap do
  bar
end
```

Before:
Selecting @block.outer selecting all of the following
After:
Selects from do to end. Does not include `foo.tap `


theHamsta (approved):
Thanks for this awesome contribution!

theHamsta:
@sevanteri can you try to update the fennel queries? They seem to be outdated.

PR #127, opened 2021-10-05
Small Fennel fixes

sevanteri:
Not actually sure why the GH action failed with the error it did, but here's something.

Again, there was [some tree-sitter changes for Fennel](https://github.com/TravonteD/tree-sitter-fennel/pull/31); hopefully these fixes are enough.


theHamsta:
Seems to be called `(binding)` now.

PR #132, opened 2021-10-27
Added lookbehind functionality

fabiocaruso:
Lookbehind is like lookahead but in the other direction. This can be useful if you want to select the function name from the function body with a custom query. If lookahead and lookbehind are configured as true, lookahead is dominant. As a further improvement i suggest that lookahead and lookbehind should be assigned to each keymap individually, because this will allow bindings like vim's integrated f for search forward and F for search backward.

### DIFF
--- a/lua/nvim-treesitter-textobjects.lua
+++ b/lua/nvim-treesitter-textobjects.lua
@@ -30,6 +30,7 @@ function M.init()
           return M.has_textobjects(lang) or has_some_textobject_mapping(lang)
         end,
         lookahead = false,
+        lookbehind = false,
         keymaps = {},
       },
       move = {
--- a/lua/nvim-treesitter/textobjects/shared.lua
+++ b/lua/nvim-treesitter/textobjects/shared.lua
@@ -52,6 +52,9 @@ function M.textobject_at_point(query_string, pos, bufnr, opts)
   local lookahead_match_length
   local lookahead_largest_range
   local lookahead_earliest_start
+  local lookbehind_match_length
+  local lookbehind_largest_range
+  local lookbehind_earliest_start
 
   for _, m in pairs(matches) do
     if m.node and ts_utils.is_in_node_range(m.node, row, col) then
@@ -86,6 +89,20 @@ function M.textobject_at_point(query_string, pos, bufnr, opts)
           lookahead_earliest_start = start_byte
         end
       end
+    elseif opts.lookbehind then
+      local start_line, start_col, start_byte = m.node:start()
+      if start_line < row or start_line == row and start_col < col then
+        local length = ts_utils.node_length(m.node)
+        if
+          not lookbehind_earliest_start
+          or lookbehind_earliest_start < start_byte
+          or (lookbehind_earliest_start == start_byte and lookbehind_match_length > length)
+        then
+          lookbehind_match_length = length
+          lookbehind_largest_range = m
+          lookbehind_earliest_start = start_byte
+        end
+      end
     end
   end
 
@@ -99,6 +116,8 @@ function M.textobject_at_point(query_string, pos, bufnr, opts)
     end
   elseif lookahead_largest_range then
     return bufnr, { lookahead_largest_range.node:range() }, lookahead_largest_range.node
+  elseif lookbehind_largest_range then
+    return bufnr, { lookbehind_largest_range.node:range() }, lookbehind_largest_range.node
   end
 end
 
--- a/.github/workflows/check_query_files.yml
+++ b/.github/workflows/check_query_files.yml
@@ -18,10 +18,12 @@ jobs:
         run: npm i -g tree-sitter
 
       - name: Prepare
+        env:
+          NVIM_TAG: v0.5.1
         run: |
           sudo apt-get update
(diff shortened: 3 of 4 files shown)


lua/nvim-treesitter-textobjects.lua
```diff
@@ -30,6 +30,7 @@ function M.init()
           return M.has_textobjects(lang) or has_some_textobject_mapping(lang)
         end,
         lookahead = false,
+				lookbehind = false,
```
theHamsta, line 33: stylua

.github/workflows/update-readme.yml
```diff
@@ -17,13 +17,15 @@ jobs:
         run: |
           sudo apt-get update
           sudo add-apt-repository universe
-          wget https://github.com/neovim/neovim/releases/download/nightly/nvim.appimage
+          wget https://github.com/neovim/neovim/releases/download/v0.5.0/nvim.appimage
```
theHamsta, line 20: can you create a separate PR for that. Please also use v0.5.1

lua/nvim-treesitter/textobjects/shared.lua
```diff
@@ -86,6 +89,20 @@ function M.textobject_at_point(query_string, pos, bufnr, opts)
           lookahead_earliest_start = start_byte
         end
       end
+    elseif opts.lookbehind then
```
theHamsta, line 92: it appear that throught the `else` lookahead and lookbehind are exclusive. I don't think they need to be. E.g. when nothing is found lookahead, it could look behind. If they are intended to be exclusive it should also be only one config key to configure them, but I'd perfer to allow to have them both.

lua/nvim-treesitter/textobjects/shared.lua
```diff
@@ -99,6 +116,8 @@ function M.textobject_at_point(query_string, pos, bufnr, opts)
     end
   elseif lookahead_largest_range then
     return bufnr, { lookahead_largest_range.node:range() }, lookahead_largest_range.node
+  elseif lookbehind_largest_range then
+    return bufnr, { lookbehind_largest_range.node:range() }, lookbehind_largest_range.node
   end
 end
```
theHamsta, line 123: can you document your changes in the README?

theHamsta (changes requested):
Really nice addition! I really appreciate when our plugins motivate new contributors to become active! I just added a few minor remarks.

lua/nvim-treesitter/textobjects/shared.lua
theHamsta: and sorry that it took me so long to review. I was quite busy at work (GTC arriving)

.github/workflows/update-readme.yml
theHamsta: Don't worry. We don't have to overcomplicate things. Just update the version to v0.5.1

.github/workflows/update-readme.yml
```diff
@@ -17,13 +17,15 @@ jobs:
...
           chmod u+x nvim.appimage
           mkdir -p ~/.local/share/nvim/site/pack/nvim-treesitter-textobject/start
           ln -s $(pwd) ~/.local/share/nvim/site/pack/nvim-treesitter-textobject/start
           mkdir -p ~/.local/share/nvim/site/pack/nvim-treesitter/start
           cd ~/.local/share/nvim/site/pack/nvim-treesitter/start
           git clone https://github.com/nvim-treesitter/nvim-treesitter.git
+          cd nvim-treesitter
+          git checkout 0.5-compat
```
theHamsta, line 28: Why checkout 0.5 compat? It will automatically checkout the current branch. PR should be tested and not current 0.5-compat

fabiocaruso: I didn't changed this file, it's from an older commit. I think it's the only diff of 0.5-compat and master and since we merge it from 0.5-compat to master git appended this commit to this pr.

lua/nvim-treesitter/textobjects/shared.lua
fabiocaruso: No problem, i don't have much time either (school). Thank you for your time reviewing this pr and developing this plugin! It's great!

lua/nvim-treesitter/textobjects/shared.lua
fabiocaruso: It's not intended. I changed it.

.github/workflows/update-readme.yml
theHamsta: I see. Yes this was my fault then. Can you rebase your changes on master then? `git rebase -i origin/master` where origin is our remote. You'll probably need to drop the commit that added this.

Do you want me to fix this?

fabiocaruso: Yes i think it's better you fix this. Thanks for your help & patience.

theHamsta: That's great! You then fixed three issues in this PR. Not bad. Thank you!

theHamsta (approved):

theHamsta:
Thank you again for this nice addition @fabiocaruso !

PR #134, opened 2021-10-31
Improve ruby text object performance

npezza93:
Greatly improves the performance of ruby textobjects

Before:
705-715ms

After:
480-500ms

Commenting out this entire file brings performance down too 420-430ms. Any tips to get a few more ms out of these queries?

### DIFF
--- a/queries/ruby/textobjects.scm
+++ b/queries/ruby/textobjects.scm
@@ -1,193 +1,40 @@
-;; @functions
+; @functions
+ ((method . name: (identifier) (method_parameters)? . (_) @function.inner (_)? @function.end .)
+   (#make-range! "function.inner" @function.inner @function.end)) @function.outer
+ ((singleton_method . name: (identifier) (method_parameters)? . (_) @function.inner (_)? @function.end .)
+   (#make-range! "function.inner" @function.inner @function.end)) @function.outer
+
+; @blocks
+((block (block_parameters)? . (_) @block.inner (_)? @block.inner.end .)
+  (#make-range! "block.inner" @block.inner @block.inner.end)) @block.outer
+((do_block (block_parameters)? . (_) @block.inner (_)? @block.inner.end .)
+  (#make-range! "block.inner" @block.inner @block.inner.end)) @block.outer
+
+; @classes
 (
-  [
-    (method
-      . name: (identifier)
-      !parameters
-      . (_) @_start (_) @_end .)
-
-    (method
-      . name: (identifier)
-      . parameters: (method_parameters)
-      . (_) @_start (_) @_end .)
-
-    (singleton_method
-      . object: (_)
-      . name: (identifier)
-      !parameters
-      . (_) @_start (_) @_end .)
-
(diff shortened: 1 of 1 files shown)


theHamsta:
The best thing to do would be to port `find_best_match` to use the native query API (`iter_group_results` is horribly slow because it construct and deconstruct to many tables) and use `find_best_match` to simplify the get textobjects call.
https://github.com/theHamsta/nvim-treesitter/blob/3e139eff64972096e23e3c6fcd7fa8ce1f147118/lua/nvim-treesitter/query.lua#L232-L232

theHamsta:
@TravonteD @semanticart could you Ruby experts test those queries? I've never used that language.

theHamsta:
@npezza93 You can check the query with tree-sitter CLI to see what speed of light we can expect. When nvim-treesitter-textobjects would use the native query API directly it would have roughly that performance.

npezza93:
@theHamsta Happy to. Give me a bit to get that set up and ill report back

antonk52:
@theHamsta sorry, never worked with ruby. Though happy to test ecma related changes next time 🙂

theHamsta:
Just saw you in the diffs. Sorry for the annoyance :sweat_smile:

npezza93:
@theHamsta How do i go about telling the cli to use this textobjects.scm file? I have the cli working but not sure how to tell it to look at this file.

npezza93:
@theHamsta  Nevermind. Helps if i use the query command haha.

```
❯ time tree-sitter query ~/.local/share/nvim/plugged/nvim-treesitter-textobjects/queries/ruby/textobjects.scm app/models/account.rb

tree-sitter query  app/models/account.rb  0.05s user 0.01s system 96% cpu 0.065 total
```

theHamsta:
At some point we also need to achieve the 50ms in this plugin. But thanks that you showed that there's also on CLI a huge difference.

npezza93:
Heres the hyperfine version for a little more accurate numbers
```
❯ hyperfine 'tree-sitter query ~/Documents/nvim-treesitter-textobjects/queries/ruby/textobjects.scm app/models/account.rb'
Benchmark 1: tree-sitter query ~/Documents/nvim-treesitter-textobjects/queries/ruby/textobjects.scm app/models/account.rb
  Time (mean ± σ):      48.7 ms ±   3.5 ms    [User: 44.0 ms, System: 3.0 ms]
  Range (min … max):    41.3 ms …  58.4 ms    56 runs


❯ hyperfine 'tree-sitter query ~/Documents/nvim-treesitter-textobjects/queries/ruby/textobjects.scm app/models/account.rb'
Benchmark 1: tree-sitter query ~/Documents/nvim-treesitter-textobjects/queries/ruby/textobjects.scm app/models/account.rb
  Time (mean ± σ):     209.2 ms ±  10.8 ms    [User: 201.8 ms, System: 5.1 ms]
  Range (min … max):   194.4 ms … 232.7 ms    12 runs
```

PR #136, opened 2021-11-26
Rust and Java comment text objects and fix trailing whitespace.

caojoshua:
* Rust and Java comment text objects
* Fix trailing whitespace in edited files

### DIFF
--- a/queries/java/textobjects.scm
+++ b/queries/java/textobjects.scm
@@ -38,12 +38,13 @@
 (method_invocation (argument_list) @call.inner)
 
 ;; parameters
-(formal_parameters    
+(formal_parameters
   "," @_start .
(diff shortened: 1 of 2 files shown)


theHamsta (approved):
Thank you!

caojoshua:
Think the check timed out. It was checking elm textobjects, so shouldn't be related to this change. I'll let the maintainers decide if they want to override or take some other action.

theHamsta:
This is might be a regression due to the update of nightly nvim to tree-sitter v0.20.1 @maxxnino a timeout of query loading for elm that didn't happen before.

maxxnino:
yeah, i think because of this. Because this never happen before, until recently

PR #137, opened 2021-11-27
Adds relevant text objects for Elm

bwells:
Adds
  * comment.outer
  * conditional.inner
  * conditional.outer
  * function.inner
  * function.outer
  * parameter.inner
  * parameter.outer

I tried to add @block.inner and @block.outer for let/in expressions, but as of yet the query necessary given the available AST structure is beyond me. 

The behavior of @function.outer in particular is to my tastes, but may run counter to behavior in other languages. 

The full module level function block in Elm is composed of up to three parts, the function itself and one or both of a type annotation and doc comment above the function. The current implementation will grab these for @function.outer as greedily as possible. 

I feel selecting the type annotation is certainly correct and matches the existing behavior of python including decorators in @function.outer. I feel the doc comment should also be includes as its seems uncommon I would want to select an entire function and type annotation to move or delete without also wanting the doc comment, but it's obviously not strictly the function itself. I'd be very open to a mapping that allows combining @comment.outer + @function.outer, but I'm unaware of how to increase the selection in visual mode with further text objects.

### DIFF
--- a/queries/elm/textobjects.scm
+++ b/queries/elm/textobjects.scm
@@ -1,0 +1,58 @@
+;; Functions
+; top level function with type annotation and doc comment
+(
+  (module_declaration)
+  (block_comment) @function.outer.start
+  .
+  (type_annotation)
+  .
+  (value_declaration
+    body: (_)? @function.inner) @function.outer
+)
+
+; top level function with type annotation
+(
+  (module_declaration)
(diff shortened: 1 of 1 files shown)


theHamsta (approved):
LGTM! We can tackle the more challenging queries later.

PR #163, opened 2022-01-17
fix(java): use line_comment/block_comment instead of comment

theHamsta:
This fixes the nightly running CI that failed due to the update of the Java parser.

### DIFF
--- a/queries/java/textobjects.scm
+++ b/queries/java/textobjects.scm
@@ -45,6 +45,9 @@
 (formal_parameters
   . (formal_parameter) @parameter.inner
   . ","? @_end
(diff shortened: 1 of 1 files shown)


queries/java/textobjects.scm
```diff
@@ -45,6 +45,9 @@
 (formal_parameters
   . (formal_parameter) @parameter.inner
   . ","? @_end
- (#make-range! "parameter.outer" @_start @parameter.inner))
+ (#make-range! "parameter.outer" @_end @parameter.inner))
```
theHamsta, line 48: `@_start` was an invalid capture that is not present in the query.

clason (approved):

PR #165, opened 2022-01-18
fix(lua): adapt queries to upstream switch

clason:
nvim-treesitter switched to a different Lua parser in https://github.com/nvim-treesitter/nvim-treesitter/pull/2272, which requires adaptation of queries

@MunifTanjim


theHamsta (approved):
Nice, thank you! We get rid of a lot of hacks to make it working with the old parser!.

PR #182, opened 2022-02-07
Change config for swap and move to take a table of capture groups to increase configurability

HungryJoe:
While I like this plugin a lot, the biggest complaint that I have (and have read) about it is that it requires a lot of keybinds to use effectively since they must be one per-capture-group per-action. Furthermore, many people (myself included) don't really care about the specific capture group being acted upon; rather, they want to act upon the first or closest capture group that makes sense for the action they're trying.

To be more concrete: I work a lot in Python. What I really want from this plugin are two things: (1) to easily move between major, meaningful parts of a file, and (2) to easily swap elements of lists (of arguments, dictionary entries, etc.). The over-broadness of `@parameter`, as pointed out in PR #86, is evidence that I am not alone in this desire, at least for the `swap` action/submodule.

My proposed solution is to change the values in the keybind maps for `move` and `swap` from single capture groups to lists of them. For `move`, the capture group chosen will be the one that starts (or ends) nearest to the cursor; for `swap`, it will be the one under the cursor that comes first in the configured list.

I'm new to Lua, so please be kind about criticisms relating to the quality of the code.

### DIFF
--- a/lua/nvim-treesitter/textobjects/attach.lua
+++ b/lua/nvim-treesitter/textobjects/attach.lua
@@ -10,18 +10,18 @@ function M.make_attach(normal_mode_functions, submodule)
     lang = lang or parsers.get_buf_lang(bufnr)
 
     for _, function_call in pairs(normal_mode_functions) do
-      for mapping, query in pairs(config[function_call] or {}) do
+      for mapping, config_queries in pairs(config[function_call] or {}) do
         if not queries.get_query(lang, "textobjects") then
-          query = nil
+          config_queries = nil
         end
-        if query then
+        if config_queries then
           local cmd = ":lua require'nvim-treesitter.textobjects."
             .. submodule
             .. "'."
             .. function_call
-            .. "('"
-            .. query
-            .. "')<CR>"
+            .. "({'"
+            .. table.concat(config_queries, "','")
+            .. "'})<CR>"
           api.nvim_buf_set_keymap(bufnr, "n", mapping, cmd, { silent = true, noremap = true })
         end
       end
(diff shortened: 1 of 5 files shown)


lua/nvim-treesitter/textobjects/attach.lua
```diff
@@ -10,18 +10,18 @@ function M.make_attach(normal_mode_functions, submodule)
     lang = lang or parsers.get_buf_lang(bufnr)
 
     for _, function_call in pairs(normal_mode_functions) do
-      for mapping, query in pairs(config[function_call] or {}) do
+      for mapping, config_queries in pairs(config[function_call] or {}) do
         if not queries.get_query(lang, "textobjects") then
-          query = nil
+          config_queries = nil
         end
-        if query then
```
theHamsta, line 17: This looks very nice! Would it be possible to leave the old config working? E.g. by only appling the table.concat if it is a table `type(config_queries) == 'string' and config_queries or table.concat(config_queries)`

HungryJoe: This is a good idea; I've implemented it in the `move` and `swap` functions themselves rather than at all of the call-sites.

theHamsta: But how will `table.concat` work then?

HungryJoe: Oops, sorry. I pushed another commit to fix it, which also fixed LSP interop since it uses the same `make_attach` function as `move` and `swap`.

theHamsta:
Can you rebase? Is this ready to merge?

theHamsta (approved):

Feel-ix-343:
Does this still work?

kiyoon:
@Feel-ix-343 I think you need to put it in a "query key of the table.

`{ query = { "@function.inner", "@parameter.inner" } }`

PR #185, opened 2022-02-22
Java: Add arguments support and fix the bug when selecting first parameter by @parameter.outer

qsdrqs:
Hi, I added two commits for java textobjects. The first commit adds arguments support when using @parameter.inner and outer. It looks like
image
and
image
I have tested the textobjects function and swap function and it works fine.

In the second commit, I fix a bug when selecting the first parameter by @parameter.outer, before this commit, using @parameter.outer to select the first parameter will select the second parameter like
image
After this commit, it will successfully select the correct area like
image

Thank you for your review!


theHamsta (approved):
Looks good! Thank you a lot!

qsdrqs:
Sorry, I have no idea why this check failed. Could you please give me some hints to solve this problem?

PR #186, opened 2022-02-23
fix(fennel): each_clause does no longer exist

theHamsta:
See https://github.com/TravonteD/tree-sitter-fennel/pull/34/files

 The query tries to select from `[` to everything until ")"`. This seems to be be the behavior before. Though I would either select the loop body or the bindings....


theHamsta:
@sevanteri what is the indented behavior?

sevanteri:
The intended behavior was to select the whole `(each ...)` expression with `@loop.outer` and only the body with `@loop.inner`.

PR #196, opened 2022-04-08
fix: inner function should not include braces

telemachus:


telemachus:
In this pull request, I've updated bash, c (and therefore also c++ and cuda), dart, go, java, javascript, and php. If there are other brace-languages that I missed, please let me know.

This should close #112 and, I think, the underlying problem in #102. Also, these queries may get revised if nvim-treesitter and treesitter itself support `*` and `+` fully for captures.

theHamsta (approved):

tom-anders:
I think we have the same problem for `@conditional.inner` and `@loop.inner`

telemachus:
> I think we have the same problem for `@conditional.inner` and `@loop.inner`

Yup, that makes sense. I also missed Rust. (I thought of it only after this was pulled.)

I assume that something like what I did here with `#make-range` should work for conditions and loops too. I can't do it immediately, but over the next couple of weekends, I should be able to put together another pull request.

tom-anders:
Cool, thanks!

PR #203, opened 2022-04-16
[Elixir] Adds support for anonymous functions

thmsmlr:
This patch extends @function.inner and @function.outer to also support
anonymous functions of the style

```
fn(x) ->
  x * x
end
```


theHamsta (approved):
Thank you!

PR #207, opened 2022-04-28
Exclude curly braces on @function.inner in JavaScript arrow functions

melkster:
#196 excludes curly braces for `@function.inner`, but not for arrow functions in JavaScript.

I'm completely new to treesitter queries and Scheme so feel free to critique my solution.

### DIFF
--- a/queries/javascript/textobjects.scm
+++ b/queries/javascript/textobjects.scm
@@ -17,7 +17,11 @@
   (function_declaration) @function.outer) @function.outer.start
 
 (arrow_function
-  body: (_) @function.inner) @function.outer
+  . body: (_) @function.inner) @function.outer
+
+(arrow_function
+  body: (statement_block . "{" . (_) @_start @_end (_)? @_end . "}"
+ (#make-range! "function.inner" @_start @_end)))
 
 (method_definition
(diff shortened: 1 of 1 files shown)


queries/javascript/textobjects.scm
```diff
@@ -17,7 +17,11 @@
   (function_declaration) @function.outer) @function.outer.start
 
 (arrow_function
-  body: (_) @function.inner) @function.outer
+  . body: (_) @function.inner) @function.outer
+
+(arrow_function
+  body: (statement_block . "{" . (_) @_start @_end (_)? @_end . "}"
```
theHamsta, line 23: I think you can remove the first `@_end` `#make-range` can handle the case that `@_end` wasn't found.

theHamsta: but it shouldn't do any harm

theHamsta (approved):

queries/javascript/textobjects.scm
melkster: Sorry I haven't had the time to respond. Thanks for merging my PR! Feel free to remove the redundant `@_end` in the future!

PR #216, opened 2022-06-07
feat(html): add `@attribute.inner`, `outer`

bennypowers:
`daa` - delete around attribute (i.e. delete entire attribute)
`cia` - change inner attribute (i.e. attr value)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,6 +5,8 @@ You can define text objects based on nodes of the grammar by adding queries in `
 Each capture group can be declared as `inner` or `outer`.
 
 ```
+@attribute.inner
+@attribute.outer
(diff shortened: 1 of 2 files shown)


theHamsta:
Can you add this to CONTRIBUTING.md

bennypowers:
I've added `@attribute.{inn,out}er` to CONTRIBUTING.md

NB that `@attribute.outer` will leave behind the surrounding space. this is perhaps unexpected:

```html
<x-l thing="1" other="2">
            ^ cursor here
```
Then `daa`
```html
<x-l  other="2">
```
Note: 2 spaces

theHamsta:
Maybe you can fix this in a follow-up PR? I think this is ready to merge.

PR #223, opened 2022-07-04
Add support for hack lang @function.outer

mihirgogate:
Will add support for @function.inner, @class.outer, @class.inner in a subsquent diff


theHamsta:
Looks good! Thank you!

theHamsta (approved):

PR #224, opened 2022-07-05
Improve @function.inner for scala

rminehan:
Previously `@function.inner` would only capture function bodies wrapped in braces:

```scala
// Would capture
def foo: Int = {
  ...
}

// Wouldn't capture
def foo: Int = 3 * 10
```

### DIFF
--- a/queries/scala/textobjects.scm
+++ b/queries/scala/textobjects.scm
@@ -5,7 +5,7 @@
   body: (template_body)? @class.inner) @class.outer
 
 (function_definition
-  body: (block) @function.inner) @function.outer
+  body: (expression) @function.inner) @function.outer
 
 (parameter
   name: (identifier) @parameter.inner) @parameter.outer


queries/scala/textobjects.scm
```diff
@@ -5,7 +5,7 @@
   body: (template_body)? @class.inner) @class.outer
 
 (function_definition
-  body: (block) @function.inner) @function.outer
+  body: (expression) @function.inner) @function.outer
```
theHamsta, line 8: why not both?

theHamsta: `[(block) (expression)]`

theHamsta: @rminehan

theHamsta:
Thank you for your contribution!

theHamsta (approved):

PR #233, opened 2022-07-30
Register key mappings to which-key

### LINKED ISSUE #232: Allow to provide description for keymaps
I really enjoy this extension, really improves my dev workflow. I rely on muscular memory a lot, but there are certain key combinations that I don't use that often and that I don't remember. To work around that I use which key, that allows me to see the bindings I have and what they map to. Wich key is able to extract the description provided to a keymap, which makes it awesome. It will be equally awesome if I could have descriptions for my text objects. 

I want to be able to provide a description to each key mapping and that the plugin uses that description when creating the maps

I could map the keys myself, but I need to know which functions to call and with wich params, so at least adding that to the docs will help. 

**Additional context**

I am an user of nvim wich key and other tools that also rely on keymap description like telescope

theHamsta:
Fixes #232

@danielo515 are you using https://github.com/folke/which-key.nvim ? This should set description for this plugin. How can I register for telescope's which_key? Do want to additionally want to overwrite the automatically generated descriptions?

I don't think there is a generic way to set a description for a key mapping. ~~@folke maybe it would be an idea to add a "description" key to `opts` of `nvim_buf_add_keymap` that can be retrieved by new API function? It is difficult for third party plugins to annotate their keybindings that is not taylored to one specific which-key plugin ~~ (`nvim_buf_add_keymap` already has `desc`)

grafik

### DIFF
--- a/lua/nvim-treesitter/textobjects/attach.lua
+++ b/lua/nvim-treesitter/textobjects/attach.lua
@@ -1,7 +1,6 @@
 local configs = require "nvim-treesitter.configs"
 local parsers = require "nvim-treesitter.parsers"
 local queries = require "nvim-treesitter.query"
-local api = vim.api
 local M = {}
 
 function M.make_attach(normal_mode_functions, submodule)
@@ -10,26 +9,23 @@ function M.make_attach(normal_mode_functions, submodule)
     lang = lang or parsers.get_buf_lang(bufnr)
 
     for _, function_call in pairs(normal_mode_functions) do
+      local description = function_call:gsub("_", " ")
       for mapping, config_queries in pairs(config[function_call] or {}) do
         if not queries.get_query(lang, "textobjects") then
           config_queries = nil
         end
+        local desc
+        if type(config_queries) == "table" then
+          desc = config_queries.desc
+          config_queries = config_queries.query
+        end
+        if not desc then
+          desc = description:gsub("^%l", string.upper) .. " " .. config_queries
+        end
         if config_queries then
-          local config_query_str
-          if type(config_queries) == "string" then
-            config_query_str = "'" .. config_queries .. "'"
-          else
-            config_query_str = "{'" .. table.concat(config_queries, "','") .. "'}"
-          end
-
-          local cmd = ":lua require'nvim-treesitter.textobjects."
-            .. submodule
-            .. "'."
-            .. function_call
-            .. "("
-            .. config_query_str
-            .. ")<CR>"
-          api.nvim_buf_set_keymap(bufnr, "n", mapping, cmd, { silent = true, noremap = true })
+          vim.keymap.set("n", mapping, function()
+            require("nvim-treesitter.textobjects." .. submodule)[function_call](config_queries)
+          end, { buffer = bufnr, silent = true, remap = false, desc = desc })
         end
       end
     end
@@ -46,8 +42,7 @@ function M.make_detach(normal_mode_functions, submodule)
         query = nil
       end
       if query then
-        api.nvim_buf_del_keymap(bufnr, "o", mapping)
-        api.nvim_buf_del_keymap(bufnr, "v", mapping)
+        vim.keymap.del({ "o", "x" }, mapping, { buffer = bufnr })
       end
     end
     for _, function_call in pairs(normal_mode_functions) do
@@ -58,7 +53,7 @@ function M.make_detach(normal_mode_functions, submodule)
           query = nil
         end
         if query then
-          api.nvim_buf_del_keymap(bufnr, "n", mapping)
+          vim.keymap.del("n", mapping, { buffer = bufnr })
         end
       end
     end
--- a/lua/nvim-treesitter/textobjects/select.lua
+++ b/lua/nvim-treesitter/textobjects/select.lua
@@ -129,14 +129,21 @@ function M.attach(bufnr, lang)
   lang = lang or parsers.get_buf_lang(buf)
 
   for mapping, query in pairs(config.keymaps) do
+    local desc
+    if type(query) == "table" then
+      desc = query.desc
+      query = query.query
+    end
+    if not desc then
+      desc = "Select textobject " .. query
+    end
     if not queries.get_query(lang, "textobjects") then
       query = nil
     end
     if query then
-      local cmd_o = ":lua require'nvim-treesitter.textobjects.select'.select_textobject('" .. query .. "', 'o')<CR>"
-      api.nvim_buf_set_keymap(buf, "o", mapping, cmd_o, { silent = true, noremap = true })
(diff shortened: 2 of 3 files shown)


theHamsta:
This is inaccurate in a way that we dynamically register the keymaps per file type and only when tree-sitter is attached.

theHamsta:
The description is inaccurate in a way that we only set the bindings dynamically when tree-sitter is enabled for the buffer. The which-key hints will however appear for all buffers. Not just the ones for which the keybindings were enabled

theHamsta:
Ohh! `nvim_buf_set_keymap` already has a `desc` key!

theHamsta:
grafik

lua/nvim-treesitter/textobjects/attach.lua
```diff
@@ -29,7 +33,13 @@ function M.make_attach(normal_mode_functions, submodule)
             .. "("
             .. config_query_str
             .. ")<CR>"
-          api.nvim_buf_set_keymap(bufnr, "n", mapping, cmd, { silent = true, noremap = true })
```
theHamsta, line 32: if desired I could allow mapping to be also be optionally a table and then access mapping.desc

danielo515:
> Ohh! `nvim_buf_set_keymap` already has a `desc` key!

Yes, that is the option I was referring to

lua/nvim-treesitter/textobjects/attach.lua
```diff
@@ -10,6 +10,10 @@ function M.make_attach(normal_mode_functions, submodule)
     lang = lang or parsers.get_buf_lang(bufnr)
 
     for _, function_call in pairs(normal_mode_functions) do
+      local description = function_call:gsub("_", " ")
+      if type(function_call) == "table" then
```
danielo515, line 14: Should this type check happen before the gsub call? If the function is not a string but a table I expect this to fail.

lua/nvim-treesitter/textobjects/attach.lua
danielo515: You already did right? That is just perfect

theHamsta: No, not yet. At the moment it will just generate a description for all commands but you can't set your own. Is setting your own description still needed?

lua/nvim-treesitter/textobjects/attach.lua
theHamsta: You're right! Thank you!

lua/nvim-treesitter/textobjects/attach.lua
danielo515: Yes, setting my own description is why I opened the issue in the first place. I can see a lot of people not being happy with the auto generated ones.

theHamsta:
Can you test now?

grafik

Config like this 
grafik

lua/nvim-treesitter/textobjects/attach.lua
```diff
@@ -10,10 +10,24 @@ function M.make_attach(normal_mode_functions, submodule)
     lang = lang or parsers.get_buf_lang(bufnr)
 
     for _, function_call in pairs(normal_mode_functions) do
+      local description
+      if type(function_call) == "table" then
+        function_call, description = unpack(function_call)
```
kyazdani42, line 15: is the function call an "array" table ? If it is a record like table this will not work.

lua/nvim-treesitter/textobjects/attach.lua
```diff
@@ -29,7 +43,7 @@ function M.make_attach(normal_mode_functions, submodule)
             .. "("
             .. config_query_str
             .. ")<CR>"
-          api.nvim_buf_set_keymap(bufnr, "n", mapping, cmd, { silent = true, noremap = true })
+          vim.api.nvim_buf_set_keymap(bufnr, "n", mapping, cmd, { silent = true, noremap = true, desc = desc })
```
kyazdani42, line 46: why not use `vim.keymap.set` ?

lua/nvim-treesitter/textobjects/select.lua
```diff
@@ -48,14 +48,22 @@ function M.attach(bufnr, lang)
...
     if not queries.get_query(lang, "textobjects") then
       query = nil
     end
     if query then
       local cmd_o = ":lua require'nvim-treesitter.textobjects.select'.select_textobject('" .. query .. "', 'o')<CR>"
-      api.nvim_buf_set_keymap(buf, "o", mapping, cmd_o, { silent = true, noremap = true })
+      api.nvim_buf_set_keymap(buf, "o", mapping, cmd_o, { silent = true, noremap = true, desc = desc })
       local cmd_x = ":lua require'nvim-treesitter.textobjects.select'.select_textobject('" .. query .. "', 'x')<CR>"
-      api.nvim_buf_set_keymap(buf, "x", mapping, cmd_x, { silent = true, noremap = true })
+      api.nvim_buf_set_keymap(buf, "x", mapping, cmd_x, { silent = true, noremap = true, desc = desc })
```
kyazdani42, line 66: same here

kyazdani42 (approved):

lua/nvim-treesitter/textobjects/select.lua
```diff
@@ -48,14 +48,22 @@ function M.attach(bufnr, lang)
...
+      query = query.query
+    end
+    if not desc then
+      desc = "Select textobject " .. query
+    end
     if not queries.get_query(lang, "textobjects") then
       query = nil
     end
     if query then
       local cmd_o = ":lua require'nvim-treesitter.textobjects.select'.select_textobject('" .. query .. "', 'o')<CR>"
```
clason, line 63: Same here: we should make this a pure Lua mapping; that's the whole point of this API.

lua/nvim-treesitter/textobjects/attach.lua
danielo515: I think the intention is to have a record like table, with a query and a description keys

theHamsta: It's not used at the moment, I intended to have `{function, description}`. It is not used atm. Better to remove?

theHamsta: I removed this feature which was intended for internal documentation of the functions.

danielo515: But why? I thought the entire PR was to allow end users to customize the description of the created keybindings.

theHamsta: You don't understand. There were three sources for descriptions 1) auto-generated, 2) set by the plugin 3) set by the user. It would always prefer the higher number. 2) didn't exist. So it's 1) auto-generated, 2) set by the user

theHamsta:
@kyazdani42 good to go?

PR #235, opened 2022-07-30
include surrounding whitespace

AckslD:
This is an initial attempt to handle preceding/succeeding whitespace of textobjects and I'm looking for feedback. Initially mainly about the behaviour, which is opt-in. I think the current implementation can be made much better but that's a second question and I would first like to know if this makes sense :)

What this change does is that if the user enables it, it expands any textobject to include preceding or succeeding whitespace. It always tries to include any preceeding whitespace and if there is none instead tries to include succeeding whitespace (but never both). The ides is that it would work much like eg the built-in `aw`, `ap` etc. See the following screencast for a more concrete view:

https://user-images.githubusercontent.com/23341710/181994568-52c018b8-861d-4ec9-a2e9-847e88100e88.mp4

The implementation is currently very simple and probably quite inefficient since it just tries to walk one character backward of forward and see if there is more whitespace to consume. Maybe one could use a regex or so instead. I also considered using `ts_utils.get_previous_node` and use the end of that for the start of the current node, but that doesn't work so well if you're for example selecting the first argument to a function.

Related #69

### DIFF
--- a/lua/nvim-treesitter/textobjects/select.lua
+++ b/lua/nvim-treesitter/textobjects/select.lua
@@ -8,49 +8,15 @@ local ts_utils = require "nvim-treesitter.ts_utils"
 
 local M = {}
 
-function M.select_textobject(query_string, keymap_mode)
-  local lookahead = configs.get_module("textobjects.select").lookahead
-  local lookbehind = configs.get_module("textobjects.select").lookbehind
-  local include_surrounding_whitespace = configs.get_module("textobjects.select").include_surrounding_whitespace
-  local bufnr, textobject =
-    shared.textobject_at_point(query_string, nil, nil, { lookahead = lookahead, lookbehind = lookbehind })
-  if textobject then
-    if include_surrounding_whitespace then
-      textobject = M.include_surrounding_whitespace(bufnr, textobject)
-    end
-    ts_utils.update_selection(bufnr, textobject, M.detect_selection_mode(query_string, keymap_mode))
+local function get_char_after_position(bufnr, row, col)
+  if row == nil then
+    return nil
   end
+  return vim.api.nvim_buf_get_text(bufnr, row, col, row, col + 1, {})[1]
 end
 
-function M.include_surrounding_whitespace(bufnr, textobject)
-  local start_row, start_col, end_row, end_col = unpack(textobject)
-  local extended = false
-  while M.is_whitespace_after(bufnr, end_row, end_col) do
-    extended = true
-    end_row, end_col = M.next_position(bufnr, end_row, end_col, true)
-  end
-  if extended then
-    -- don't extend in both directions
-    return { start_row, start_col, end_row, end_col }
-  end
-  local next_row, next_col = M.next_position(bufnr, start_row, start_col, false)
-  while M.is_whitespace_after(bufnr, next_row, next_col) do
-    start_row = next_row
-    start_col = next_col
-    next_row, next_col = M.next_position(bufnr, start_row, start_col, false)
-  end
-  return { start_row, start_col, end_row, end_col }
-end
-
-function M.is_whitespace_after(bufnr, row, col)
-  if col == #M.get_line(bufnr, row) then
-    if row == vim.api.nvim_buf_line_count(bufnr) then
-      return false
-    end
-    row = row + 1
-    col = 0
-  end
-  local char = M.get_char_after_position(bufnr, row, col)
+local function is_whitespace_after(bufnr, row, col)
+  local char = get_char_after_position(bufnr, row, col)
   if char == nil then
     return false
   end
@@ -60,15 +26,12 @@ function M.is_whitespace_after(bufnr, row, col)
   return string.match(char, "%s")
 end
 
-function M.get_char_after_position(bufnr, row, col)
-  if row == nil then
-    return nil
-  end
-  return vim.api.nvim_buf_get_text(bufnr, row, col, row, col + 1, {})[1]
+local function get_line(bufnr, row)
+  return vim.api.nvim_buf_get_lines(bufnr, row, row + 1, false)[1]
 end
 
-function M.next_position(bufnr, row, col, forward)
-  local max_col = #M.get_line(bufnr, row)
+local function next_position(bufnr, row, col, forward)
+  local max_col = #get_line(bufnr, row)
   local max_row = vim.api.nvim_buf_line_count(bufnr)
   if forward then
     if col == max_col then
@@ -86,7 +49,7 @@ function M.next_position(bufnr, row, col, forward)
         return nil
       end
       row = row - 1
-      col = #M.get_line(bufnr, row)
+      col = #get_line(bufnr, row)
     else
       col = col - 1
     end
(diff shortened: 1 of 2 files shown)


lua/nvim-treesitter/textobjects/select.lua
```diff
@@ -11,13 +11,86 @@ local M = {}
...
     shared.textobject_at_point(query_string, nil, nil, { lookahead = lookahead, lookbehind = lookbehind })
   if textobject then
+    if include_surrounding_whitespace then
+      textobject = M.include_surrounding_whitespace(bufnr, textobject)
+    end
     ts_utils.update_selection(bufnr, textobject, M.detect_selection_mode(query_string, keymap_mode))
   end
 end
 
+function M.include_surrounding_whitespace(bufnr, textobject)
```
theHamsta, line 25: all those added helper functions don't need to be public

theHamsta:
One detail, I don't know whether this matters but the built-in `ap` select for me following empty lines vs the previous empty lines of this PR. I don't know whether it is important to be consistent with the built-in behavior.

lua/nvim-treesitter/textobjects/select.lua
```diff
@@ -11,13 +11,86 @@ local M = {}
...
+      col = col - 1
+    end
+  end
+  return row, col
+end
+
+function M.get_line(bufnr, row)
+  return vim.api.nvim_buf_get_lines(bufnr, row, row + 1, false)[1]
+end
+
```
theHamsta, line 93: can you document this feature in the README?

lua/nvim-treesitter/textobjects/select.lua
AckslD: done

lua/nvim-treesitter/textobjects/select.lua
AckslD: done

AckslD:
good point @theHamsta! I changed it to instead prefer succeeding whitespace :+1:

theHamsta:
Thank you!

melkster:
This is really great! You're comparing `af` with `ap` which makes sense to me, but after trying this out I noticed that there's one thing that's inconsistent with `af` compared to `ap`. Pasting with `p` on the empty line between paragraphs after doing `dap` correctly places the paragraph, leaving one empty line above and below.

However, doing `daf` and pasting on an empty line between functions leaves no empty line above and two below.

Here's a simple example to illustrate:

```lua

local function f1() end

local function f2() end

local function f3() end
```

With the cursor on `f1()`'s line, doing `daf`, then `}` to move to the empty line between `f2` and `f3`, and then doing `p` leaves this:

```lua

local function f2() end
local function f1() end


local function f3() end
```

Doing the same thing with `dap` instead we would get this:

```lua

local function f2() end

local function f1() end

local function f3() end
```

Both `ap` and `af` include the whitespace below, but `ap` does it line-wise, while `af` does it character-wise. One solution would be for `af` to select the whitespace above instead. And in the case where we're doing `af` on an anonymous function, we could still select the whitespace below as we do now.

AckslD:
Hi @melkster! This is a good point and something I thought about but wasn't fully sure how to handle. Indeed, `ap` is line-wise but `af` is by default char-wise, which probably makes sense since in many languages, functions are not necessarily restricted to lines. But in some cases I'd rather have it line-wise (for example in python), which is why I made https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/70. However, I realise now that it is not completely clear to me how these two settings should work together. I'll try to think a bit more about your suggestion above.

melkster:
Oh, I didn't know about #70, is that disabled by default? I tried doing this

```lua
require('nvim-treesitter.configs').setup({
  textobjects = {
    select = {
      selection_modes = {
        ['@parameter.outer'] = 'v',
        ['@function.outer'] = 'V',
        ['@class.outer'] = '<c-v>',
      },
    },
  },
})
```

but `af` still seems to only be character-wise. I tried it in Lua, TypeScript, Python and Java.

AckslD:
How are you trying this? Note that this setting is just for operator pending mode for now and not for visual mode. So if you do `vaf` you'll still get charwise but if you do `Vaf` you'll get linewise. The difference if you do eg `daf` or similar. Does that make sense?

melkster:
@AckslD Aha! Yes, I was mainly using `vaf`. `vap` switches from character-wise to line-wise regardless of which kind of visual mode you're in, and so I just assumed that `vaf` did that as well.

With the config above `daf` works for me for functions that are in the top-level, i.e. not inside a class or similar. Perhaps this discussion is more appropriate to have in #70, but maybe it should also be possible to configure methods to also be line-wise? I.e., to support something like this:

```lua
require('nvim-treesitter.configs').setup({
  textobjects = {
    select = {
      selection_modes = {
        ['@method.outer'] = 'V',
      },
    },
  },
})
```

What do you think?

AckslD:
@melkster, what exactly is not working for methods, it doesn't use linewise mode? And it the above config snippet not working?

melkster:
@AckslD Yeah, sorry I wasn't very clear about what I meant. For methods, `daf` still is character-wise. I also tried adding the keymap `['am'] = '@method.outer'` like so:

```lua
require('nvim-treesitter.configs').setup {
  ensure_installed = 'all',
  textobjects = {
    select = {
      keymaps = {
        ['af'] = '@function.outer',
        ['if'] = '@function.inner',
        ['ac'] = '@class.outer',
        ['ic'] = '@class.inner',
        ['am'] = '@method.outer',
        ['im'] = '@method.inner',
      },
      selection_modes = {
        ['@function.outer'] = 'V',
        ['@class.outer'] = 'V',
        ['@method.outer'] = 'V',
      },
    },
  },
}
```

But `dam` does nothing (i.e. nothing happens).

AckslD:
No problem @melkster! Was just making sure I understood what you meant :) I see, this seems to be a separate issue, if how to have a text object for methods, rather than how to select linewise? For which language is this btw?

melkster:
@AckslD The languages in which I've reproduced this issue with methods are TypeScript, JavaScript, Python and C++.

I noticed now however that they seem respect my line-wise configuration, but the problem is that they include one line too many, resulting in the first line of the method below them being included in `daf` as well.

AckslD:
@melkster Yes I think this is the combination of `include_surrounding_whitespace` and `linewise` mode is not really supported atm. We might need to check what the mode is before including the surrounding whitespace to know what to do.

AckslD:
@melkster I think this should be fixed now in #258. Feel free to try it out and let me know.

melkster:
@AckslD I seem to be having the same problem after updating :thinking:

This is my config:

```lua
require('nvim-treesitter.configs').setup {
  ensure_installed = 'all',
  textobjects = {
    select = {
      enable = true,
      include_surrounding_whitespace = true,
      keymaps = {
        ['af'] = '@function.outer',
        ['if'] = '@function.inner',
      },
      selection_modes = {
        ['@function.outer'] = 'V',
        ['@class.outer'] = 'V',
      },
    },
  },
}
```
I tried in the previously mentioned languages.

AckslD:
What is the problem exactly? Is it

> but the problem is that they include one line too many,

? Using `daf`, `vaf` or `Vaf`?

melkster:
@AckslD Correct 🙂

melkster:
@AckslD Using the following config:

<details>
<summary><b>Click to expand</b></summary>

```lua
-- Ignore default config and plugins
vim.opt.runtimepath:remove(vim.fn.expand('~/.config/nvim'))
vim.opt.packpath:remove(vim.fn.expand('~/.local/share/nvim/site'))

-- Append test directory
local test_dir = vim.fn.expand('~/code-other/nvim-test-config')
vim.opt.runtimepath:append(vim.fn.expand(test_dir))
vim.opt.packpath:append(vim.fn.expand(test_dir))

-- Install packer
local install_path = test_dir .. '/pack/packer/start/packer.nvim'
local install_plugins = false

if vim.fn.empty(vim.fn.glob(install_path)) > 0 then
  vim.cmd('!git clone https://github.com/wbthomason/packer.nvim ' .. install_path)
  vim.cmd('packadd packer.nvim')
  install_plugins = true
end

local packer = require('packer')

vim.g.mapleader = ' '

packer.init({
  package_root = test_dir .. '/pack',
  compile_path = test_dir .. '/plugin/packer_compiled.lua'
})

packer.startup(function()
  function Use(module)
    use(require(string.format('configs.%s', module)))
  end

  -- Packer can manage itself
  packer.use 'wbthomason/packer.nvim'

  use { 'nvim-treesitter/nvim-treesitter',
    run = ':TSUpdate',
    requires = 'nvim-treesitter/nvim-treesitter-textobjects',
    config = function ()
      require('nvim-treesitter.configs').setup {
        textobjects = {
          select = {
            enable = true,
            include_surrounding_whitespace = true,
            keymaps = {
              ['af'] = '@function.outer',
              ['if'] = '@function.inner',
            },
            selection_modes = {
              ['@function.outer'] = 'V',
              ['@class.outer'] = 'V',
            },
          },
        },
      }
    end
  }


  if install_plugins then
    packer.sync()
  else
    -- load plugins at your earliest convenience
    vim.defer_fn(function()
      vim.cmd('doautocmd User LoadPlugins')
    end, 1)
  end
end)

vim.o.termguicolors = true
vim.keymap.set('n', '<c-q>', '<cmd>qa<cr>')
```

</details>

Steps to reproduce:

1. `:TSInstall python`
2. Restart Neovim (not sure if this is strictly necessary)
3. Create the following file `foo.py`:

   ```python
   class test:

       def foo():
           pass
           return "foo"

       def bar():
           pass
           return "bar"
   ```

4. `> nvim foo.py`
5. With the cursor on the first `pass` in the function `foo()` (or anywhere else in `foo()`), do `daf`

This produces the following:

```python
class test:

        pass
        return "bar"
```

But I expect this to happen:

```python
   class test:

       def bar():
           pass
           return "bar"
```

AckslD:
Thanks @melkster, this is helpful! I can confirm that I can reproduce the issue, I'll try to see what is going wrong.

PR #239, opened 2022-08-04
Create ecma for typescript inheritance to avoid errors when inheriting javascript directly.

singlexyz:
Create ecma for typescript inheritance to avoid errors when inheriting javascript directly.


theHamsta:
Thank you!

PR #243, opened 2022-08-12
feat: add support for heex

sbaildon:
It's a subset of the `html` queries; and the `element` node has been renamed to `tag`, because that's the name in the `heex` grammar


theHamsta:
Thank you!

theHamsta (approved):

Issue #217, opened 2022-06-19
Support Operating Pending Mode with Motions


barrett-ruth:
**Is your feature request related to a problem? Please describe.**
Support an operator-pending mode for moving through the textobjects (e.g. `3]f` moving 3 functions down)

**Additional context**
I know this has been brought up several times. However, it seems to be completely ignored in each issue.

Could I at least get a status update on this issue, advice for how to implement it myself, or an outright denial of adding this feature? A lot of people seem to be sitting on it.

asmodeus812:
Any news on that, is is that hard to implement into the existing code base ?

theHamsta:
Well, the we originally wanted to re-architect nvim-treestitter-textobject to use nvim's query API directly which would avoid performance issues. nvim-treesitter-textobject uses an old abstraction layer (the logic is primarily in nvim-treesitter) and a lot of features got hacked around that abstraction layer. This kind of lead to a code freeze especially in select as the refactor never happened. A lot features were just special-case integrated by non-member contributors which made the the implementation unnecessary complex.

I've tried to implement this feature about more and year ago. It was not that trivial and I found the complexity not worth it to integrate it in the current implementation. I told myself that I'd rather refactor the helper functions in nvim-treesitter to search for multiple textobjects directly.

But I'm open to contributions that'd add the feature using the current code base.

asmodeus812:
Right, gotcha. Okay, given the current code base what are the negatives of simply repeating the action for example. Looking at the implementation, there is a way to make use of v:count1 and just loop over the relevant piece of code that drives ts.goto_node. Is that a feasible solution, until the more substantial refactoring takes place ?

kiyoon:
Closing as completed. Thanks for the contribution!

PR #246, opened 2022-08-15
Allow count motion movement

asmodeus812:
Added basic capability to enumerate movement operations to allow faster jumping across textobjects. This allows the user to do something like 3]m or 3[m to move forward/backward by a multiple/count of the motion. 

https://github.com/nvim-treesitter/nvim-treesitter-textobjects/issues/217

### DIFF
--- a/lua/nvim-treesitter/textobjects/move.lua
+++ b/lua/nvim-treesitter/textobjects/move.lua
@@ -6,14 +6,15 @@ local configs = require "nvim-treesitter.configs"
 
 local M = {}
 
-local function move(query_strings, forward, start, bufnr)
+local function move(query_strings, forward, start, winid)
   query_strings = shared.make_query_strings_table(query_strings)
-  bufnr = bufnr or vim.api.nvim_get_current_buf()
+  winid = winid or vim.api.nvim_get_current_win()
+  local bufnr = vim.api.nvim_win_get_buf(winid)
 
   local config = configs.get_module "textobjects.move"
   local function filter_function(match)
     local range = { match.node:range() }
-    local row, col = unpack(vim.api.nvim_win_get_cursor(0))
+    local row, col = unpack(vim.api.nvim_win_get_cursor(winid))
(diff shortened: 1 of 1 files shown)


lua/nvim-treesitter/textobjects/move.lua
```diff
@@ -9,11 +9,13 @@ local M = {}
 local function move(query_strings, forward, start, bufnr)
   query_strings = shared.make_query_strings_table(query_strings)
   bufnr = bufnr or vim.api.nvim_get_current_buf()
-  local row, col = unpack(vim.api.nvim_win_get_cursor(0))
-  row = row - 1 -- nvim_win_get_cursor is (1,0)-indexed
 
+  local config = configs.get_module "textobjects.move"
   local function filter_function(match)
     local range = { match.node:range() }
+    local row, col = unpack(vim.api.nvim_win_get_cursor(0))
```
theHamsta, line 16: This kind of contradicts the `buf` argument (when always using 0). At the moment we didn't expose `buf` publicly. Can you change the 4th argument of this function to `win` and use `nvim_win_get_buf` to get the buffer for the window `win`. `nvim_get_current_win()` can be used if `win` is nil

theHamsta:
Thank you a lot for the contribution. I just found a tiny nitpick

lua/nvim-treesitter/textobjects/move.lua
```diff
@@ -44,24 +46,25 @@ local function move(query_strings, forward, start, bufnr)
...
-    if current_match then
-      local score = scoring_function(current_match)
-      if not best_match then
-        best_match = current_match
-        best_score = score
-      end
-      if score > best_score then
-        best_match = current_match
-        best_score = score
+  for i = 1, vim.v.count1 do
```
theHamsta, line 49: can you fix the lints? use `_` instead of `i` and format with `stylua`

theHamsta (approved):

PR #247, opened 2022-08-15
Update ruby queries after new body changes

npezza93:
With the merging of https://github.com/tree-sitter/tree-sitter-ruby/pull/224 we now have access to a body node for classes, methods, and blocks which greatly simplifies the queries for capturing the inner bits of these objects. It also is a huge performance improvement.


theHamsta:
Nice! Thank you for observing the changes in the upstream repo! Does this also require changes for the highlights?

theHamsta:
I think we need to wait for https://github.com/nvim-treesitter/nvim-treesitter/pull/3301

npezza93:
@theHamsta > I think it only makes some hidden nodes visible, right?
That's correct, nothing needs to happen in the highlights

theHamsta (approved):

PR #250, opened 2022-08-18
Unify `@call.inner` capture

### LINKED ISSUE #195: Do not include parentheses in @call.inner
With the following keymaps:

```lua
['af'] = '@call.outer'
['if'] = '@call.inner'
```

With the following example code:

```lua
my_function(argument1, argument2)
```

When pressing `dif` that would delete the arguments _and_ the parentheses like so:

```lua
my_function
```

But I personally think it would make more sense if the parentheses were not deleted:

```lua
my_function()
```

I often want to delete/replace the arguments to a function but I've never found myself wanting to delete the parentheses as well.

`@call.inner` does not include the parentheses.

I could use `i(` instead but this is a more generic command which works in many cases, but not all cases. For instance in Lua the parentheses can be omitted if there's only one argument to the function, for instance `require 'telescope'`.

echasnovski:
This PR unifies `@call.inner` captures across languages that currently support it. Resolves #195.

Most languages currently have `@call.inner` select arguments *with* parenthesis.
General rule for selecting inner function call is taken to be "all arguments, without parenthesis, of function call". As a reference, captures from [cmake](https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/4234c446d14370b3cd7604bd8e2e51ae2680f5ee/queries/cmake/textobjects.scm#L26-L30) are used, which after language-specific tweaks seem to work for most of the cases.

Notes:

- Although Cmake is already fully implemented, it is not mentioned in README.md (missing support of `@call.inner`). Didn't do anything, as it is an issue of 'scripts/update-readme.lua'.
- Having separate queries for `@call.outer` and `@call.inner` is needed to account for empty arguments (like `print()`): it doesn't have `@call.inner` capture but has `@call.outer`.
- Having `(_)?` instead of `(_)` in `"(" . (_) @_start (_)? @_end . ")"` expression is needed to work on call with single argument.
- Every explicit change was manually tested in examples containing function call with 0, 1, and more arguments. All seem to pass.
- In languages where function call is a variation of `f a b` (Fennel, Haskell), `@call.outer` is whole expression while `@call.inner` is all arguments (everything except first term). General idea for `@call.inner` query is taken from Fennel's function definition (which seems to work on my tested examples, but I might miss something here).
- With some of the languages I interacted for the first time here, so can totally miss something.

Affected languages:

- C
- C#
- C++. Done by C inheritance.
- Cuda. Done by C++ inheritance.
- Dart. There seems to be no hierarchical way of querying function call. Best I can tell, function call is an `identifier` followed by `selector` with `argument_part`. Currently it selects whole `expression_statement` as `@call.outer` which is suboptimal (it is too general and includes final `;`).
- Fennel. The `@call.inner` currently selects function *name* instead of call *arguments*. Changed to use function arguments.
- Glsl. Done by C inheritance.
- Go
- Haskell
- Java
- Javascript. Done by Ecma inheritance.
- Julia
- Lua. There are two queries to account for all function call cases: `print('aaa')`, `print'aaa'`, and `print{1, 2, 3}`.
- Python
- ~~R. Already fully implemented~~.
- Rust
- Tsx. Done by Typescript inheritance.
- Typescript. Done by Ecma inheritance.
- Zig

### DIFF
--- a/queries/lua/textobjects.scm
+++ b/queries/lua/textobjects.scm
@@ -4,7 +4,11 @@
 
 ; call
 
-(function_call) @call.outer (arguments) @call.inner
+(function_call) @call.outer ((arguments) @call.inner (#match? @call.inner "^[^\\(]"))
+
+(function_call
+  arguments: (arguments . "(" . (_) @_start (_)? @_end . ")"
+  (#make-range! "call.inner" @_start @_end))) @call.outer
 
 ; class
 
--- a/queries/lua/textobjects.scm
+++ b/queries/lua/textobjects.scm
@@ -4,7 +4,9 @@
 
 ; call
 
-(function_call) @call.outer ((arguments) @call.inner (#match? @call.inner "^[^\\(]"))
+(function_call
+  (arguments) @call.inner
+  (#match? @call.inner "^[^\\(]")) @call.outer
 
 (function_call
   arguments: (arguments . "(" . (_) @_start (_)? @_end . ")"
--- a/queries/c/textobjects.scm
+++ b/queries/c/textobjects.scm
@@ -41,7 +41,9 @@
 (comment) @comment.outer
 
 (call_expression) @call.outer
-(call_expression (_) @call.inner)
+(call_expression
+  arguments: (argument_list . "(" . (_) @_start (_)? @_end . ")"
+  (#make-range! "call.inner" @_start @_end)))
 
 ; Statements
 
--- a/queries/c_sharp/textobjects.scm
+++ b/queries/c_sharp/textobjects.scm
@@ -32,8 +32,10 @@
   body: (switch_body) @conditional.inner) @conditional.outer
 
 ;; calls
-(invocation_expression 
-  (argument_list) @call.inner) @call.outer
+(invocation_expression) @call.outer
+(invocation_expression
+  arguments: (argument_list . "(" . (_) @_start (_)? @_end . ")"
+  (#make-range! "call.inner" @_start @_end)))
 
 ;; blocks
 (_ (block) @block.inner) @block.outer
--- a/queries/dart/textobjects.scm
+++ b/queries/dart/textobjects.scm
@@ -55,8 +55,18 @@
  (#make-range! "parameter.outer" @_start @parameter.inner))
 
 ; call
-(expression_statement
-  (selector) @call.inner) @call.outer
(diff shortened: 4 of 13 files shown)


kylechui:
FWIW, I've tried it with Python/Lua and it seems to work without a hitch!

queries/lua/textobjects.scm
```diff
@@ -4,7 +4,11 @@
 
 ; call
 
-(function_call) @call.outer (arguments) @call.inner
+(function_call) @call.outer ((arguments) @call.inner (#match? @call.inner "^[^\\(]"))
```
kylechui, line 7: @echasnovski Hi there, I was reading through the source documentation on Tree-sitter queries so that I could hopefully also contribute here/fix some queries as the need arises. Most of the other queries in this PR make a decent amount of sense; from what I gather, `#make-range!` just takes a range of nodes from `@_start` to `@_end` (inclusive) and assigns that to the `@call.inner` capture here. However, what is this line 7 supposed to do? It seems like it would just be a second definition for `@call.inner` for Lua files?

Edit: Didn't mean to start a review, just wanted to leave a short question oops ;-;

echasnovski: Basically it can be read as written: capture `function_call` as `@call.outer`; capture `arguments` as `@call.inner` only if it matches `"^[^\\(]"` (i.e. doesn't start with `(`). This query accounts for (in my opinion) the most confusing part of Lua - ability to omit parenthesis in function call in case of a single string or table argument. This is needed because "main" query defines `@call.inner` as something strictly inside parenthesis of a function call. 

But thanks for the question, as it prompted me to make this query a little bit cleaner by narrowing `@call.outer` to not overlap with second query.

echasnovski:
@theHamsta, sorry to bother, but is there any way to get a review on this? There are two surround plugins waiting for this to be merged :)

theHamsta (approved):
Thank you! This looks good! It's a bit sad for me that the actual function got removed for `@call.inner` but probably this should be its own capture.

echasnovski:
Thanks for merging this!

 > It's a bit sad for me that the actual function got removed for @call.inner but probably this should be its own capture.

If this is about Fennel, then yeah, it does seem like function name should be its own capture. Something like `@call.name`, probably.

PR #255, opened 2022-08-24
feat(jsx): created the missing jsx queries

### LINKED ISSUE #222: referenced JSX does not exist
**Language informations**

Hello, I noticed there is some references to the jsx language within the text objects files but I was not able to find that fule anywhere. Is it a mistake?

I have a couple of queries already for JSX, will you accept a PR?
Regards

danielo515:
fixes #222

### DIFF
--- a/queries/jsx/textobjects.scm
+++ b/queries/jsx/textobjects.scm
@@ -1,0 +1,5 @@
+(
+ (jsx_attribute (property_identifier)@arg.name
+                (_)@parameter.inner
+  )@parameter.outer
+)
--- a/queries/jsx/textobjects.scm
+++ b/queries/jsx/textobjects.scm
@@ -1,5 +1,3 @@
-(
- (jsx_attribute (property_identifier)@arg.name
-                (_)@parameter.inner
-  )@parameter.outer
-)
+; inherits: ecma
+((jsx_attribute
+   (property_identifier) (_) @parameter.inner) @parameter.outer)
--- a/queries/javascript/textobjects.scm
+++ b/queries/javascript/textobjects.scm
@@ -1,1 +1,1 @@
-; inherits: ecma
+; inherits: ecma,jsx


theHamsta:
but now you need to add `jsx` here again: https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/master/queries/javascript/textobjects.scm for JS, typescript and tsx

danielo515:
> but now you need to add `jsx` here again: https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/master/queries/javascript/textobjects.scm for JS, typescript and tsx

Can you explain it a but further? As far as I know, tsx already references JSX

theHamsta:
>>    but now you need to add jsx here again: https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/master/queries/javascript/textobjects.scm for JS, typescript and tsx

> Can you explain it a but further? As far as I know, tsx already references JSX

But JS not anymore https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/master/queries/javascript/textobjects.scm

danielo515:
@theHamsta do you mean that typescript should also extend JSX right. Typescript does not allow JSX in plain TS files, so it is not needed there

theHamsta:
> @theHamsta do you mean that typescript should also extend JSX right. Typescript does not allow JSX in plain TS files, so it is not needed there

that might be true, but javascript includes jsx

danielo515:
> > @theHamsta do you mean that typescript should also extend JSX right. Typescript does not allow JSX in plain TS files, so it is not needed there
> 
> that might be true, but javascript includes jsx

Yes, yes, and I will be adding it. I mistakenly wrote typescript twice, when the first one should have been Javascript, meaning that I will be adding it to Javascript but not to typescript. Sorry for the confusion

danielo515:
@danielo515 I added the suggestions. However, I am not sure about how the inherit mechanism works, so I don't know if what I did is correct

queries/jsx/textobjects.scm
```diff
@@ -0,0 +1,6 @@
+; inherits: ecma
```
theHamsta, line 1: i don't think this include is necessary. jsx is just a optional mixin for js and tsx

queries/jsx/textobjects.scm
```diff
@@ -0,0 +1,6 @@
...
+(
+ (jsx_attribute (property_identifier)@arg.name
```
theHamsta, line 3: `@arg.name` is not in CONTRIBUTING.md

queries/jsx/textobjects.scm
danielo515: The reason I added this was because I am not sure if a JSX file will be resolved to both jsx and JavaScript or just, jsx. If it is just jsx you probably want your JavaScript queries too right? I included ecma because JavaScript would result in circular dependency

theHamsta:
Thank you!

theHamsta (approved):

Issue #278, opened 2022-09-02
'move' mappings are not created


loqusion:
**Describe the bug**

'move' mappings are not created

**To Reproduce**

Here's my treesitter config. The mappings I made for 'move' don't work, and when I enter `:map ]m`, etc to check if the mappings are made, it says "No mapping found". However, the 'select' mappings are still made.

```lua
  require('nvim-treesitter.configs').setup {
    ensure_installed = 'all',
    ignore_install = { 'php', 'phpdoc' },
    additional_vim_regex_highlighting = false,
    highlight = {
      enable = true,
      disable = {
        'vim',
        'help',
      },
    },
    indent = {
      -- Enabling this removes the leading space from /*...*...*/ comments! This is annoying!!
      enable = false,
    },
    incremental_selection = {
      enable = false,
      keymaps = {
        init_selection = 'gnn',
        scope_incremental = 'gna',
        node_incremental = 'grn',
        node_decremental = 'grm',
      },
    },
    autotag = {
      enable = true,
      filetypes = { 'html', 'javascript', 'javascriptreact', 'typescriptreact', 'svelte', 'vue', 'markdown' },
    },
    context_commentstring = {
      -- https://github.com/JoosepAlviste/nvim-ts-context-commentstring#commentnvim
      enable = true,
      enable_autocmd = false,
    },
    endwise = { enable = true },
    matchup = { enable = true },
    query_linter = {
      enable = true,
      use_virtual_text = true,
      lint_events = { 'BufWrite', 'CursorHold' },
    },
    tree_setter = {
      enable = false,
    },
    textobjects = {
      select = {
        enable = true,
        -- Automatically jump forward to textobj, similar to targets.vim
        lookahead = true,
        keymaps = {
          ['af'] = '@function.outer',
          ['if'] = '@function.inner',
          ['il'] = '@loop.inner',
          ['al'] = '@loop.outer',
          -- ['icd'] = '@conditional.inner',
          -- ['acd'] = '@conditional.outer',
          -- ['acm'] = '@comment.outer',
          ['ast'] = '@statement.outer',
          ['isc'] = '@scopename.inner',
          ['iB'] = '@block.inner',
          ['aB'] = '@block.outer',
          ['ia'] = '@parameter.inner',
          ['aa'] = '@parameter.outer',
        },
      },
      move = {
        enable = true,
        set_jumps = true, -- whether to set jumps in the jumplist
        goto_next_start = {
[...]

PR #286, opened 2022-09-03
fix: nil query in closure when using mapping

mpaliwoda:
Should fix https://github.com/nvim-treesitter/nvim-treesitter-textobjects/issues/278 edit: actually not sure whether it fixes the issue as a whole or rather just [@NEX-S problem](https://github.com/nvim-treesitter/nvim-treesitter-textobjects/issues/278#issuecomment-1236142161)

Not sure whether bad argument error was because `config_queries` was set to `nil` at the end of a loop or it simply went out of scope but introducing another local variable made it work

Took the liberty of slight refactoring, hope that's ok

### DIFF
--- a/lua/nvim-treesitter/textobjects/attach.lua
+++ b/lua/nvim-treesitter/textobjects/attach.lua
@@ -5,30 +5,29 @@ local M = {}
 
 function M.make_attach(normal_mode_functions, submodule)
   return function(bufnr, lang)
-    local config = configs.get_module("textobjects." .. submodule)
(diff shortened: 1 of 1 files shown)


theHamsta:
Fancy, I essentially made the same commit https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/289 :smile:

theHamsta (approved):
Thank you!

theHamsta (approved):

PR #291, opened 2022-09-07
feat(lua): add table_constructor fields as parameters

AckslD:


AckslD:
@LamprosPitsillos you mean specifically to use `@field.inner` instead of `@parameter.inner` for the capture name? I think it makes more sense to use `parameter.inner` for all since this is how it's done in other languages as well. For example in python we don't distinguish function arguments from entries of a dict.

@theHamsta any thougths?

AckslD:
See also discussion in #86

PR #293, opened 2022-09-12
feat: python support for top-level @statement.outer

bnprks:
This adds support for top-level statements in Python `@statement.outer`, which enables statement selection in a scripting context where there is no enclosing block.


theHamsta:
Thank you!

PR #295, opened 2022-09-16
Update README.md

briandipalma:


kuator:
When october is near

theHamsta (approved):

theHamsta:
Though I think the `xor` was intentional it's probably better to stay with English language

PR #298, opened 2022-09-20
feat: improve PHP queries

laytan:
- excludes the braces for all the inside queries
- add query for if statements
- add query for plain for loops

### DIFF
--- a/queries/php/textobjects.scm
+++ b/queries/php/textobjects.scm
@@ -1,54 +1,52 @@
 ;; functions
 (function_definition
   body: (compound_statement . "{" . (_) @_start @_end (_)? @_end . "}"
-  (#make-range! "function.inner" @_start @_end))) @function.outer
+  (#make-range! "function.inner" @_start @_end)))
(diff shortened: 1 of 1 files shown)


laytan:
I have one addition to make which I forgot

gbprod:
We can also introduce textobjects for `call` (I use it for some days) : 

```query
; call

(function_call_expression) @call.outer

(member_call_expression) @call.outer

(function_call_expression
    arguments: (arguments . "(" . (_) @_start (_)? @_end . ")"
    (#make-range! "call.inner" @_start @_end)))

(member_call_expression
    arguments: (arguments . "(" . (_) @_start (_)? @_end . ")"
    (#make-range! "call.inner" @_start @_end)))
```

theHamsta (approved):

theHamsta:
@gbprod i think you can create a follow up PR.

Thank you @laytan !

unansweredocd:
Thank you so much for adding this!

PR #304, opened 2022-10-06
Add php statement queries

gbprod:
Not sure that is complete. 
Does conditionnals and loop should be considered as statement here ?

### DIFF
--- a/queries/php/textobjects.scm
+++ b/queries/php/textobjects.scm
@@ -96,3 +96,32 @@
   arguments: (arguments . "(" . (_) @_start (_)? @_end . ")"
   (#make-range! "call.inner" @_start @_end)))
 
+;; statement
+[
+  (expression_statement)
+  (declare_statement)
+  (return_statement)
+  (namespace_use_declaration)
+  (namespace_definition)
+  (if_statement)
+  (empty_statement)
+  (switch_statement)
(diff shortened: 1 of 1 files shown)


gbprod:
Regarding to `c` textobjects, it seems that conditionnals and loop should be considered as statement.
I've completed this pull request.

gbprod:
List comes from grammar : https://github.com/tree-sitter/tree-sitter-php/blob/master/grammar.js#L123

unansweredocd:
Do you happen to know whether there is a textobject that selects a callback function?

gbprod:
> Do you happen to know whether there is a textobject that selects a callback function?

Do you mean an anonymous function ? (like `$f = function() { return; };`)
Currently, the "function" textobject doesn't catch anon functions, it should IMHO.

unansweredocd:
Yes I mean an anonymous function. I figured it would have its own textobject but what you're saying makes sense, function "should" catch anon functions.

> > Do you happen to know whether there is a textobject that selects a callback function?
> 
> Do you mean an anonymous function ? (like `$f = function() { return; };`) Currently, the "function" textobject doesn't catch anon functions, it should IMHO.

gbprod:
Make sense, I'll try to make something in another pull request

PR #309, opened 2022-10-12
Clarify the need for extends modeline

phelipetls:
...when overriding/extending textobjects.

I just found out that, in v0.8.0, you're supposed to start the file with `; extends`, otherwise the file will be ignored (only one query file will be considered).

See https://github.com/neovim/neovim/pull/20105/ and https://github.com/neovim/neovim/issues/20172


theHamsta (approved):
Thank you!

PR #310, opened 2022-10-12
Respect offset directive when getting text object range

phelipetls:
I found it hard to use the "#make-range!" pattern to create a text object to select inside JavaScript template strings

```js
const templateString = `
  Hello world
`
```

I tried to use something like this but it didn't work:

```scheme
(template_string . "`" _ @_start _? @_end "`" . (#make-range! "mytextobject" @_start @_end))
```

I'd expect this to work though, but it didn't:

```scheme
((template_string) @mytextobject (#offset! @mytextobject 0 1 0 -1))
```

So this patch will make it work now, the offset directive will be respected.

### DIFF
--- a/lua/nvim-treesitter/textobjects/shared.lua
+++ b/lua/nvim-treesitter/textobjects/shared.lua
@@ -120,18 +120,26 @@ function M.textobject_at_point(query_string, pos, bufnr, opts)
     end
   end
 
+  local get_range = function(match)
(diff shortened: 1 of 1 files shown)


theHamsta:
Nice! Thank you!

theHamsta:
At some point nvim-treesitter-textobjects really needs to migrate to the new query system and stop using `lua/nvim-treesitter/query.lua`

Issue #210, opened 2022-05-15
Automatically set filetype in peek_definition_code


igorlfs:
**Is your feature request related to a problem? Please describe.**
When using `peek_definition_code`, the filetype isn't set in the hover window. Therefore, to enable treesitter's highlighting, you have to manually set it.

**Describe the solution you'd like**
Ideally, there's a way for the LSP to query the filetypes, allowing treesitter to automatically set them.

**Describe alternatives you've considered**
Manually setting the filetype and I think it's also possible to write an autocommand?

PR #312, opened 2022-10-16
feat: set filetype in peek_definition_code

wodesuck:
To make ftplugin work in preview buffer (e.g. set tabstop for special filetype)
Also enable treesitter highlighting #210


theHamsta:
Can you rebase to make test pass?

PR #317, opened 2022-11-07
fix(rust): remove `if_let_expression` and `while_let_expression`

### LINKED ISSUE #316: Query: invalid node type at position 1400 for language rust
**Describe the bug**

After installing nvim-treesitter-textobject, I get the following error on every rust file:

```
Error detected while processing BufReadPost Autocommands for "*":
Error executing lua callback: /usr/share/nvim/runtime/filetype.lua:22: Error executing lua: /usr/share/nvim/runtime/
filetype.lua:23: Vim(append):Error executing lua callback: /usr/share/nvim/runtime/lua/vim/treesitter/query.lua:219:
 query: invalid node type at position 1400 for language rust
stack traceback:
        [C]: in function '_ts_parse_query'
        /usr/share/nvim/runtime/lua/vim/treesitter/query.lua:219: in function 'get_query'
        ...cker/start/nvim-treesitter/lua/nvim-treesitter/query.lua:104: in function 'get_query'
        ...r-textobjects/lua/nvim-treesitter/textobjects/attach.lua:9: in function 'attach'
        ...er/start/nvim-treesitter/lua/nvim-treesitter/configs.lua:505: in function 'attach_module'
        ...er/start/nvim-treesitter/lua/nvim-treesitter/configs.lua:528: in function 'reattach_module'
        ...er/start/nvim-treesitter/lua/nvim-treesitter/configs.lua:131: in function <...er/start/nvim-treesitter/lu
a/nvim-treesitter/configs.lua:130>
```

**To Reproduce**

1. Install the plugin
2. Open any rust file

**Expected behavior**

No errors

**Output of `:checkhealth nvim-treesitter`**

<details>
<code>

========================================================================
## Installation
  - OK: `tree-sitter` found 0.20.7 (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v19.0.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: cc (GCC) 12.2.0
  - OK: Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

## Parser/Features H L F I J
  - cpp            ✓ ✓ ✓ ✓ ✓
  - dockerfile     ✓ . . . ✓
  - html           ✓ ✓ ✓ ✓ ✓
  - yaml           ✓ ✓ ✓ ✓ ✓
  - c              ✓ ✓ ✓ ✓ ✓
  - make           ✓ . ✓ . ✓
  - r              ✓ ✓ . ✓ ✓
  - scala          ✓ . ✓ . ✓
  - lua            ✓ ✓ ✓ ✓ ✓
  - bash           ✓ ✓ ✓ . ✓
  - toml           ✓ ✓ ✓ ✓ ✓
  - rust           ✓ ✓ ✓ ✓ ✓

         x) errors found in the query, try to run :TSUpdate {lang}

</code>
</details>

**Output of `nvim --version`**
```
NVIM v0.8.0
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by builduser

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info
```

**Additional context**

It used to work just fine, until today when I did PackerSync.

numToStr:
I briefly checked and it seems `{if,while}_let_expression` can be replace 1:1 by existing `{if,while}_expression` queries.

This PR also removes `meta_arguments` related queries because it was giving the same error.

Fixes #316

### DIFF
--- a/queries/rust/textobjects.scm
+++ b/queries/rust/textobjects.scm
@@ -47,24 +47,11 @@
 
 (match_expression) @conditional.outer
 
-(if_let_expression
-  consequence: (block)?
-  @conditional.inner) @conditional.outer
-
-(if_let_expression
-  alternative: (else_clause (block) @conditional.inner))
-
-(if_let_expression
-  pattern: (_) @conditional.inner)
-
 ;; loops
 (loop_expression
   (_)? @loop.inner) @loop.outer
 
 (while_expression
-  (_)? @loop.inner) @loop.outer
-
-(while_let_expression
   (_)? @loop.inner) @loop.outer
 
 (for_expression
@@ -137,11 +124,3 @@
 ((type_arguments
   . (_) @parameter.inner . ","? @_end)
  (#make-range! "parameter.outer" @parameter.inner @_end))
-
-((meta_arguments
-  "," @_start . (_) @parameter.inner)
- (#make-range! "parameter.outer" @_start @parameter.inner))
-((meta_arguments
-  . (_) @parameter.inner . ","? @_end)
- (#make-range! "parameter.outer" @parameter.inner @_end))
-


queries/rust/textobjects.scm
```diff
@@ -47,26 +47,13 @@
...
 
-(if_let_expression
-  consequence: (block)?
-  @conditional.inner) @conditional.outer
-
-(if_let_expression
-  alternative: (else_clause (block) @conditional.inner))
-
-(if_let_expression
-  pattern: (_) @conditional.inner)
```
nagua, line 58: I think, that this node isn't handled in the normal `if_expression` path.

queries/rust/textobjects.scm
```diff
@@ -47,26 +47,13 @@
 
 (match_expression) @conditional.outer
 
-(if_let_expression
-  consequence: (block)?
```
nagua, line 51: I'm not sure how much it matters, but in the `if_expression` equivalent the `(block)` isn't optional as it is in this case.

nagua:
Hello @numToStr I also just encountered this problem today and I was also looking at the queries. I'm not very well versed in tree-sitter queries and tree-sitter, but it seems to me that we would loose functionality with this PR.

numToStr:
> but it seems to me that we would loose functionality with this PR.

It is already crashing the editor. IMO it is worse than loosing functionality.

That being said, I am also not sure what those `if_let_expression` were exactly capturing. And with my limited `if let` code all seems fine. But yes, I am open to suggestions.

queries/rust/textobjects.scm
nagua: The equivalent to this should be:

```
(if_expression
  condition: (let_condition pattern: (_) @conditional.inner))
```

But I'm not sure if this would clash with this one:

```
(if_expression
  condition: (_) @conditional.inner)
```

If I understand this correctly we would now assign `conditional.inner` to two different parts of the AST. So this might not work. But again I'm not that deep into tree-sitter semantics.

numToStr: With this PR and considering the following code.

```rust
if let Some(i) = emoticon {
    println!("Matched {:?}!", i);
}
```

`@conditional.inner` will capture `let Some(i) = emoticon` and `@conditional.outer` will select the whole if expression, which makes sense to me. What do you think?

queries/rust/textobjects.scm
numToStr: I don't think rust even allows optional block in `if` expression.

numToStr:
I am not sure why the CI is failing, rust queries are fine though.

miaouPlop:
>  I am not sure why the CI is failing, rust queries are fine though.

The output of the CI:

> ...YZc5/usr/share/nvim/runtime/lua/vim/treesitter/query.lua:219: query: invalid node type at position 11 for language julia
Error: Process completed with exit code 1.

So Rust is probably not the only language impacted by the changes made in treesitter.

numToStr:
> So Rust is probably not the only language impacted by the changes made in treesitter.

Yes, might be but that is kinda unrelated to this PR :)

anstadnik:
Hi, hate to bump things, but currently this plugin breaks the editor. Can this PR be merged somehow?
AFAIK, the failing tests are not due to the rust stuff.

ClSlaid:
> So Rust is probably not the only language impacted by the changes made in treesitter.

Maybe that's because `Julia` is also broken.

> E5108: Error executing lua /usr/share/nvim/runtime/lua/vim/treesitter/query.lua:219: query: invalid node type at position 607 for language julia

queries/rust/textobjects.scm
p00f: > But I'm not sure if this would clash with this one:

It doesn't - if the cursor is inside `Some(foo)` in `if let Some(foo) = None {}` it matches only the pattern, if your cursor is outside it but inside `let Some(foo) = None` it matches the whole condition

PR #318, opened 2022-11-08
fix(c_sharp): modified queries to avoid inner objects to contain braces

icyd:
Don't use C#, but these changes seem to be enough


Goxore:
Works as expected

lxalln:
Would this be possible to extend to the loops/conditional blocks too?

PR #327, opened 2022-11-26
feat: support regex patterns

### LINKED ISSUE #193: Regex Patterns
Users want to manipulate the inner and outer regexp patterns

Assuming `r` is bound to the new regexp pattern text object...

```js
// some module.js
const RE = /-(.*)-/mg;
//            ^ cursor here
```
`yir` - yanks `-(.*)-`
`yar` - yanks `/-(.*)-/mg`

```html
<!-- some document.html -->
<input pattern="[0-9]+-[a-zA-Z]+"/>
<!--             ^ cursor here -->
```  
`yir` - yanks `[0-9]+-[a-zA-Z]+`
`yar` - yanks `"[0-9]+-[a-zA-Z]+"`

(HTML example depends on https://github.com/nvim-treesitter/nvim-treesitter/pull/2577)

one would have to construct their own selection painstakingly, as `/` could be a legal regexp char

ObserverOfTime:
These are some textobjects I was already using locally.
I'll add support for a few more languages soon.

Closes #193

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,6 +21,8 @@ Each capture group can be declared as `inner` or `outer`.
 @block.outer
 @parameter.inner
 @parameter.outer
+@regex.inner
+@regex.outer
 
 # For LaTeX frames
 @frame.inner
--- a/queries/bash/textobjects.scm
+++ b/queries/bash/textobjects.scm
@@ -15,3 +15,5 @@
   (_) @loop.inner ) @loop.outer
 
 (comment) @comment.outer
+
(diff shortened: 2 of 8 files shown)


ObserverOfTime:
In Perl, `regex.inner` will also select the `/` delimiters in these cases:

```perl
m/bar/i  # patter_matcher_m
qr/foo/i  # regex_pattern_qr
```

Using just `offset` won't work since the suffix can be of arbitrary length
and combining `make-range` with `offset` is apparently not possible.

theHamsta (approved):

PR #330, opened 2022-12-08
Fix GitHub Actions check query files

kiyoon:
The latest Ubuntu doesn't come with libfuse2, so the appimage wouldn't run without this dependency as shown below.  
This pull request fixes this issue and re-enables checking query files Action.

Problem:  
image

Fix:  
image


theHamsta (approved):

theHamsta:
Thank you! Just wanted to do the same!

PR #342, opened 2022-12-22
Use <cmd> in keymapping to prevent flicker on some UIs

phgz:
Using `:` in the mapping causes some UIs to glitch when repeating a command (`.`) on a textobject. `<cmd>` prevents that.

flicker


kiyoon:
Thank you!

PR #343, opened 2022-12-22
feat: make swap functions repeatable

theHamsta:
Revisiting #52 #55 . Would love to have a solution that works with Lua functions.

Problem: `vim.o.opfunc` doesn't seem accept Lua functions or I'm doing something wrong here

### DIFF
--- a/lua/nvim-treesitter/textobjects/attach.lua
+++ b/lua/nvim-treesitter/textobjects/attach.lua
@@ -3,8 +3,17 @@ local parsers = require "nvim-treesitter.parsers"
 local queries = require "nvim-treesitter.query"
 local M = {}
 
-function M.make_attach(normal_mode_functions, submodule, keymap_modes)
+local function make_repeatable(fn)
+  return function()
+    _G._nvim_treesitter_textobject_last_function = fn
+    vim.o.opfunc = "v:lua._nvim_treesitter_textobject_last_function"
+    vim.api.nvim_feedkeys("g@l", "n", false)
+  end
+end
+
+function M.make_attach(normal_mode_functions, submodule, keymap_modes, opts)
   keymap_modes = keymap_modes or "n"
+  opts = opts or {}
   return function(bufnr, lang)
     lang = lang or parsers.get_buf_lang(bufnr)
     if not queries.get_query(lang, "textobjects") then
@@ -26,9 +35,18 @@ function M.make_attach(normal_mode_functions, submodule, keymap_modes)
           mapping_description = function_description .. " " .. query_metadata
         end
 
-        vim.keymap.set(keymap_modes, mapping, function()
+        local fn = function()
           require("nvim-treesitter.textobjects." .. submodule)[function_call](query)
-        end, { buffer = bufnr, silent = true, remap = false, desc = mapping_description })
+        end
+        if opts.repeatable then
+          fn = make_repeatable(fn)
+        end
+        vim.keymap.set(
+          keymap_modes,
+          mapping,
+          fn,
+          { buffer = bufnr, silent = true, remap = false, desc = mapping_description }
+        )
       end
     end
   end
--- a/lua/nvim-treesitter/textobjects/swap.lua
+++ b/lua/nvim-treesitter/textobjects/swap.lua
@@ -38,7 +38,7 @@ end
 
 local normal_mode_functions = { "swap_next", "swap_previous" }
 
-M.attach = attach.make_attach(normal_mode_functions, "swap", "n")
+M.attach = attach.make_attach(normal_mode_functions, "swap", "n", { repeatable = true })
 M.detach = attach.make_detach(normal_mode_functions, "swap", "n")
 
 M.commands = {


lua/nvim-treesitter/textobjects/attach.lua
```diff
@@ -26,9 +27,23 @@ function M.make_attach(normal_mode_functions, submodule, keymap_modes)
           mapping_description = function_description .. " " .. query_metadata
         end
 
-        vim.keymap.set(keymap_modes, mapping, function()
-          require("nvim-treesitter.textobjects." .. submodule)[function_call](query)
-        end, { buffer = bufnr, silent = true, remap = false, desc = mapping_description })
+        local function fn()
+          if opts.repeatable then
+            _G._nvim_treesitter_textobject_last_function = function()
```
theHamsta, line 32: not nice :cry:

lua/nvim-treesitter/textobjects/attach.lua
```diff
@@ -26,9 +27,23 @@ function M.make_attach(normal_mode_functions, submodule, keymap_modes)
...
+              require("nvim-treesitter.textobjects." .. submodule)[function_call](query)
+            end
+            vim.o.opfunc = "v:lua._nvim_treesitter_textobject_last_function"
+            vim.api.nvim_feedkeys("g@l", "n", false)
```
theHamsta, line 36: going via `opfunc` is also a bit of a hack

theHamsta: need to check for which modes this can work

phgz:
> Problem: `vim.o.opfunc` doesn't seem accept Lua functions or I'm doing something wrong here

I think it effectively can't at the moment. Have you check **gitsigns** implementation?
https://github.com/lewis6991/gitsigns.nvim/blob/main/lua/gitsigns/repeat.lua

I've tried it locally on swap and it seems to work well. I particularly like the non `_G` usage of it. Other than that your implementation seems nice.

theHamsta:
@phgz you're right that it makes sense to create a `make_repeatable` function and not have it in the swap module. Then, we can also easily extend it to other modules and improve it once there's a better way to do it via Neovim API.

theHamsta:
I wonder whether we should have `make_repeatable` in nvim-treesitter https://github.com/nvim-treesitter/nvim-treesitter/issues/3261 (though I find it unintuitive for repeat to work in visual mode). Maybe it's not to bad to have a separate implementation in each of the repos or remove here once nvim-treesitter has this function.

theHamsta:
@phgz Ah, I remember why we don't do it the gitsigns.nvim way. We can't do because we have a `-` in our repo name. :smile: Then it won't work with `v:lua`

kiyoon:
I just tested this, and it seems to suffer from #344 again.

theHamsta:
> I just tested this, and it seems to suffer from #344 again.

How can this affect visual mode? It only modifies a normal mode command.

kiyoon:
For some reason, #342 is included in your pull request. Can you merge the branch with the master?

kiyoon:
The functionality works well for me.

lua/nvim-treesitter/textobjects/attach.lua
kiyoon: If I understand correctly, the `opfunc` doesn't accept lua function, so you hacked the function to be a vimscript-like function and used it?
And you're not sure if the repeatable function works with select move etc?

theHamsta:
@kiyoon you were right. This PR was missing a rebase, so it didn't include the revert. @kiyoon you think this is ready?

kiyoon (approved):
The functionality is implemented with no known side effects

lua/nvim-treesitter/textobjects/attach.lua
theHamsta: yes, I made it a vim function through a global. https://github.com/lewis6991/gitsigns.nvim/blob/main/lua/gitsigns/repeat.lua does not work because we would need to write `require'nvim-treesitter.textobject'...` and Vim does not parse the opfunc when there is a `-` in the string.

I don't know for which other functions `.` could be applicable. Typically `.` is to repeat edits. Swap is the only module that performs edits. There was a desire to introduce `;` for repeating the motions but this would be a different functionality and PR.

lua/nvim-treesitter/textobjects/attach.lua
theHamsta: but https://github.com/lewis6991/gitsigns.nvim/blob/main/lua/gitsigns/repeat.lua does not work due to `-` in the name.

lua/nvim-treesitter/textobjects/attach.lua
phgz: > There was a desire to introduce ; for repeating the motions but this would be a different functionality and PR.

That would be nice! At the moment I kind of hack it through "dahu/vim-fanfingtastic" and  "Houl/repmo-vim".

JosefLitos:
> We can't do because we have a `-` in our repo name. 😄 Then it won't work with `v:lua`.

@theHamsta are you sure it doesn't work? I just made a PR for `syntax-tree-surfer` using this very approach and it worked fine. The only limitation is formatting-wise - you cannot use `()` around required module, but this works just fine: `vim.go.opfunc = [[v:lua.require'dash-name'.dot_repeat]]`

PR #346, opened 2022-12-22
Revert "Use <cmd> in keymapping to prevent flicker on some UIs"

### LINKED ISSUE #344: Visual mode commands are broken
Since #342 was merged, if you use e.g. `vaf` where `af` is [mapped](https://github.com/michelesr/nvim-config/blob/20ed9bff1474aafc0ac71cc90c4f4a6096a6ce04/lua/config/tresitter.lua#L41-L42) as:

```
x  af          *@<Cmd>lua require'nvim-treesitter.textobjects.select'.select_textobject('@function.outer', 'x')<CR>
                 Select textobject @function.outer
```
The text is not selected correctly, instead, a new line gets added to the text. See attached screenshots.

Cursor position

screenshot-2022-12-22-14:17:56

Pressing `vaf` after #342 

screenshot-2022-12-22-14:18:02

Pressing `vaf` before #342
screenshot-2022-12-22-14:18:28

<details>

Output of checkhealth:

```
========================================================================
## Installation
  - OK: `tree-sitter` found 0.20.7 (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v19.2.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: cc (GCC) 12.2.0
  - OK: Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

## OS Info:
{
  machine = "x86_64",
  release = "6.0.12-arch1-1",
  sysname = "Linux",
  version = "#1 SMP PREEMPT_DYNAMIC Thu, 08 Dec 2022 11:03:38 +0000"
}

## Parser/Features         H L F I J
  - python              ✓ ✓ ✓ ✓ ✓
  - json                ✓ ✓ ✓ ✓ .
  - bash                ✓ ✓ ✓ . ✓
  - lua                 ✓ ✓ ✓ ✓ ✓
  - html                ✓ ✓ ✓ ✓ ✓
  - c                   ✓ ✓ ✓ ✓ ✓
  - go                  ✓ ✓ ✓ ✓ ✓
  - markdown            ✓ . ✓ . ✓
  - ruby                ✓ ✓ ✓ ✓ ✓
  - markdown_inline     ✓ . . . ✓
  - rust                ✓ ✓ ✓ ✓ ✓
  - query               ✓ ✓ ✓ ✓ ✓
  - yaml                ✓ ✓ ✓ ✓ ✓

         x) errors found in the query, try to run :TSUpdate {lang}
```

</details>

NVIM version:

```
NVIM v0.8.1
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by builduser

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info
```

theHamsta:
This reverts commit 4c9bcaaabcc9b978871b97ee6142417f77aa0b53.

Fixes #344

### DIFF
--- a/lua/nvim-treesitter/textobjects/select.lua
+++ b/lua/nvim-treesitter/textobjects/select.lua
@@ -176,9 +176,9 @@ function M.attach(bufnr, lang)
       --vim.keymap.set({ "o", "x" }, mapping, function()
       --require("nvim-treesitter.textobjects.select").select_textobject(query)
       --end, { buffer = buf, silent = true, remap = false, desc = desc })
-      local cmd_o = "<cmd>lua require'nvim-treesitter.textobjects.select'.select_textobject('" .. query .. "', 'o')<CR>"
+      local cmd_o = ":lua require'nvim-treesitter.textobjects.select'.select_textobject('" .. query .. "', 'o')<CR>"
       api.nvim_buf_set_keymap(buf, "o", mapping, cmd_o, { silent = true, noremap = true, desc = desc })
-      local cmd_x = "<cmd>lua require'nvim-treesitter.textobjects.select'.select_textobject('" .. query .. "', 'x')<CR>"
+      local cmd_x = ":lua require'nvim-treesitter.textobjects.select'.select_textobject('" .. query .. "', 'x')<CR>"
       api.nvim_buf_set_keymap(buf, "x", mapping, cmd_x, { silent = true, noremap = true, desc = desc })
     end
   end


theHamsta:
@kiyoon  This is why I wanted to add tests. At the moment, it is almost impossible to merge even small changes without manual testing as it might break one of the features that somebody added before.

I don't know why `<cmd>` doesn't work. Maybe it is a similar issue as why `vim.keymap.set` does not work or maybe it is expected that `<cmd> leaves visual mode.

theHamsta:
@michelesr this does fix the issue for me. The proper fix to address https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/342 would probably be to investigate why `vim.keymap.set` does not work

theHamsta:
@phgz is it expected that GUIs flicker on `:` even when the mapping is silent?

theHamsta:
> @theHamsta About writing tests, do you mind if I write tests with python using pynvim?

I'm fine with it when it works well for you. In nvim-treesitter, we use plenary. I launches nvim for all `describe` invocations in `**/*_spec.lua`. https://github.com/nvim-treesitter/nvim-treesitter/blob/master/tests/query/injection_spec.lua

kiyoon:
> I'm fine with it when it works well for you. In nvim-treesitter, we use plenary. I launches nvim for all `describe` invocations in `**/*_spec.lua`. https://github.com/nvim-treesitter/nvim-treesitter/blob/master/tests/query/injection_spec.lua

Thanks for the pointer. I am just more used to Python, and I only use lua for neovim plugin development. Especially if the tests are going to be more complicated (like pulling source code from an open source and checking text objects on that code), I find Python easier to work with.
I'll make a pull request adding very simple basic test so you can easily understand the logic. It won't take long time for the initial one, and I'll put that in `scripts/python_tests` directory.

theHamsta:
> > I'm fine with it when it works well for you. In nvim-treesitter, we use plenary. I launches nvim for all `describe` invocations in `**/*_spec.lua`. https://github.com/nvim-treesitter/nvim-treesitter/blob/master/tests/query/injection_spec.lua
> 
> Thanks for the pointer. I am just more used to Python, and I only use lua for neovim plugin development. Especially if the tests are going to be more complicated (like pulling source code from an open source and checking text objects on that code), I find Python easier to work with. I'll make a pull request adding very simple basic test so you can easily understand the logic. It won't take long time for the initial one, and I'll put that in `scripts/python_tests` directory.

Interesting. I don't understand yet how you get the ground truth when you pull code from external repos. Do you want to describe the desired result separately in test files? For indent we checked in minimal source files https://github.com/nvim-treesitter/nvim-treesitter/tree/master/tests/indent (the actual test runner is in `common.lua`). I was rather thinking in testing the functionality than the individual queries. `nvim-treesitter` already can do query testing if you just want to test whether the queries span the correct ranges (they are already used for highlight/injection range tests), although they could be improved for ranges to mark start/end.

There are some features that I don't feel to confident about especially when there are bug reports like X doesn't work in a certain mode.

kiyoon:
I want to pull the exact same code for testing by pulling a specific commit (I do this because I just don't want to include so many random codes in the repo).

Test version 1:
- Define several sets of ground truth. Test using all modes if possible.

Test version 2:
- Include version 1 features
- Additional tests like performing textobjects at a random position and remember the results. (buffer change stats, cursor pos, selection range etc.). This is a pseudo ground-truth.
- The next commit should not change this behaviour most of the time. If this changes, we have a good starting point where to inspect the issues.

kiyoon:
@theHamsta About testing the functionality: I was thinking of performing `vaf`, `Vaf` and see if it selects the entire function, for example. This is the most high-level test, meaning it will break if anything (query, key binding, etc.) breaks, and if it works, most likely nothing went wrong under the hood.

phgz:
See https://github.com/nvim-treesitter/nvim-treesitter/pull/4015

PR #347, opened 2022-12-22
fix(julia): Remove `type_parameters` field name

savq:
See https://github.com/tree-sitter/tree-sitter-julia/pull/85 and https://github.com/nvim-treesitter/nvim-treesitter/pull/4016


theHamsta:
Thank you!

PR #349, opened 2022-12-23
fix (select): Rewrite detect_selection_mode function

phgz:
@theHamsta Could you test that with the suggested changes made in nvim-treesitter (https://github.com/nvim-treesitter/nvim-treesitter/pull/4015)?

On my end, everything seems to be working well. In particular, I can do `Vaf` or `vaf` and it will respect the selection mode. If user has put a preference for selection mode in config, then config will take precedence.

*Edit*: ~I need to remove debug print statement.~

I also reintroduced the `<cmd>` instead of `:` in mapping. We could translate it directly to `vim.keymap.set` in this PR.

### DIFF
--- a/lua/nvim-treesitter/textobjects/select.lua
+++ b/lua/nvim-treesitter/textobjects/select.lua
@@ -162,14 +162,9 @@ function M.attach(bufnr, lang)
       query = nil
     end
     if query then
-      --- Does not currently work in visual mode
-      --vim.keymap.set({ "o", "x" }, mapping, function()
-      --require("nvim-treesitter.textobjects.select").select_textobject(query)
-      --end, { buffer = buf, silent = true, remap = false, desc = desc })
-      local cmd_o = "<cmd>lua require'nvim-treesitter.textobjects.select'.select_textobject('" .. query .. "', 'o')<CR>"
-      api.nvim_buf_set_keymap(buf, "o", mapping, cmd_o, { silent = true, noremap = true, desc = desc })
-      local cmd_x = "<cmd>lua require'nvim-treesitter.textobjects.select'.select_textobject('" .. query .. "', 'x')<CR>"
-      api.nvim_buf_set_keymap(buf, "x", mapping, cmd_x, { silent = true, noremap = true, desc = desc })
+      vim.keymap.set({ "o", "x" }, mapping, function()
+        require("nvim-treesitter.textobjects.select").select_textobject(query)
+      end, { buffer = buf, silent = true, remap = false, desc = desc })
     end
   end
 end
--- a/lua/nvim-treesitter/textobjects/select.lua
+++ b/lua/nvim-treesitter/textobjects/select.lua
@@ -134,15 +134,12 @@ function M.detect_selection_mode(query_string, keymap_mode)
     selection_mode = selection_modes or "v"
   end
 
-  if method == "visual" or method == "operator-pending" then
-    -- According to "mode()" mapping, if we are in operator pending mode or visual mode,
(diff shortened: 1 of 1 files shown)


lua/nvim-treesitter/textobjects/select.lua
```diff
@@ -176,9 +166,9 @@ function M.attach(bufnr, lang)
       --vim.keymap.set({ "o", "x" }, mapping, function()
       --require("nvim-treesitter.textobjects.select").select_textobject(query)
       --end, { buffer = buf, silent = true, remap = false, desc = desc })
```
theHamsta, line 168: if `vim.keymap.set` would work (see lines above, that would be preferred

theHamsta:
Works fine for me! Even with `Vaf`. `<c-v>af` also works even though the selection only gets visible after moving the cursor (I guess that's unrelated to your changes and maybe a Neovim bug).

theHamsta (approved):

phgz:
> <c-v>af also works even though the selection only gets visible after moving the cursor

On my side, the seletion is visible right away, that's strange 🤔.

theHamsta:
> > `<c-v>af` also works even though the selection only gets visible after moving the cursor
> 
> On my side, the seletion is visible right away, that's strange thinking.

As I said, to me it looked like a nvim bug. I'm using Nightly.

phgz:
I use `NVIM v0.9.0-dev-550+g3ea1524cf`.

selection

kiyoon:
#344 happens again for me. I'm using Nvim 0.8.1. Maybe the `<cmd>` change is affecting this still.

phgz:
I replaced old mapping with `vim.keymap.set`.

@kiyoon, have you tested with https://github.com/nvim-treesitter/nvim-treesitter/pull/4015?

phgz:
For an obscure reason, `<cmd>` works correctly for repeat motion (ex. `dap.` (delete a parameter)), but with `vim.keymap.set`, it redoes the last basic action, i.e. if last parameter was 4 char long, it deletes 4 chars.

theHamsta:
Whatever works is fine for me :sweat_smile:

phgz:
I have been able to reproduce the visual glitch, but as you say it's only visual.

Well, when all is good for the changes in **nvim-treesitter**, this is ready to integrate.

theHamsta:
I merged your upstream changes. I think this is ready then. @kiyoon can you merge when this is fine for you?

kiyoon:
> I merged your upstream changes. I think this is ready then. @kiyoon can you merge when this is fine for you?

Sure, sorry for the delay, I haven't been able to test yet. I will do within a day.

kiyoon (approved):
Working well as expected. Tested `@function.inner/outer`, `@parameter.inner/outer` etc. Additionally, other functionalities like swapping arguments with `.` repeat still works. With nvim 0.8.1 it shows visual selection right away as well.

kiyoon:
I noticed something that may be good to note for the future reference.
Prior to this update, the selection will be cancelled if there's no such statement in the file (e.g. calling `@function.inner` without any function will cancel the selection).
With this update, it will ignore the mapping and still keep the original selection.

The latter makes it more consistent to the default vim behaviour (`vi(` will not do anything if there's no parentheses)

phgz:
Does it also help for incremental selection? What was the problem with it?

kiyoon:
> Does it also help for incremental selection? What was the problem with it?

I didn't test the incremental selection. I meant, it's a **good** change in behaviour that it makes it consistent to original vim. Thanks for your contribution!

theHamsta:
There were some open issues with incremental selection that I couldn't reproduce.

PR #350, opened 2022-12-27
feat(lsp_interop): Add max win height option for floating window

phgz:
Add the possibility to specify a maximum height for floating window, so it does not fill all the screen.

### DIFF
--- a/lua/nvim-treesitter/textobjects/lsp_interop.lua
+++ b/lua/nvim-treesitter/textobjects/lsp_interop.lua
@@ -45,10 +45,8 @@ function M.preview_location(location, context)
   end
 
   local config = configs.get_module "textobjects.lsp_interop"
-  local opts = {}
-  if config.max_height > 0 then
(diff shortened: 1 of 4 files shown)


lua/nvim-treesitter/textobjects/lsp_interop.lua
```diff
@@ -46,6 +46,9 @@ function M.preview_location(location, context)
 
   local config = configs.get_module "textobjects.lsp_interop"
   local opts = {}
+  if config.max_height > 0 then
```
theHamsta, line 49: couldn't you just forward all module setting to `open_floating_preview` or create a setting or create a `floating_preview_opts` key? You could key the border option for backward compatibility

phgz: I like the `floating_preview_opts` key idea, more general. Will change it soon.

theHamsta (approved):

Issue #340, opened 2022-12-17
Support `;` for repeating movements


ghostbuster91:
I am relatively new to vim, and I don't know whether this should just work but I have broke something in my config or it is just impossible to do that. If it is the latter I would like to understand the reason behind it.

**Is your feature request related to a problem? Please describe.**
I would like to be able to repeat motion `]m`/`]M` etc with `;`, the same way that I repeat motion from `f`/`F`/`t`/`T`

**Describe alternatives you've considered**
-

**Additional context**
I am aware of the existence of vim-repeat that allows repeating custom actions, I am surprised that there is nothing like that for repeating motions.

navdeeprana:
https://github.com/vim-scripts/repeatable-motions.vim

kiyoon:
> https://github.com/vim-scripts/repeatable-motions.vim
> 
> This exists, however I have not checked it.

Thanks for the pointer. Interesting plugin, although I tried it with TS textobjects and it raises a bunch of errors.

As far as I know, using the same `;` key is not possible and you need to make a separate binding for this TS textobjects, if it gets implemented.

In order to use `;` key, you need to override all `f`, `F`, `t`, `T` keys to work as usual and remember the last movement with TS textobjects. This is a dirty solution which can potentially cause problems with other plugins.. This is also how `repeatable-motions.vim` implements this.

https://github.com/vim-scripts/repeatable-motions.vim/blob/d51360269b312a34922d3de412505af069565e77/plugin/repeatable-motions.vim#L390-L393

kiyoon:
I tried several stuff.

1. Adding `{ repeatable = true }` here:
https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/37a810aaa3066f2794753547ad81c7a1fe256a66/lua/nvim-treesitter/textobjects/move.lua#L87

It works in normal mode but not in visual mode.

2. Using `tpope/vim-repeat`
```vim
map <Plug>nextfunc <cmd>TSTextobjectGotoNextStart @function.outer<CR>
silent! call repeat#set("\<Plug>nextfunc", v:count)
map <C-A-l> <Plug>nextfunc
```

It works in normal mode but not in visual mode.

theHamsta:
> Adding { repeatable = true } here:

[...]

PR #359, opened 2023-01-13
Repeat movements like `;` and `,`

kiyoon:
Implements #340 

Supports normal, visual and operator-pending modes, `v:count1` count repeat, and optionally builtin f, F, t, T mappings if you want to use the same `;` bindings.

```lua
local tstextobj_move = require "nvim-treesitter.textobjects.move"

-- Repeat movement with ; and ,
-- ensure ; goes forward and , goes backward
vim.keymap.set({ "n", "x", "o" }, ";", tstextobj_move.repeat_last_move_next)
vim.keymap.set({ "n", "x", "o" }, ",", tstextobj_move.repeat_last_move_previous)

-- vim way: ; goes to the direction you were moving.
-- vim.keymap.set({ "n", "x", "o" }, ";", tstextobj_move.repeat_last_move)
-- vim.keymap.set({ "n", "x", "o" }, ",", tstextobj_move.repeat_last_move_opposite)

-- Make builtin f, F, t, T also repeatable with ; and ,
vim.keymap.set({ "n", "x", "o" }, "f", tstextobj_move.builtin_f)
vim.keymap.set({ "n", "x", "o" }, "F", tstextobj_move.builtin_F)
vim.keymap.set({ "n", "x", "o" }, "t", tstextobj_move.builtin_t)
vim.keymap.set({ "n", "x", "o" }, "T", tstextobj_move.builtin_T)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,16 @@ vim.keymap.set({ "n", "x", "o" }, "t", ts_repeat_move.builtin_t)
 vim.keymap.set({ "n", "x", "o" }, "T", ts_repeat_move.builtin_T)
 ```
 
+You can even make a custom repeat behaviour.
+
+````lua
+local function repeat_last_move_next_end()
+  -- This repeats last query with always forward direction and end of the range.
+  return ts_repeat_move.repeat_last_move({forward = true, start = false})
+end
+vim.keymap.set({ "n", "x", "o" }, "<end>", repeat_last_move_next_end)
+```
+
 Furthermore, you can make any custom movements (e.g. from another plugin) repeatable with the same keys.
 This doesn't need to be treesitter-related.
 
@@ -187,7 +197,7 @@ local right2_repeat, left2_repeat = ts_repeat_move.make_repeatable_move_pair(rig
 
 vim.keymap.set({ "n", "x", "o" }, "l", right2_repeat)
 vim.keymap.set({ "n", "x", "o" }, "h", left2_repeat)
-```
+````
 
 Alternative way is to use a repeatable movement managing plugin such as [nvim-next](https://github.com/ghostbuster91/nvim-next).
 
--- a/lua/nvim-treesitter/textobjects/move.lua
+++ b/lua/nvim-treesitter/textobjects/move.lua
@@ -6,6 +6,10 @@ local configs = require "nvim-treesitter.configs"
 
 local M = {}
 
+M.last_move = nil
+-- { func = move, args = { ... } }
+-- { func = builtin_find, args = { ... } }
+
 local function move(query_strings, forward, start, winid)
   query_strings = shared.make_query_strings_table(query_strings)
   winid = winid or vim.api.nvim_get_current_win()
@@ -70,16 +74,157 @@ local function move(query_strings, forward, start, winid)
 end
 
 M.goto_next_start = function(query_strings)
-  move(query_strings, "forward", "start")
+  local args = { query_strings, "forward", "start" }
+  move(unpack(args))
+  M.last_move = { func = move, args = args }
 end
 M.goto_next_end = function(query_strings)
-  move(query_strings, "forward", not "start")
+  local args = { query_strings, "forward", not "start" }
+  move(unpack(args))
+  M.last_move = { func = move, args = args }
 end
 M.goto_previous_start = function(query_strings)
-  move(query_strings, not "forward", "start")
+  local args = { query_strings, not "forward", "start" }
+  move(unpack(args))
+  M.last_move = { func = move, args = args }
 end
 M.goto_previous_end = function(query_strings)
-  move(query_strings, not "forward", not "start")
+  local args = { query_strings, not "forward", not "start" }
+  move(unpack(args))
+  M.last_move = { func = move, args = args }
+end
+
+-- implements naive f, F, t, T with repeat support
+local function builtin_find(forward, inclusive, char, repeating)
+  -- forward = true -> f, t
+  -- inclusive = false -> t, T
+  -- if repeating with till (t or T, inclusive = false) then search from the next character
+  -- returns nil if cancelled or char
+  char = char or vim.fn.nr2char(vim.fn.getchar())
+  repeating = repeating or false
+
+  if char == vim.fn.nr2char(27) then
+    -- escape
+    return nil
+  end
+
+  local line = vim.api.nvim_get_current_line()
+  local cursor = vim.api.nvim_win_get_cursor(0)
+
+  -- find the count-th occurrence of the char in the line
+  local found
+  for _ = 1, vim.v.count1 do
+    if forward then
+      if not inclusive and repeating then
+        cursor[2] = cursor[2] + 1
+      end
+      found = line:find(char, cursor[2] + 2, true)
+    else
+      -- reverse find from the cursor position
+      if not inclusive and repeating then
+        cursor[2] = cursor[2] - 1
+      end
+
+      found = line:reverse():find(char, #line - cursor[2] + 1, true)
+      if found then
+        found = #line - found + 1
+      end
+    end
+
+    if not found then
+      return char
+    end
+
+    if forward then
+      if not inclusive then
+        found = found - 1
+      end
+    else
+      if not inclusive then
+        found = found + 1
+      end
+    end
+
+    cursor[2] = found - 1
+    repeating = true -- after the first iteration, search from the next character if not inclusive.
+  end
+
+  -- move to the found position
+  vim.api.nvim_win_set_cursor(0, { cursor[1], cursor[2] })
+  return char
+end
+
+M.builtin_f = function()
+  local char = builtin_find("forward", "inclusive")
+  if builtin_find ~= nil then
+    M.last_move = { func = builtin_find, args = { "forward", "inclusive", char, "repeating" } }
+  end
+end
+
+M.builtin_F = function()
+  local char = builtin_find(not "forward", "inclusive")
+  if builtin_find ~= nil then
+    M.last_move = { func = builtin_find, args = { not "forward", "inclusive", char, "repeating" } }
+  end
+end
+
+M.builtin_t = function()
+  local char = builtin_find("forward", not "inclusive")
+  if builtin_find ~= nil then
+    M.last_move = { func = builtin_find, args = { "forward", not "inclusive", char, "repeating" } }
+  end
+end
+
+M.builtin_T = function()
+  local char = builtin_find(not "forward", not "inclusive")
+  if builtin_find ~= nil then
+    M.last_move = { func = builtin_find, args = { not "forward", not "inclusive", char, "repeating" } }
+  end
+end
+
+M.repeat_last_move = function()
+  if M.last_move then
+    M.last_move.func(unpack(M.last_move.args))
+  end
+end
+
+M.repeat_last_move_opposite = function()
+  if M.last_move then
+    local args = { unpack(M.last_move.args) } -- copy the table
+    if M.last_move.func == move then
+      args[2] = not args[2] -- reverse the direction
+    else
+      -- builtin_find
+      args[1] = not args[1] -- reverse the direction
+    end
+    M.last_move.func(unpack(args))
+  end
+end
+
+M.repeat_last_move_next = function()
+  if M.last_move then
+    local args = { unpack(M.last_move.args) } -- copy the table
+    if M.last_move.func == move then
+      args[2] = "forward" -- set the direction to forward
+    else
+      -- builtin_find
+      args[1] = "forward" -- set the direction to forward
+    end
+    M.last_move.func(unpack(args))
+  end
+end
+
+M.repeat_last_move_previous = function()
+  if M.last_move then
+    local args = { unpack(M.last_move.args) } -- copy the table
+    if M.last_move.func == move then
+      args[2] = not "forward" -- set the direction to backward
+    else
+      -- builtin_find
+      args[1] = not "forward" -- set the direction to backward
+    end
+    M.last_move.func(unpack(args))
+  end
 end
 
 local nxo_mode_functions = { "goto_next_start", "goto_next_end", "goto_previous_start", "goto_previous_end" }
@@ -115,6 +260,30 @@ M.commands = {
       "-nargs=1",
       "-complete=custom,nvim_treesitter_textobjects#available_textobjects",
     },
+  },
+  TSTextobjectRepeatLastMove = {
+    run = M.repeat_last_move,
+  },
+  TSTextobjectRepeatLastMoveOpposite = {
+    run = M.repeat_last_move_opposite,
+  },
+  TSTextobjectRepeatLastMoveNext = {
+    run = M.repeat_last_move_next,
+  },
+  TSTextobjectRepeatLastMovePrevious = {
+    run = M.repeat_last_move_previous,
+  },
+  TSTextobjectBuiltinf = {
+    run = M.builtin_f,
+  },
+  TSTextobjectBuiltinF = {
+    run = M.builtin_F,
+  },
+  TSTextobjectBuiltint = {
+    run = M.builtin_t,
+  },
+  TSTextobjectBuiltinT = {
+    run = M.builtin_T,
   },
 }
 
--- a/lua/nvim-treesitter/textobjects/move.lua
+++ b/lua/nvim-treesitter/textobjects/move.lua
@@ -17,7 +17,7 @@ M.last_move = nil
 -- move_fn's first argument must be a boolean indicating whether to move forward (true) or backward (false)
 -- Then you can use four functions:
 --   M.repeat_last_move
---   M.repeat_last_move_oppisite
+--   M.repeat_last_move_opposite
 --   M.repeat_last_move_next
 --   M.repeat_last_move_previous
 function M.make_repeatable_move(move_fn)
--- a/lua/nvim-treesitter/textobjects/move.lua
+++ b/lua/nvim-treesitter/textobjects/move.lua
@@ -103,23 +103,51 @@ end
 local move_repeatable = repeatable_move.make_repeatable_move(move)
 
 M.goto_next_start = function(query_strings_regex, query_group)
-  move_repeatable { forward = true, start = true, query_strings_regex = query_strings_regex, query_group = query_group }
+  move_repeatable {
+    forward = true,
+    start = true,
+    query_strings_regex = query_strings_regex,
+    query_group = query_group,
+  }
 end
 M.goto_next_end = function(query_strings_regex, query_group)
-  move_repeatable { forward = true, start = true, query_strings_regex = query_strings_regex, query_group = query_group }
+  move_repeatable {
+    forward = true,
+    start = false,
+    query_strings_regex = query_strings_regex,
+    query_group = query_group,
+  }
 end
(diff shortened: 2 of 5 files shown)


kiyoon:
The consistency tests all pass. However note that currently it only performs movements in visual mode.

lua/nvim-treesitter/textobjects/move.lua
```diff
@@ -70,16 +74,157 @@ local function move(query_strings, forward, start, winid)
...
+  if M.last_move then
+    local args = { unpack(M.last_move.args) } -- copy the table
+    if M.last_move.func == move then
+      args[2] = not "forward" -- set the direction to backward
+    else
+      -- builtin_find
+      args[1] = not "forward" -- set the direction to backward
+    end
+    M.last_move.func(unpack(args))
+  end
```
ghostbuster91, line 227: Could you explain to me why there are four methods: `repeat_last_move`, `repeat_last_move_opposite`, `repeat_last_move_next` and `repeat_last_move_previous` instead of just having two: `repeat_last_move` and `repeat_last_move_opposite`?

kiyoon: `repeat_last_move` goes to the last direction. If you moved backwards, it will move backwards.  
`repeat_last_move_next` ensures that it will move forward.

lua/nvim-treesitter/textobjects/move.lua
```diff
@@ -70,16 +74,157 @@ local function move(query_strings, forward, start, winid)
...
+    -- escape
+    return nil
+  end
+
+  local line = vim.api.nvim_get_current_line()
+  local cursor = vim.api.nvim_win_get_cursor(0)
+
+  -- find the count-th occurrence of the char in the line
+  local found
+  for _ = 1, vim.v.count1 do
```
theHamsta, line 116: I know the other plugin used this code too. Could you explain why we can't use `normal! f`?

lua/nvim-treesitter/textobjects/move.lua
```diff
@@ -70,16 +74,157 @@ local function move(query_strings, forward, start, winid)
...
+  char = char or vim.fn.nr2char(vim.fn.getchar())
+  repeating = repeating or false
+
+  if char == vim.fn.nr2char(27) then
+    -- escape
+    return nil
+  end
+
+  local line = vim.api.nvim_get_current_line()
+  local cursor = vim.api.nvim_win_get_cursor(0)
```
theHamsta, line 112: our move functions can use `winid`. Wouldn't it be useful to have builtin_find support it as well?

lua/nvim-treesitter/textobjects/move.lua
```diff
@@ -116,6 +261,30 @@ M.commands = {
       "-complete=custom,nvim_treesitter_textobjects#available_textobjects",
     },
   },
```
theHamsta, line 263: for stuff like this it would be useful to add unit tests. Similar to the ones added here for a bug in swap nodes

https://github.com/nvim-treesitter/nvim-treesitter/blob/51d220e0bd8922caabcb3e5de936fc599a1bd210/tests/unit/ts_utils_spec.lua#L110-L148

You could quickly test different combinations and if there will be ever a bug with the current implementation. The PR could easily explain what went wrong. Plenary tests don't require a lot of code to set them up. Just set example test and apply the functions that are just added.

We can defer plenary tests to a different tests, but I think that they tackle a different problem that your Python snapshot tests.

lua/nvim-treesitter/textobjects/move.lua
kiyoon: The problem is we want to repeat, without having to enter the same character again.

lua/nvim-treesitter/textobjects/move.lua
kiyoon: Done!

lua/nvim-treesitter/textobjects/move.lua
kiyoon: My python tests can generate snapshots, but it doesn't mean that's the only option. We can just manually create the yaml file that describe the unit test. If you don't mind I can add the builtin movements to that python tests as we have a good starting point.

If you think it's still better to have a plenary test in the repo, is it possible for you to create a basic template because you're more experienced with that?

lua/nvim-treesitter/textobjects/move.lua
kiyoon: For example if you do `f$` which will find the dollar sign, pressing `;` should  just follow the next dollar sign, not the next function or something else. If we do `normal! f` there's no way to know which character has been searched.

kiyoon: Oh sorry, I get your point now. Ignore the comments above. I'll test if it works with normal!

kiyoon: The problem is till `t` commands.

From text
```
treesitter
```

You run `tt` and it goes before the second t. Running `tt` again wouldn't change the position, but pressing `;` will move it to the before-third-t.

So, if we want to do `normal! t` then we need to go to the right first with `normal! l` and then performa `normal! t`.
The problem is that if we didn't find anything in that line, it shouldn't move to the right. So we need to detect if it found something. However with `normal! t` it is not easy to detect it and cancel it. Only way is to check the cursor position and then move back which will create an unnecessary cursor move forward and backward.

kiyoon: Another issue is, if we want to get the position of the cursor, we need winid. I don't think there's a way to call `vim.cmd[[normal! f]]` with respect to the winid?

lua/nvim-treesitter/textobjects/move.lua
theHamsta: The advantage of plenary is that you don't need to do any setup. Just run `:PlenaryBustedFile %` to run the currently opened test file. For CI you just need to run a headless nvim run `-c ":PlenaryBustedDirectory`. If you're trying to do the same thing in Python you would need to implement all the logic they have in plenary. Most contributors would know how they work because Neovim also uses busted and many plugins use plenary for testing.

Just create a file `foo_spec.lua` with
```lua
describe("foo suite", function()
  it("can succeed", function()
    assert(1 + 1 == tonumber(vim.fn.eval("1 + 1")))
  end)
  it("it can fail", function()
    assert(1 + 1 == tonumber(vim.fn.eval("1 + 2")))
  end)
end)
```
is enough for a test suite. We don't need to install any dependency apart from the plenary plugin to run tests in Neovim. So I think adding plenary in addition to your Python tests wouldn't add to much bloat and overhead. Testing Lua functions directly in process is just more convent than to serialize and de-serialize over rtp. I'm talking about those kind of sanity checks that you would execute on two line examples. Python might be a better choice when you want to test whole repos or automated test where you only place example files in the repo.

But let's leave that for a another PR.

kiyoon: I also agree that learning something standardised and native to the program is better for everyone. Yeah let's have both plenary and python tests running, and they can both test different things. Python can test more of actual functionality by using feedkeys and plenary can test actual function implementations. I will take a look at it when I have time. Thank you!

lua/nvim-treesitter/textobjects/move.lua
theHamsta: So the behavior of this is `tt` but without considering the current position. Ok. 

I think you can do run a command also in the context of a certain window. But maybe that's simpler.

theHamsta (approved):
I like how this is extensive by using `M.last_move` outside of this plugin.

lua/nvim-treesitter/textobjects/move.lua
```diff
@@ -70,16 +74,158 @@ local function move(query_strings, forward, start, winid)
 end
 
 M.goto_next_start = function(query_strings)
```
theHamsta, line 76: can't this be generalized? By transforming wrapping a function? This could be used to make external functions of the user repeatable. Like
``` lua
function M.make_repeatable_move(move_fn)
    return function(...)
         M.last_move = { func = move_fn, args = {...} }
         move_fn(...)
    end
end

Comment (theHamsta) [reply lua/nvim-treesitter/textobjects/move.lua:76]: or even 
```
function M.make_repeatable_move(move_fn)
    return function(...)
         M.last_move = function() move_fn(...) end
         move_fn(...)
    end
end
```

Comment (kiyoon) [reply lua/nvim-treesitter/textobjects/move.lua:76]: The only problem is that when you repeat movement you need to know the direction. Unfortunately, the argument position of `forward` is in a random place at the moment. So in `M.repeat_last_move_next` for example I had to recognise the name of the function and change the `forward` value to `true`.

Comment (kiyoon) [reply lua/nvim-treesitter/textobjects/move.lua:76]: If we can move the forward argument position to always the first maybe we can use this design

Comment (kiyoon) [reply lua/nvim-treesitter/textobjects/move.lua:76]: Or, just remove all arguments except accepting one opts table with `forward=` key.

File: lua/nvim-treesitter/textobjects/move.lua (hunk 2 above)
Comment (kiyoon) [reply lua/nvim-treesitter/textobjects/move.lua:117]: > So the behavior of this is `tt` but without considering the current position. Ok.
> 
> I think you can do run a command also in the context of a certain window. But maybe that's simpler.

Yes, but only when `repeating=true` to match the vim's behaviour.

File: lua/nvim-treesitter/textobjects/move.lua (hunk 5 above)
Comment (theHamsta) [reply lua/nvim-treesitter/textobjects/move.lua:76]: ah, I get it now. Because you can repeat but in opposite direction.

Comment (theHamsta) [reply lua/nvim-treesitter/textobjects/move.lua:76]: so you would need to define `make_repeatable_move` in terms of a `forward_fn` and `backwarn_fn` pair.

Comment (kiyoon) [reply lua/nvim-treesitter/textobjects/move.lua:76]: Yes, but I think making it extensible is a good idea. Other plugins can just make a wrapper function that follows the rule (opt should have forward key).

I also want to use it in my plugin.

Comment (kiyoon) [reply lua/nvim-treesitter/textobjects/move.lua:76]: > so you would need to define `make_repeatable_move` in terms of a `forward_fn` and `backwarn_fn` pair.
> 
> We can leave it as is. So every function to extend from user space must have the direction as boolish first parameter to work.

Separating with two functions is also a good idea. Should we choose the table argument design though?(maybe more future proof as it can accept more than forward, like start and end etc. if it matters)

Comment (ghostbuster91) [reply lua/nvim-treesitter/textobjects/move.lua:76]: Yes!!! I took that idea already even further: https://github.com/ghostbuster91/nvim-next

I don't have any experience in developing nvim plugins so any feedback appreciated. If you would like to help me with this effort just let me know and I would be happy to add you as maintainers :)

Comment (theHamsta) [reply lua/nvim-treesitter/textobjects/move.lua:76]: yes, it should be fine as it is now. But it would be nice to have a function that would automatically turn a user's function into a repeatable one, so that they don't need to understand how `last_move` works. Docs can explain that the first parameter needs to be the direction.

Comment (theHamsta) [reply lua/nvim-treesitter/textobjects/move.lua:76]: I like @ghostbuster91's approach

Comment (kiyoon) [reply lua/nvim-treesitter/textobjects/move.lua:76]: never mind, we can make it accept both boolean and table in the future so it is still future proof. Implementing this now.

Comment (kiyoon) [reply lua/nvim-treesitter/textobjects/move.lua:95]: I just pushed the change.

Comment (ghostbuster91) [reply lua/nvim-treesitter/textobjects/move.lua:95]: @theHamsta I am glad to hear that. I don't think that it is possible to have this repeatable magic without any manual call (unless that is somehow implemented directly in the core). But again I am quite new to this space so I might be missing something obvious.

File: lua/nvim-treesitter/textobjects/move.lua [hunk 6]
```diff
@@ -6,7 +6,28 @@ local configs = require "nvim-treesitter.configs"
...
+-- { func = builtin_find, args = { ... } }
+-- register any other function, but make sure the the first args is `forward` boolean.
+
+-- If you are a plugin developer,
+-- you can register your own independent move functions using this
+-- even if they are not related to treesitter-textobjects.
+-- move_fn's first argument must be a boolean indicating whether to move forward (true) or backward (false)
+-- Then you can use four functions:
+--   M.repeat_last_move
+--   M.repeat_last_move_oppisite
```
theHamsta, line 20: typo

lua/nvim-treesitter/textobjects/move.lua
kiyoon: > @theHamsta I am glad to hear that. I don't think that it is possible to have this repeatable magic without any manual call (unless that is somehow implemented directly in the core). But again I am quite new to this space so I might be missing something obvious.

I just saw your repo. Thanks for sharing!  
It would have been really nice if it was nvim's builtin to register repeatability, or plugins like plenary that everyone has.  
However, requiring more dependencies and further setup call isn't ideal.

The idea is that instead of requiring a plugin and setup call, you can make plugin developers use our function to make it repeatable, or making users turn existing plugins' movement into a repeatable one. It is possible with `make_repeatable_move` now.

theHamsta: @kiyoon so your idea was to make the first arg a table to be more future proof?

I think there are the following approaches:
- make function magical by having it the first arg boolean
- make function magical by having it the first arg a table
- make function pair magical and use both return values to have two magical functions
- not have magical functions but set up everything via config

kiyoon: I think the first arg can be a boolean for now, and if we want to extend it we can use a table and check the type of the first arg. So the approach 1 and 2 are kind of similar.

kiyoon: We can provide option 3 along with the one we have. I'll make this as well.

kiyoon: I just pushed the change.

kiyoon: I think option 4 is limiting because other plugins can't use our functions under the hood, and users need to config everything right?

kiyoon: I don't know if this is a general concern. If move_fn accepts buf_id or win_id, the repeat will only perform at the same buffer not the current buffer. Should we have a feature to automatically set them to 0?

kiyoon: > I don't know if this is a general concern. If move_fn accepts buf_id or win_id, the repeat will only perform at the same buffer not the current buffer. Should we have a feature to automatically set them to 0?

But honestly users wouldn't mind using the move command again and then perform repeat. Who would move around the buffers and use repeat?

kiyoon: > * make function magical by having it the first arg boolean
> * make function magical by having it the first arg a table

If we introduce the table stuff later, then whoever is directly accessing `M.last_move` to perform whatever custom stuff they want to do will face problems later.  
So we can reserve the table usage in the doc and implement it now, so it doesn't cause any confusion in the future.

ghostbuster91: > The idea is that instead of requiring a plugin and setup call, you can make plugin developers use our function to make it repeatable, or making users turn existing plugins' movement into a repeatable one. It is possible with make_repeatable_move now.

Sure, the setup call is not necessary, it is just a convenient way of setting it. There is already an option to just use the wrapper method and turn any function into a repeatable pair, just as you did (I've updated the readme to make it clear). 

I think that it is better to have it as a separate plugin than as a part of nvim-treesitter-textobjects since not everyone needs to have your plugin installed in order to benefit from custom repeatable movements. IMO It also fits better into nvim philosophy where you have many small things that you compose together.

kiyoon: Update: I just noticed that supporting both types is infeasible as we don't know how the custom function behaves.  
Should we use a boolean or a table?

kiyoon: @ghostbuster91 That's true. I think there are many plugins that are supposedly doing this, but for some reason I can't get them to work nicely with lua functions.

kiyoon: I think this plugin can offer a native solution to do this, so everyone using treesitter textobjects benefit from this functionality. However, it is completely optional whether you want to extend it to a separate plugin and utilise other plugins at the same time. This way, most people wouldn't have to change settings to use repeat motions, but some people can optionally do this step to make it more customised.

theHamsta: @kiyoon you think `M.last_move` can be private now that users have some functions that doesn't need them understand `last_move`?

theHamsta (changes requested):
Could you explain this new functionality in the README?

Thank you a lot for this awesome feature!

theHamsta:
@ghostbuster91 do you want us to mention your plugin? I think this feature is small enough to be integrated into our move module, but your plugin could provide configurations for commonly used move plugins (like hop.nvim etc.) since we can't obviously maintain this for more than our `move` module.

kiyoon:
@theHamsta Should we make this feature a separate file from `move.lua` so it is more clear which part is treesitter related and which not?

I can separate this to `repeatable_move.lua` including all the builtin movements so it's clear why they exist.

theHamsta:
> @theHamsta Should we make this feature a separate file from `move.lua` so it is more clear which part is treesitter related and which not?
> 
> I can separate this to `repeatable_move.lua` including all the builtin movements so it's clear why they exist.

Sure, go ahead. I though about this too.

lua/nvim-treesitter/textobjects/move.lua
kiyoon: > @kiyoon you think `M.last_move` can be private now that users have some functions that doesn't need them understand `last_move`?

There are some more complicated move behaviours where they want to change the arguments for the repeat functionality.
Builtin t etc. are the examples where you want the repeat functionality to be different from just executing the same function again.
So maybe leaving it open will make it more general to be used by others.

Wait I can edit your comment as well? I did this accidentally and I'm super confused

kiyoon: @theHamsta Furthermore, people can even customise the last move bahaviour by making their custom function that uses `M.last_move`.

For example, they can make `repeat_last_move_next_start_inner`

ghostbuster91:
> do you want us to mention your plugin?

@theHamsta That would be cool, thanks!

> your plugin could provide configurations for commonly used move plugins (like hop.nvim etc.

yep, that is my plan :)

kiyoon:
@ghostbuster91 Sure, that's no problem. I'll mention your plugin in the README

kiyoon:
This is ready

README.md
```diff
@@ -147,6 +147,60 @@ require'nvim-treesitter.configs'.setup {
...
+
+-- make sure forward function comes first
+local right2_repeat, left2_repeat = ts_repeat_move.make_repeatable_move_pair(right2, left2)
+-- Or, use `make_repeatable_move` or `set_last_move` functions. See the code for instructions.
+
+vim.keymap.set({ "n", "x", "o" }, "l", right2_repeat)
+vim.keymap.set({ "n", "x", "o" }, "h", left2_repeat)
+````
+
+Alternative way is to use a repeatable movement managing plugin such as [nvim-next](https://github.com/ghostbuster91/nvim-next).
```
ghostbuster91, line 202: :heart:

theHamsta (approved):

PR #360, opened 2023-01-14
Gradual movement (regex query string, goto_next and goto_previous without specifying start / end)

kiyoon:
Before merging #359 I wanted to show you this.

At the moment, my keymaps are quite polluted.

```lua
      goto_next_start = {
        ["]m"] = "@function.outer",
        ["]f"] = "@call.outer",
        ["]d"] = "@conditional.outer",
        ["]o"] = "@loop.outer",
        ["]s"] = "@statement.outer",
        ["]a"] = "@parameter.outer",
        ["]c"] = "@comment.outer",
        ["]b"] = "@block.outer",
        ["]l"] = { query = "@class.outer", desc = "next class start" },
        ["]]m"] = "@function.inner",
        ["]]f"] = "@call.inner",
        ["]]d"] = "@conditional.inner",
        ["]]o"] = "@loop.inner",
        ["]]a"] = "@parameter.inner",
        ["]]b"] = "@block.inner",
        ["]]l"] = { query = "@class.inner", desc = "next class start inner" },
      },
      goto_next_end = {
        ["]M"] = "@function.outer",
        ["]F"] = "@call.outer",
        ["]D"] = "@conditional.outer",
        ["]O"] = "@loop.outer",
        ["]S"] = "@statement.outer",
        ["]A"] = "@parameter.outer",
        ["]C"] = "@comment.outer",
        ["]B"] = "@block.outer",
        ["]L"] = "@class.outer",
        ["]]M"] = "@function.inner",
        ["]]F"] = "@call.inner",
        ["]]D"] = "@conditional.inner",
        ["]]O"] = "@loop.inner",
        ["]]A"] = "@parameter.inner",
        ["]]B"] = "@block.inner",
        ["]]L"] = "@class.inner",
      },
      goto_previous_start = {
        ["[m"] = "@function.outer",
        ["[f"] = "@call.outer",
        ["[d"] = "@conditional.outer",
        ["[o"] = "@loop.outer",
        ["[s"] = "@statement.outer",
        ["[a"] = "@parameter.outer",
        ["[c"] = "@comment.outer",
        ["[b"] = "@block.outer",
        ["[l"] = "@class.outer",
        ["[[m"] = "@function.inner",
        ["[[f"] = "@call.inner",
        ["[[d"] = "@conditional.inner",
        ["[[o"] = "@loop.inner",
        ["[[a"] = "@parameter.inner",
        ["[[b"] = "@block.inner",
        ["[[l"] = "@class.inner",
      },
      goto_previous_end = {
        ["[M"] = "@function.outer",
        ["[F"] = "@call.outer",
        ["[D"] = "@conditional.outer",
        ["[O"] = "@loop.outer",
        ["[S"] = "@statement.outer",
        ["[A"] = "@parameter.outer",
        ["[C"] = "@comment.outer",
        ["[B"] = "@block.outer",
        ["[L"] = "@class.outer",
        ["[[M"] = "@function.inner",
        ["[[F"] = "@call.inner",
        ["[[D"] = "@conditional.inner",
        ["[[O"] = "@loop.inner",
        ["[[A"] = "@parameter.inner",
        ["[[B"] = "@block.inner",
        ["[[L"] = "@class.inner",
      },
```

This patch enables you to group things together if you wish.  
```lua
      goto_next = {
        ["]m"] = "@function.*",
        ["]f"] = "@call.*",
        ["]o"] = { query = { "@loop.*", "@conditional.*" }, desc = "Next loop or conditional" },
        ["]s"] = "@statement.*",
        ["]a"] = "@parameter.*",
        ["]c"] = "@comment.*",
        ["]b"] = "@block.*",
        ["]l"] = { query = "@class.*", desc = "next class start/end" },
      },
      goto_previous = {
        ["[m"] = "@function.*",
        ["[f"] = "@call.*",
        ["[o"] = { query = { "@loop.*", "@conditional.*" }, desc = "Previous loop or conditional" },
        ["[s"] = "@statement.*",
        ["[a"] = "@parameter.*",
        ["[c"] = "@comment.*",
        ["[b"] = "@block.*",
        ["[l"] = { query = "@class.*", desc = "previous class start/end" },
      },
```

(I just noticed that you can already add multiple queries, however it isn't documented)  

Along with the power of repeatable movements, for some people it may be a good bonus as they can go to the right part they want without going for a complicated keymapping.

The regex queries are supported in the original `goto_next_start` etc. as well.  
It will stay mostly compatible with the previous setup, except that now the dot will match any character. If you have a custom query named `@functionQinner` then it will be matched with `@function.inner`

### DIFF
--- a/lua/nvim-treesitter/textobjects/move.lua
+++ b/lua/nvim-treesitter/textobjects/move.lua
@@ -24,17 +24,8 @@ local function move(query_strings_regex, query_group, forward, start, winid)
   end
   winid = winid or vim.api.nvim_get_current_win()
   local bufnr = vim.api.nvim_win_get_buf(winid)
-
-  -- Get query strings from regex
-  local available_textobjects = shared.available_textobjects(parsers.get_buf_lang(bufnr), query_group)
-  local query_strings = {}
-  for _, query_string_regex in ipairs(query_strings_regex) do
-    for _, available_textobject in ipairs(available_textobjects) do
-      if string.match("@" .. available_textobject, query_string_regex) then
-        table.insert(query_strings, "@" .. available_textobject)
-      end
-    end
-  end
+  local query_strings =
+    shared.get_query_strings_from_regex(query_strings_regex, query_group, parsers.get_buf_lang(bufnr))
 
   local config = configs.get_module "textobjects.move"
 
--- a/lua/nvim-treesitter/textobjects/move.lua
+++ b/lua/nvim-treesitter/textobjects/move.lua
@@ -87,7 +87,6 @@ local function move(query_strings_regex, forward, start, winid)
 
         if current_match then
           local score = scoring_function(start_, current_match)
-          print(score)
           if not best_match then
             best_match = current_match
             best_score = score
--- a/lua/nvim-treesitter/textobjects/move.lua
+++ b/lua/nvim-treesitter/textobjects/move.lua
@@ -3,21 +3,61 @@ local attach = require "nvim-treesitter.textobjects.attach"
 local shared = require "nvim-treesitter.textobjects.shared"
 local queries = require "nvim-treesitter.query"
 local configs = require "nvim-treesitter.configs"
+local parsers = require "nvim-treesitter.parsers"
 
 local M = {}
 
-local function move(query_strings, forward, start, winid)
-  query_strings = shared.make_query_strings_table(query_strings)
+local function move(query_strings_regex, forward, start, winid)
+  -- start: bool or nil. If true, choose the start of the node, and false is for the end.
+  -- If nil, it considers both and chooses the closer side.
+  query_strings_regex = shared.make_query_strings_table(query_strings_regex)
+  local starts
+  if start == nil then
+    starts = { true, false }
+  else
+    if start then
+      starts = { true }
+    else
+      starts = { false }
+    end
+  end
   winid = winid or vim.api.nvim_get_current_win()
   local bufnr = vim.api.nvim_win_get_buf(winid)
 
+  -- Get query strings from regex
+  local available_textobjects = shared.available_textobjects(parsers.get_buf_lang(bufnr))
+  local query_strings = {}
+  for _, query_string_regex in ipairs(query_strings_regex) do
+    for _, available_textobject in ipairs(available_textobjects) do
+      if string.match("@" .. available_textobject, query_string_regex) then
+        table.insert(query_strings, "@" .. available_textobject)
+      end
+    end
+  end
+
   local config = configs.get_module "textobjects.move"
-  local function filter_function(match)
+
+  -- score is a byte position.
+  local function scoring_function(start_, match)
+    local score, _
+    if start_ then
+      _, _, score = match.node:start()
+    else
+      _, _, score = match.node:end_()
+    end
+    if forward then
+      return -score
+    else
+      return score
+    end
+  end
+
+  local function filter_function(start_, match)
     local range = { match.node:range() }
     local row, col = unpack(vim.api.nvim_win_get_cursor(winid))
     row = row - 1 -- nvim_win_get_cursor is (1,0)-indexed
 
-    if not start then
+    if not start_ then
       if range[4] == 0 then
         range[1] = range[3] - 1
         range[2] = range[4]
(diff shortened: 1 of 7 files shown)


theHamsta:
How complicated would it be it integrate https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/60/files in your PR for `move`?

lua/nvim-treesitter/textobjects/move.lua
```diff
@@ -33,39 +73,35 @@ local function move(query_strings, forward, start, winid)
...
+      for _, start_ in ipairs(starts) do
+        local current_match = queries.find_best_match(bufnr, query_string, "textobjects", function(match)
+          return filter_function(start_, match)
+        end, function(match)
+          return scoring_function(start_, match)
+        end)
+
+        if current_match then
+          local score = scoring_function(start_, current_match)
+          print(score)
```
theHamsta, line 90: debug statement

kiyoon:
> How complicated would it be it integrate https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/60/files in your PR for `move`? At the moment, you always call `"textobjects"` for `find_best_match`, would it be possible to allow queryies from other queries types as well?

Any reason it hasn't been merged?

Where should the `locals.scm` file be located in order to test this?

lua/nvim-treesitter/textobjects/move.lua
kiyoon: Done

theHamsta:
> > How complicated would it be it integrate https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/60/files in your PR for `move`? At the moment, you always call `"textobjects"` for `find_best_match`, would it be possible to allow queryies from other queries types as well?
> 
> Any reason it hasn't been merged?
> 
> Where should the `locals.scm` file be located in order to test this?

It isn't about `"locals"` in specifc, in order for textobjects to work, you need to have `queries/<lang>/textobjects.scm` in you path. The idea of this PR is to also allow `queries/<lang>/<something>.scm`. `nvim-treesitter` provides queries for `highlights.scm`/`locals.scm`/`indent.scm`/`folds.scm`. So all of these should work. They just need to be in your runtimepath, e.g. relative to some plugin path.

I was a bit hesitant whether we should allow this right now and now it is a bit out-dated. But since you're touching the config anyways you could also make it more general. At least it should be possible to add alternative query groups in retroperspecitve

kiyoon:
@theHamsta I see, the scope movement is actually amazing!  
Now I added the query group to the movement.  
It is only in movement atm and not in others yet.

kiyoon:
Should I make multiple query groups selectable? I feel like this will make it super complicated and most people wouldn't really select from multiple groups?

kiyoon:
Finished integrating #60

kiyoon:
Tests passing

theHamsta (approved):
Wow, cool! I actually only meant to consider it for movement to see whether the config stays compatible without adding too much code. But cool that you made it working for all categories

lua/nvim-treesitter/textobjects/select.lua
```diff
@@ -145,33 +146,52 @@ function M.detect_selection_mode(query_string, keymap_mode)
...
+    if not available then
+      query_string = nil
+    end
+
+    if query_string then
+      local cmd_o = function()
+        M.select_textobject(query_string, query_group, "o")
+      end
+      local cmd_x = function()
+        M.select_textobject(query_string, query_group, "x")
```
theHamsta, line 185: @kiyoon weren't lua functions causing problems in visual mode? There was a reason why this still was a string.

kiyoon: @theHamsta I don't know, I think maybe before the patch from nvim-treesitter there were lots of weird bugs even with `<cmd>` and now it seems okay. It at least passes the tests with visual selection

theHamsta: that's good. glad that we can finally get rid of this vimscript artifact.

kiyoon: @theHamsta By the way, when does the `detach` function exactly called? I think my workflow is usually limited to having not many buffers at the same time, but it doesn't seem like it's being called whenever you close the buffer? Or I am doing something wrong

Issue #365, opened 2023-01-17
Error when sourcing a session


scheatkode:
**Describe the bug**
An error occurs on sourcing a `Session.vim` file.

**To Reproduce**
1. Open a file or two
2. `:mksession`
3. Restart Neovim
4. `:source Session.vim`

**Expected behavior**
No error.

**Output of `:checkhealth nvim-treesitter`**

<details>
<code>

nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - OK: `tree-sitter` found 0.20.7 (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v18.13.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: cc (GCC) 12.2.0
  - OK: Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

## OS Info:
{
  machine = "x86_64",
  release = "6.1.5-arch2-1",
  sysname = "Linux",
  version = "#1 SMP PREEMPT_DYNAMIC Thu, 12 Jan 2023 22:42:33 +0000"
}

## Parser/Features         H L F I J
  - markdown            ✓ . ✓ . ✓
  - norg                . . . . .
  - lua                 ✓ ✓ ✓ ✓ ✓
  - bash                ✓ ✓ ✓ . ✓
  - sql                 ✓ . . . ✓
  - make                ✓ . ✓ . ✓
  - rust                ✓ ✓ ✓ ✓ ✓
  - css                 ✓ . ✓ ✓ ✓
  - html                ✓ ✓ ✓ ✓ ✓
  - javascript          ✓ ✓ ✓ ✓ ✓
  - jsonc               ✓ ✓ ✓ ✓ ✓
  - scss                ✓ . ✓ ✓ .
  - svelte              ✓ . ✓ ✓ ✓
  - typescript          ✓ ✓ ✓ ✓ ✓
  - yaml                ✓ ✓ ✓ ✓ ✓
  - prisma              ✓ . . . .
  - gitattributes       ✓ . . . ✓
  - cpp                 ✓ ✓ ✓ ✓ ✓
  - jq                  ✓ . . . ✓
  - go                  ✓ ✓ ✓ ✓ ✓
[...]

PR #363, opened 2023-01-17
mapcheck before keymap.del

kiyoon:
Resolves #362

PR #366, opened 2023-01-17
fix keymap mode type error

kiyoon:
Fix type mismatch error introduced in #363 and reported in #365

### DIFF
--- a/lua/nvim-treesitter/textobjects/attach.lua
+++ b/lua/nvim-treesitter/textobjects/attach.lua
@@ -12,8 +12,13 @@ local function make_dot_repeatable(fn)
 end
 
 function M.make_attach(functions, submodule, keymap_modes, opts)
-  keymap_modes = keymap_modes or "n"
+  if type(keymap_modes) == "string" then
+    keymap_modes = { keymap_modes }
+  elseif type(keymap_modes) ~= "table" then
+    keymap_modes = { "n" }
+  end
   opts = opts or {}
+
   return function(bufnr, lang)
     lang = lang or parsers.get_buf_lang(bufnr)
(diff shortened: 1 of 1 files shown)


lua/nvim-treesitter/textobjects/attach.lua
```diff
@@ -12,8 +12,13 @@ local function make_dot_repeatable(fn)
...
 
 function M.make_attach(functions, submodule, keymap_modes, opts)
-  keymap_modes = keymap_modes or "n"
+  if type(keymap_modes) == "string" then
+    keymap_modes = { keymap_modes }
+  elseif type(keymap_modes) ~= "table" then
+    keymap_modes = { "n" }
+  end
   opts = opts or {}
+
```
scheatkode, line 21: I tested with this fix but the issue persists. The error originates from `make_detach`.

scheatkode (commented):
Hoping this helps.

lua/nvim-treesitter/textobjects/attach.lua
kiyoon: Did I make another mistake?

kiyoon: I see, sorry maybe I need more sleep haha

scheatkode: That's the right fix, just not the right location :sweat_smile:. `make_detach` is the origin of the error while you applied the fix for `make_attach`.

kiyoon: I thought I changed the detach part but I was rushing. Sorry about that. Now with #367 it should be good right?

PR #375, opened 2023-01-28
Add tests, Make builtin find consistent (when t and repeating)

kiyoon:
Added tests for:
1. repeatable move builtins (by comparing actual movement from vim's builtins)
2. select consistency (`viad`, `cia`, `dia` should be the same)

One test fails because currently there is a glitch.
Using https://github.com/nvim-treesitter/nvim-treesitter/pull/4228 will make all tests pass.

By doing this, I noticed that the builtin find movements we implemented have a very minor difference.  
For some reason, if you use `t` and repeat with `;`, then the default count is 2, so `;` and `2;` should work the same way.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,59 @@
+name: Tests
+
+on:
+  push:
+    branches:
+      - 'master'
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    branches:
+      - 'master'
+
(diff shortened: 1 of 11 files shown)


theHamsta (approved):
Looks nice! Especially since we can now easily add plenary test.

It probably wouldn't have been necessary to write whole runners: a few minimal tests on fixed example would probably also have detected many errors, but the runner approach would be able to easily add failing examples for edge cases.

PR #378, opened 2023-02-04
Add @comment.inner to elixir documentation elements

skbolton:
Added a capture to the inner contents of an elixir documentation tag. These tend to come in two forms: the string form and the options form

String form

```elixir
@moduledoc """
Hey look! Some documentation.
"""
```

Performing a `@comment.inner` change on the documentation would leave you in the following state.

```elixir
@moduledoc """|"""
```

I tried the best I could to have the state be the following but I don't know if its possible easily, how I have it is the next best thing.

```elixir
@moduledoc """
|
"""
```

Options form of documentation elements.

```elixir
@doc deprecated: "Use .... instead"
```

Same `@comment.inner` edit would get you to this state.

```elixir
@doc |
```

### DIFF
--- a/queries/elixir/textobjects.scm
+++ b/queries/elixir/textobjects.scm
@@ -70,12 +70,21 @@
 ; Documentation Objects
 (unary_operator 
   operator: "@"
-  operand: (call target: ((identifier) @_identifier (#any-of? @_identifier
-    "moduledoc" 
-    "typedoc" 
-    "shortdoc" 
-    "doc"
-  )))
+  operand: (
+    call target: ((identifier) @_identifier (#any-of? @_identifier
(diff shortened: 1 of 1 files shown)


kiyoon:
Thanks for the PR!

@theHamsta What do you think? Maybe renaming it to something else like `@docstring.inner` makes more sense? Then we should also support outer and many other languages like Python too. I haven't designed many queries so I can't tell how difficult it would be.

kiyoon:
Since both docstrings are registered as `@comment.outer` already I think it makes sense to name it like `@comment.inner`. Thank you!

skbolton:
@kiyoon I agree!

PR #380, opened 2023-02-04
 fix(elixir): handle last block bug for @class.inner 

skbolton:
With the previous matches if you triggered a `@class.inner` change and you happened to be in the last block of a module it would only change that block instead of deleting all the blocks of the module.

For example with a file such as this where `|` is the cursor:

```elixir
defmodule Test do
  def hello() do
    IO.puts("hi")
  end

  def oi() do
    |IO.puts("oi")
  end
end
```

After the edit you'd have:

```elixir
defmodule Test do
  def hello() do
    IO.puts("hi")
  end

  |
end
```

The previous approach was using a single query and then trying to build
a range over all children of a block, or in the case of a single child
using it as the start and end. I could not for the life of me figure out
a way to write this to work. As soon as I broke it out to two queries:
one with a make range and one that just captures the one child things
worked.

### DIFF
--- a/queries/elixir/textobjects.scm
+++ b/queries/elixir/textobjects.scm
@@ -5,6 +5,7 @@
 ] (#make-range! "block.inner" @_do @_end)) @block.outer
 
 ; Class Objects (Modules, Protocols)
+; multiple children
 (call
   target: ((identifier) @_identifier (#any-of? @_identifier 
     "defmodule" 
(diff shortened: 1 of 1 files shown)


kiyoon:
Thank you. With this improved version however, the same problem exists when you try to delete in the first `hello()` function.

skbolton:
Ah your are right. More work to do on this. I will pull back to draft.

skbolton:
Changing my approach on this. Before my initail PR I had hacked on this for a few hours. What I have here consistenly works but it is a little annoying to have to define things as two separate queries. It feels like there is a way to make it work for both cases in a single query. But I for the life of me couldn't figure out how to make `#make-range` fork for both situations.

kiyoon:
Thank you!

Issue #249, opened 2022-08-16
Elixir: Not all functions are properly identified


halfdan:
**Describe the bug**
In addition to #245, there are two other function declarations that are currently not correctly identified by @function.inner/@function.outer queries. The following are not matched:

```
def foo(), do: 12
def foo() do
  12
end

def guarded_fn(a, b) when a == b, do: a + b
def guarded_fn(a, b) when a != b do
  a * b
end
```

**To Reproduce**
1. Add above code to editor
2. Try to move to next/prev function. 

**Expected behavior**
Functions with guards and functions without parameters are correctly identified.

**Additional context**

skbolton:
Having to deep dive treesitter to learn all that's needed to fix this. Posting my findings here so that maybe someone can speed me up on fixing this. The problem is in [this](https://github.com/skbolton/nvim-treesitter-textobjects/blob/249d90a84df63f3ffff65fcc06a45d58415672de/queries/elixir/textobjects.scm#L48-L65) rule.

The rule assumes the functions argument list will not be empty
```
  (arguments (call [
    (arguments (_) @parameter.inner . "," @_delimiter)
    (arguments ((_) @parameter.inner) @_delimiter .) 
  ] (#make-range! "parameter.outer" @parameter.inner @_delimiter))
```

What I am struggling with is getting the `#make-range` to work properly when adding an empty argument clause. If I drop the whole make range logic and just do something like this the function selection works.

```scm
(call 
  target: ((identifier) @_identifier (#any-of? @_identifier 
    "def" 
    "defmacro" 
    "defmacrop" 
    "defn" 
    "defnp" 
    "defp"
  ))
  (arguments (call [
    (arguments (_) @parameter.inner . "," @_delimiter)
    (arguments ((_) @parameter.inner) @_delimiter .) 
    (arguments)
  ]))
   (keywords
     (pair
       value: (_) @function.inner))
  )
) @function.outer
```
I don't know if the proper approach is to define several somewhat duplicated rules that do or do not use the `make-range` helper or if there is a way to only call that predicate on a part of the match logic.

skbolton:
Going to keep playing with this. I think the solution is to stop trying to define a lot of the text objects in one query and instead make separate queries for ewct text objects. By pulling the parameters query away from the functions query I think I can get everything lined up

PR #383, opened 2023-02-08
fix(elixir): separate function from parameter matching

skbolton:
Previous queries were mixing the parameter matches and the function
matches into the same query. This would make it so that functions that
didn't have parameters wouldn't match @function.inner or @function.outer
queries.

Separating them makes it easier to focus on the specifics of fulfilling
each query. This should resolve issue nvim-treesitter#249.

### DIFF
--- a/queries/elixir/textobjects.scm
+++ b/queries/elixir/textobjects.scm
@@ -28,12 +28,7 @@
   (do_block "do" . (_) @class.inner . "end")
 ) @class.outer
 
-; Function, Parameter, and Call Objects
-(anonymous_function
-  (stab_clause 
-    right: (body) @function.inner)
-) @function.outer
-
+; Parameters
 (call 
   target: ((identifier) @_identifier (#any-of? @_identifier 
     "def" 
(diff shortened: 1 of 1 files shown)


skbolton:
Sorry @kiyoon for the barage of PR's I was trying to break up my work into smaller pieces. I wasn't sure all who is involved in this project and how many eyes would see it.

kiyoon:
When I test it, it seems to be not working with functions with parameters in #249 examples?

kiyoon:
Thanks a lot for your work, and no worries, it makes it easier to focus on one problem at a time.

queries/elixir/textobjects.scm
```diff
@@ -63,10 +89,8 @@
     "defnp" 
     "defp"
   ))
-  (arguments (call [
-    (arguments (_) @parameter.inner . "," @_delimiter)
-    (arguments ((_) @parameter.inner) @_delimiter .) 
-  ] (#make-range! "parameter.outer" @parameter.inner @_delimiter))
+  (arguments
+    (call)
```
yihuikhuu, line 93:
```suggestion
    ([
      (call)
      (identifier)
      (binary_operator)
    ])
```

yihuikhuu (commented):
Changing the above appear to address the remaining issues mentioned in #249.

The `binary_operator` case captures the guard scenario.

The `identifier` case addresses another similar issue when the braces are omitted for parameterless functions:

```
def foo, do: 12
def foo do
  12
end
```

skbolton:
Thanks for the reminder on this PR @yihuikhuu. I'll take another stab at this incorporating your feedback this week.

calops:
I should have checked existing PRs before opening #605 😅

Mine seems to be a lot simpler, but maybe it's too generic? I've tested it a bit and it seems to match all call parameters in every situation (and it helps simplify other queries as well) without breaking other stuff, but maybe I'm missing something.

skbolton:
I stepped away from this for a while. I want to come back to this and make some improvements. Closing for now since this work has gone stale.

PR #385, opened 2023-02-09
update available textobjects in readme

kiyoon:


kiyoon:
I tried to manually update readme but it seems like the CI has a bug that it doesn't install some languages and it will force-push?

PR #398, opened 2023-02-18
Update TypeScript textobjects for type definition support

benyn:
`interface`, `type` and `enum` declarations are now supported as class `textobject`s.

Known issues:
- the last semicolon (`;`) of the inner text (`interface`, `type`) or the last comma of the inner text (`enum`) are missed.
- an object [declared with `as const`](https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums) isn't supported (`value: as_expression` ending with `type_identifier`)

_This is related to https://github.com/nvim-treesitter/nvim-treesitter-textobjects/discussions/396_

### DIFF
--- a/queries/typescript/textobjects.scm
+++ b/queries/typescript/textobjects.scm
@@ -1,32 +1,25 @@
 ; inherits: ecma
 
-
 ;; interface declaration as class textobject
-(interface_declaration) @class.outer
+(interface_declaration
+  body: (object_type)) @class.outer
 
 (interface_declaration
-  body: _ @class.inner) @class.outer
-
-(interface_declaration
-  body: (object_type . "{" . _ @_start @_end _? @_end . "}"
+  body: (object_type . "{" . (_) @_start @_end _? @_end . "}"
  (#make-range! "class.inner" @_start @_end)))
 
 ;; type alias declaration as class textobject
-(type_alias_declaration) @class.outer
+(type_alias_declaration
+  value: (object_type)) @class.outer
 
 (type_alias_declaration
-  value: _ @class.inner) @class.outer
-
-(type_alias_declaration
-  value: (object_type . "{" . _ @_start @_end _? @_end . "}"
+  value: (object_type . "{" . (_) @_start @_end _? @_end . "}"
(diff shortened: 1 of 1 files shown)


queries/typescript/textobjects.scm
```diff
@@ -1 +1,32 @@
...
+
+
+;; interface declaration as class textobject
+(interface_declaration) @class.outer
+
+(interface_declaration
+  body: _ @class.inner) @class.outer
+
+(interface_declaration
+  body: (object_type . "{" . _ @_start @_end _? @_end . "}"
```
theHamsta, line 11: it works for me when you remove the first `@_end`

queries/typescript/textobjects.scm
```diff
@@ -1,32 +1,25 @@
 ; inherits: ecma
 
-
 ;; interface declaration as class textobject
-(interface_declaration) @class.outer
-
 (interface_declaration
-  body: _ @class.inner) @class.outer
+  body: (object_type)) @class.outer
```
kiyoon, line 5: Can't you just remove the body, if you're not going to use `@class.inner`?

```lua
(interface_declaration) @class.outer
```

should work. Same for `type_alias_declaration` and `enum_declaration`.

kiyoon:
Also, instead of adding commits can you rebase with master, leaving only 1 clean commit and force-push?
You can use `git rebase -i master`

Thanks for the PR btw!

queries/typescript/textobjects.scm
kiyoon: Has it been resolved?

benyn:
> Also, instead of adding commits can you rebase with master, leaving only 1 clean commit and force-push?
I'm on my phone now but will do that when I'm on my laptop.

kiyoon:
@benyn No need to worry about rebasing. If you fix it to a working version with conversations resolved, I'll squash and merge this.

theHamsta:
@benyn friendly ping. I think your change is really useful. Would be great to merge this soon!

benyn:
Sorry it's been a busy week. Will try to get this in this weekend!

PR #399, opened 2023-02-25
feat: add return textobject for C/CPP/Lua/Python/Rust/Vim

dmyTRUEk:
But sadly, Rust's return textobject works only for explicit return statements, and doesn't work for most functions because [most functions return the last expression implicitly](https://doc.rust-lang.org/stable/book/ch03-03-how-functions-work.html#functions-with-return-values) (without explicit `return` keyword).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,6 +27,8 @@ Each capture group can be declared as `inner` or `outer`.
 @comment.outer
 @assignment.inner
 @assignment.outer
+@return.inner
+@return.outer
 
 # For LaTeX frames
 @frame.inner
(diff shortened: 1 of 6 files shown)


dmyTRUEk:
So in CONTRIBUTING.md said:

> You don't have to update the support matrix in the README manually. It's generated by a CI job that runs on pushes to master.

But what about new textobjects? Should i add them manually to README.md or no?

kiyoon:
> But what about new textobjects? Should i add them manually to README.md or no?

No, just update `CONTRIBUTING.md` and the readme will be updated automatically.

theHamsta (approved):

PR #401, opened 2023-02-28
feat: add thrift

amaanq:
This is my first time writing a textobjects.scm file - I would appreciate critique.

For the weird queries at the top, I don't want to match assignments that could, say, have optional attributes above them,  only the qualifiers, identifier, and value. I hope I did it right but maybe it's better to use `#make-range!` here.

### DIFF
--- a/queries/thrift/textobjects.scm
+++ b/queries/thrift/textobjects.scm
@@ -1,0 +1,130 @@
+; Assignments
+
+(namespace_uri
+  (uri_def) @assignment.lhs
+  "="
+  (uri) @assignment.inner @assignment.rhs) @assignment.outer
+
+(const
+  (
+    "const"
+    (_) @assignment.lhs
+    "="
+    (const_value) @assignment.inner @assignment.rhs
+  ) @assignment.outer)
+
+(enum
+  (
+    (enum_member) @assignment.lhs
+    "="
+    (number) @assignment.inner @assignment.rhs
+  ) @assignment.outer)
+
+(field
+  (
+    (_) @assignment.lhs
+    "="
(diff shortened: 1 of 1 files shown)


queries/thrift/textobjects.scm
```diff
@@ -0,0 +1,130 @@
...
+(struct
+    "{" (_) @class.inner "}")
+
+(union) @class.outer
+(union
+    "{" (_) @class.inner "}")
+
+(exception) @class.outer
+(exception
+    "{" (_) @class.inner "}")
```
theHamsta, line 95: this is guaranteed to be only one node?

amaanq: It's a repeat of "field"

amaanq: image

kiyoon:
Thank you. I didn't check everything yet, but assignment.inner should select either LHS or RHS of the assignment. So depending on your cursor location it should select left or right.

In my experience, if you put assignment.inner on both side in one query, then it will ignore one of them. So I recommend you to copy the query and add an inner for the left side.

kiyoon:
Also, sorry for the delay. There are languages I understand better and others that I don't know very well. If you don't mind, I'd appreciate if you can provide with a sample code that can test all of the textobjects you've defined. If not, don't worry, I'll also look into finding them. It's just for ease of testing, and plus documenting what kind of cases are tested.

queries/thrift/textobjects.scm
theHamsta: so it could be multiple nodes and this query wouldn't work then

PR #402, opened 2023-03-02
adds assignment support for cpp

amar-laksh:


kiyoon:
The LHS (left hand side) is selecting `int` and RHS `a = 12222`.
image

But it should be LHS selecting `a` and RHS `12222`

Same for inner and outer. Inner should select either LHS or RHS.

I'm not sure whether outer should select the whole statement or just `a = 12222` because I think other text object can select a statement already.

queries/cpp/textobjects.scm
```diff
@@ -55,3 +55,10 @@
 
 (new_expression
   (argument_list) @call.inner) @call.outer
+
+(declaration
+ type: (_) @assignment.lhs
```
theHamsta, line 60: Doesn't this query also work for C? C++ could inherit from it?

theHamsta: @amar-laksh friendly ping

amar-laksh: Hiya, sorry about this. Been a bit busy. It should work for assignments in C as well since the AST should be same for that part. I am going to have a look and update this by weekend?

PR #404, opened 2023-03-06
v: Add V textobjects

yochem:
The following textobjects are added:

- [x] @assignment.inner (`mut x = |[]string{}|`)
- [x] @assignment.lhs (`|mut x| = []string{}`)
- [x] @assignment.outer (`|mut x = []string{}|`)
- [x] @assignment.rhs (`mut x = |[]string{}|`)
- [ ] @attribute.inner
- [ ] @attribute.outer
- [x] @block.inner
- [x] @block.outer
- [x] @call.inner
- [x] @call.outer
- [x] @class.inner
- [x] @class.outer
- [x] @comment.inner (`// |example|`)
- [x] @comment.outer (`|// example|`)
- [x] @conditional.inner (`if true { |pass| }`)
- [x] @conditional.outer (`|if true { pass }|`)
- [ ] @frame.inner
- [ ] @frame.outer
- [x] @number.inner
- [x] @function.inner
- [x] @function.outer
- [x] @loop.inner (`for num in numbers { |pass| }`)
- [x] @loop.outer (`|for num in numbers { pass }|`)
- [x] @parameter.inner
- [x] @parameter.outer
- [ ] @regex.inner
- [ ] @regex.outer
- [x] @return.inner
- [x] @return.outer
- [ ] @scopename.inner
- [x] @statement.outer

I've based most queries on Go's from current master.

### DIFF
--- a/queries/v/textobjects.scm
+++ b/queries/v/textobjects.scm
@@ -1,0 +1,61 @@
+;; assignment
+[(short_var_declaration
+    left: (_) @assignment.lhs
+    right: (_)* @assignment.rhs @assignment.inner)
+  (assignment_statement
+    left: (_) @assignment.lhs
+    right: (_)* @assignment.rhs @assignment.inner)] @assignment.outer
+
+;; block
+(_ (block) @block.inner) @block.outer
+
+;; call
+(call_expression) @call.outer
+(call_expression
+  arguments: (argument_list . "(" . (_) @_start (_)? @_end . ")"
+  (#make-range! "call.inner" @_start @_end)))
+
+;; class: structs
+(struct_declaration
+  (struct_field_declaration_list . "{" . (_) @_start @_end (_)? @_end . "}"
+  (#make-range! "class.inner" @_start @_end)))
+
+(struct_declaration) @class.outer
+
+;; comment
+((comment) @comment.inner @comment.outer
+  (#offset! @comment.inner 0 3 0))
+
+;; conditional
+(if_expression
+  consequence: (block . "{" . (_) @_start @_end (_)? @_end . "}"
+    (#make-range! "conditional.inner" @_start @_end))?) @conditional.outer
+
+;; function
+(function_declaration
+  body: (block . "{" . (_) @_start @_end (_)? @_end . "}"
+ (#make-range! "function.inner" @_start @_end)))
+
+(function_declaration) @function.outer
+
+;; loop
+(for_statement
+  body: (block . "{" . (_) @_start @_end (_)? @_end . "}"
+    (#make-range! "loop.inner" @_start @_end))?) @loop.outer
+
+;; parameter
+(parameter_list
+  "," @_start .
+  (parameter_declaration) @parameter.inner
+ (#make-range! "parameter.outer" @_start @parameter.inner))
+
+(parameter_list
+  . (parameter_declaration) @parameter.inner
+  . ","? " "? @_end
+ (#make-range! "parameter.outer" @parameter.inner @_end))
+
+;; return
+(return_statement (_)* @return.inner) @return.outer
+
+;; statements
+(block (_) @statement.outer)
--- a/queries/v/textobjects.scm
+++ b/queries/v/textobjects.scm
@@ -1,10 +1,20 @@
 ;; assignment
 [(short_var_declaration
     left: (_) @assignment.lhs
-    right: (_)* @assignment.rhs @assignment.inner)
+    right: (_)* @assignment.rhs)
   (assignment_statement
     left: (_) @assignment.lhs
-    right: (_)* @assignment.rhs @assignment.inner)] @assignment.outer
+    right: (_)* @assignment.rhs)]
+
+[(short_var_declaration
+    left: (_) @assignment.inner)
+  (assignment_statement
+    left: (_) @assignment.inner)]
+
+[(short_var_declaration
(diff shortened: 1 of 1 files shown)


kiyoon:
Thank you. I didn't check everything yet, but `assignment.inner` should select either LHS or RHS of the assignment. So depending on your cursor location it should select left or right.

In my experience, if you put `assignment.inner` on both side in one query, then it will ignore one of them. So I recommend you to copy the query and add an inner for the left side.

yochem:
Thanks, I think this is the approach you meant. It works at least :)

yochem:
BTW: in the assignment queries I have to do the `left: ` and `right: ` part twice, for `short_var_declaration` and `assignment_statement`. Is it possible to shorten this? I tried to put the `[]` before the `left:` and `right:`, but that didn't work...

yochem:
Also, would it be an idea to have a short explanation for every predefined textobject in the readme? for example, I didn't know assignment.inner would have to work like that. Let me know if that would be a good idea, I can make a PR for it.

kiyoon:
> Also, would it be an idea to have a short explanation for every predefined textobject in the readme? for example, I didn't know assignment.inner would have to work like that. Let me know if that would be a good idea, I can make a PR for it.

I think it's a good idea. Maybe it is enough to be put in `CONTRIBUTING.md`? Currently the readme is automatically generated, so you shouldn't touch the final section.

yochem:
hhm, I think it's also nice for users to know how they could use it. Maybe then in the docs and refer to that in `CONTRIBUTING.md`?

kiyoon:
Maybe we can put a link to `CONTRIBUTING.md` in the `README.md`

queries/v/textobjects.scm
```diff
@@ -0,0 +1,71 @@
...
+;; class: structs
+(struct_declaration
+  (struct_field_declaration_list . "{" . (_) @_start @_end (_)? @_end . "}"
+  (#make-range! "class.inner" @_start @_end)))
+
+(struct_declaration) @class.outer
+
+;; comment
+((comment) @comment.inner @comment.outer
+  (#offset! @comment.inner 0 3 0))
```
kiyoon, line 37: Is `comment.inner` guaranteed to have offset of 3? Maybe it's fine for now, but to be more accurate maybe we need to offset 2 and delect the leading spaces.

yochem: I've just created a PR for `comment.inner` for python to do exactly this. Will push the fix for `V` now.

yochem: I've just created a PR for `comment.inner` for python to do exactly this. Will push the fix for `V` now.

kiyoon:
> BTW: in the assignment queries I have to do the `left: ` and `right: ` part twice, for `short_var_declaration` and `assignment_statement`. Is it possible to shorten this? I tried to put the `[]` before the `left:` and `right:`, but that didn't work...

I don't think there's a way since every keyword has different structure. I might be wrong

kiyoon:
I think sometimes people use more than one comment string.  
For example, `///////// comment`.

Should we detect this and also select only the comment part? Same goes for python and other languages.
If people want to select inner, I think it's most likely that they want to copy / delete the content only.

yochem:
Since this is easy to implement and probably how people wish it would work, I'll implement this yes :)

yochem:
Although, I'm thinking about it now, and I don't know if it is possible to either dynamically set the offset, or split the capture group based on regex. Then I thought about using the same solution as if statements, where inner keeps the `{}`:

```scm
(if_expression
  consequence: (block . "{" . (_) @_start @_end (_)? @_end . "}"
    (#make-range! "conditional.inner" @_start @_end))?) @conditional.outer
```
I tried to convert it for comments:
```scm
(comment . "/" . (_) @_start @_end (_)? @_end
  (#make-range! "comment.inner" @_start @_end)) @comment.outer
```
But can't seem to get it to work. Do you have any ideas?

kiyoon:
I think that only works when there's more nodes inside the brackets. Comment node doesn't really have further nodes inside.  
Maybe it's trickier than it looks like..

Depending on the language it's common to have three (e.g. lua uses docstring with three dashes).  
So maybe in the worst case we can support up to three, until we find the solution.

kiyoon:
Did you bring most of the queries from another language? They look very familiar (maybe the syntax of V is similar to C?).  
In that case most of the queries should be stable. I'm afraid I find it difficult to test all textobjects from languages that I don't know.

yochem:
Yes, most of them came from Go, as V itself is based on Go. I think the following are from Go's textobjects.scm:

- block
- call
- class (structs)
- conditional
- function
- loop
- parameter
- statement

So my own:
- assignment
- comment
- number
- return 

and of course some name-changes of the nodes etc.

yochem:
> I think that only works when there's more nodes inside the brackets. Comment node doesn't really have further nodes inside.
Maybe it's trickier than it looks like..

Yes, probably... We could also check if the ts parser for [comments](https://github.com/stsewd/tree-sitter-comment) wants to add a marker node. Then, when the comments TS is injected we can at easily by excluding the comment marker node.

kiyoon:
Sounds like a promising solution. But the comment string is different per language so I don't know how they would define that node.

kiyoon:
Looks good and I'm going to merge it now. Thanks for the work!

PR #405, opened 2023-03-06
python: add `@comment.inner`

yochem:
```python
# test
--- cic ---
# |
```

and
```python
#test
--- cic ---
#|
```

where `|` denotes the cursor.

I added the first as 'special case', because I think that is (and should be) how most programmers use comments: *always* having a space after the comment marker. Adding this as a special case skips having to add the space every time after something like `cic`.


kiyoon:
Looks really good, thank you!

PR #407, opened 2023-03-06
yaml: add assignment, statement and comment

yochem:


yochem:
Should I also add `@block.outer` to be the same as `@statement.outer`?

kiyoon:
Not sure how block should be defined here. We can leave it empty for now, I think the current query looks good!

PR #408, opened 2023-03-06
fish: add textobjects

yochem:
Fish's treesitter has a weird way of not using unique field values.

```fish
set -q x "hello"
```
Gives:
```
command [0, 0] - [0, 16]
  name: word [0, 0] - [0, 3]
  argument: word [0, 4] - [0, 6]
  argument: word [0, 7] - [0, 8]
  argument: double_quote_string [0, 9] - [0, 16]
```

I can't seem to get *all* arguments, to use e.g. for `@call.inner`. Let me know if this is possible, then I'll add the functionality of those

### DIFF
--- a/queries/fish/textobjects.scm
+++ b/queries/fish/textobjects.scm
@@ -11,13 +11,17 @@
   argument: (_)* @assignment.inner
   (#eq? @_name "set"))
 
-; ;; block
-; ([
-;  (case_clause)
-;  (if_statement)
-;  (switch_statement)
-;  (else_clause)
-; ] (_) @block.inner) @block.outer
+;; block
+([
+ (case_clause)
+ (if_statement)
+ (switch_statement)
+ (else_clause)
+ (for_statement)
+ (while_statement)
+]) @block.outer
+
+
 
 ;; call
 ; call.inner doesn't work because it can't select *all* arguments
--- a/queries/fish/textobjects.scm
+++ b/queries/fish/textobjects.scm
@@ -1,5 +1,50 @@
-(function_definition) @function.outer
+;; assignment
+(command
+  name: (word) @_command
+  argument: (word) @_varname @assignment.lhs @assignment.inner
+  argument: (_)* @assignment.rhs
+  (#not-lua-match? @_varname "[-].*")
+  (#eq? @_command "set")) @assignment.outer
 
+(command
+  name: (word) @_name
+  argument: (_)* @assignment.inner
+  (#eq? @_name "set"))
+
+; ;; block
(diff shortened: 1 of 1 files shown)


yochem:
Sorry, see now that it is not supported yet (https://github.com/nvim-treesitter/nvim-treesitter-textobjects/issues/102)

queries/fish/textobjects.scm
```diff
@@ -1,16 +1,66 @@
...
+  argument: (_)* @assignment.inner
+  (#eq? @_name "set"))
+
+; ;; block
+; ([
+;  (case_clause)
+;  (if_statement)
+;  (switch_statement)
+;  (else_clause)
+; ] (_) @block.inner) @block.outer
```
kiyoon, line 20: Any reason this is commented out?

yochem: Sorry, was because the `@block.inner` does not work yet. Forgot to remove it

kiyoon:
@theHamsta Is it possible to group multiple nodes without using `+`/`*`?

kiyoon:
@theHamsta I see. I couldn't really find the documentation for `make-range!` so I find it difficult to understand how to use it

theHamsta:
> @theHamsta I see. I couldn't really find the documentation for `make-range!` so I find it difficult to understand how to use it

See https://github.com/nvim-treesitter/nvim-treesitter/blob/fc1ca10bfbdee17e29374d0d1bac8ea030539dc3/lua/nvim-treesitter/tsrange.lua#L1-L154

`make-range!` doesn't work with regular queries by Neovim https://github.com/nvim-treesitter/nvim-treesitter/blob/fc1ca10bfbdee17e29374d0d1bac8ea030539dc3/lua/nvim-treesitter/query_predicates.lua#L123, but it does with our legacy query abstraction https://github.com/nvim-treesitter/nvim-treesitter/blob/fc1ca10bfbdee17e29374d0d1bac8ea030539dc3/lua/nvim-treesitter/query.lua#L278-L284. It is not documented because it is kind of a hack. Are you in the matrix channel for nvim-treesitter-core and tree-sitter-dev for query coordination? We need to find a solution adapted by upstream Neovim (and possibly Helix or tree-sitter in general) to select multiple nodes as this is not possible at the moment. `make-range!` constructs an object that looks and behaves like a regular tree-sitter node, so that regular tree-sitter APIs in Neovim work

kiyoon:
Okay, it makes sense to me now. I'm not in the matrix channel and I'm interested in signing up. Is it matrix.org account that I need?

PR #413, opened 2023-03-09
fix: more select errors

Mike325:


kiyoon:
Can you update neovim to the latest nightly? This has been resolved.

PR #423, opened 2023-03-24
fix(deprecation): use query.get_files to avoid deprecation

theHamsta:
See https://github.com/neovim/neovim/pull/22761

### DIFF
--- a/lua/nvim-treesitter-textobjects.lua
+++ b/lua/nvim-treesitter-textobjects.lua
@@ -4,7 +4,12 @@ local utils = require "nvim-treesitter.utils"
 local M = {}
 
 M.has_textobjects = function(lang)
-  return vim.treesitter.query.get_query_files(lang, "textobjects") ~= nil
+  if vim.treesitter.query.get_files then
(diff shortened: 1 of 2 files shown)


theHamsta:
The only other usage in the check query script. I think we can wait for 0.9 as users are not annoyed by that.

kiyoon:
Will it not break the older nightly 0.9 versions before the patch?

theHamsta:
We can wait one day or just do the shorter `a or b`. With the version, the message is more clear.

theHamsta:
If the deprecation warning would be less annoying, the right way would be to do an explicit version check in a few weeks. :smile:

kiyoon:
True, I think normally it's not that annoying but using nvim-notify makes it very annoying

theHamsta:
Is it ok now?

kiyoon:
@theHamsta looks alright from the code but I haven't yet actually tested it

kiyoon:
It works well on the stable and nightly neovim before the deprecation (my current installation). I think it will work with the master branch as well.

kiyoon (approved):

PR #433, opened 2023-04-06
feat(rust,python): `@parameter.*` in use/import statements

TwIStOy:


theHamsta (approved):

theHamsta:
Thank you!

kiyoon:
I don't know much about rust, but is it treating importing modules as parameters? It doesn't work on Python and it might be good to have it there as well

TwIStOy:
> I don't know much about rust, but is it treating importing modules as parameters?

Yes.
This tries to match parameters like this:
```rust
use std::process::{Command, Stdio, ExitStatus};
//                 ^-----^
```

> It doesn't work on Python and it might be good to have it there as well. 

I will add a similar query for python import statement later.

theHamsta:
I was also bit unsure whether they are parameters. They are usually located at the top of the file so thought the risk of confusion with real parameters is small. But I think you can also import within a function. I assume you want to use them as parameters for swapping and parameter movement

TwIStOy:
Or maybe a new textobject like "@import.inner"?

kiyoon:
Are use statements in rust also placed within a function, or do they appear on the top always?

So, the friction you can have is that users expect to look forward to the next function's parameter, but it can randomly choose an import statement. Maybe it won't be too bad if you're used to it.

Adding too many textobjects can be also a bit annoying. Another option is to provide a way to map "select one of multiple textobjects". However, users usually don't customise their setup this much and expect the plugin to have good defaults. I tried this with move (you can assign multiple queries) but I didn't hear anyone using it actually

TwIStOy:
> Are use statements in rust also placed within a function, or do they appear on the top always?

A use-statement is a kind of declaration statement. It can be placed anywhere.

> So, the friction you can have is that users expect to look forward to the next function's parameter, but it can randomly choose an import statement. Maybe it won't be too bad if you're used to it.

Currently, the textobject "@parameter.inner" is not only the function's parameters but also struct fields, enum fields. So it's hard to jump to just the next function's parameters.

> Adding too many textobjects can be also a bit annoying. Another option is to provide a way to map "select one of multiple textobjects". However, users usually don't customise their setup this much and expect the plugin to have good defaults. I tried this with move (you can assign multiple queries) but I didn't hear anyone using it actually

Map multiple textobjects maybe a better option. Because I think this plugin can not be used without any customization. Users should carefully select the queries in `keymaps`, `swap` or `move`. So a new kind of textobject and a way to map multiple textobjects will be better.

kiyoon:
Thanks for your opinion, @TwIStOy!

One problem with allowing multiple captures in one select command is that it's unclear which to select if there are multiple options to be chosen from. It might create lots of cases with unintuitive selections, if not done right.

But if everyone else agrees that it may be good to allow multiple captures, I can implement that. Otherwise, I think it's intuitive enough to consider import statements to be included as parameters.

kiyoon:
By the way, I just tested the queires.

In python, it doesn't work on all kinds of import statements. I think it only works in `import something` but not in `from something import something`

The below statement doesn't work.
image

Make sure also in rust you cover all kinds of syntax.

TwIStOy:
Thanks for review! I have fixed it!

kiyoon:
Thanks for the update!

Sorry for being picky, but do you think `from ...` statement is a parameter? What I meant was to include the `import ...` part in the `from .. import ...` statements.

```python
from module import foo1, foo2, foo3
#                  ^^^^^^^^^^^^^^^^
```
Only the `^` marked bits should be considered as a parameter.  
My reasoning is:

1. The module is always one word. `@parameter.*` usually select one of multiple options. You can select the whole module in `from ...` with `viW` anyway.
2. It will be hard to lookahead to the import statements, because the `from` parts are always selected.

If you have any different opinion, please correct me!

TwIStOy:
Make sense! I'm hesitating whether to treat it as a parameter or not when i was implementing it.
I will remove it from parameters.

PR #435, opened 2023-04-09
fix: detect_selection_mode()

xvzc:
# Issue
When specifying selection modes for query strings like below the selection mode that the user has specified is overwritten by `detect_selection_modes()` `in select.lua`
```
      selection_modes = {
        ["@function.outer"] = 'V', -- linewise
        ["@function.inner"] = 'V', -- linewise
        ["@parameter.outer"] = 'v', -- charwise
        ["@class.outer"] = '<c-v>', -- blockwise
      },
```
In the previose loginc, when selection_mode is 'V'(uppercase), the final return value will be 'v'(lowercase). so we cannot configure the selection mode.

### DIFF
--- a/lua/nvim-treesitter/textobjects/select.lua
+++ b/lua/nvim-treesitter/textobjects/select.lua
@@ -1,9 +1,9 @@
 local api = vim.api
-local configs = require("nvim-treesitter.configs")
-local parsers = require("nvim-treesitter.parsers")
+local configs = require "nvim-treesitter.configs"
+local parsers = require "nvim-treesitter.parsers"
 
-local shared = require("nvim-treesitter.textobjects.shared")
-local ts_utils = require("nvim-treesitter.ts_utils")
+local shared = require "nvim-treesitter.textobjects.shared"
+local ts_utils = require "nvim-treesitter.ts_utils"
 
 local M = {}
 
@@ -127,7 +127,7 @@ function M.detect_selection_mode(query_string, keymap_mode)
   }
   local method = keymap_to_method[keymap_mode]
 
-  local config = configs.get_module("textobjects.select")
+  local config = configs.get_module "textobjects.select"
   local selection_modes = val_or_return(config.selection_modes, { query_string = query_string, method = method })
   local selection_mode
   if type(selection_modes) == "table" then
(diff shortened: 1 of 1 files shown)


kiyoon:
Thanks for the PR!

Unfortunately there are some tests failing. One of them says `viad` and `dia` produces different results, for example. Can you use the `scripts/minimal_init.lua` to investigate why this is happening?

Also, I don't understand the logic very well (both before and after the change). Why is it bypassing the selection mode when it's operator-pending modes (or not)? If anyone knows some reason behind the logic it would be helpful. It must have been changed over time to tackle some issues.

I personally don't use the selection modes, but does anyone else have the same issue, not being able to configure it? I wonder why it hasn't been made into an issue yet.

xvzc:
My current configuration is like below.
```lua
selection_modes = {
    ["@function.outer"] = 'V', -- linewise
    ["@function.inner"] = 'V', -- linewise
},
```
With this configuration, I think the `@function.outer` qeury should be executed within `visual-line` mode, but it selects the area with `visual` mode
Here's what i found after a little bit of debugging with this config.

```lua
-- selection_mode == 'V' (uppercase)
local visual_mode = vim.fn.mode(1)
-- visual_mode == 'v' (lowercase)
visual_mode = visual_mode:sub(#visual_mode)
-- visual_mode == 'v' (lowercase)
selection_mode = visual_mode == "o" and selection_mode or visual_mode
-- selection_mode == 'v' (lowercase)
```
I don't know very well abount vim.fn.mode() functions, but in my case, the vim.fn.mode(1) function returns "v".
so i think we need some more conditions for this situation.

xvzc:
Can i ask for another execution of test for [799b86b](https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/435/commits/799b86bf704e665eb801eac02a595f3bd0644439)?

lua/nvim-treesitter/textobjects/select.lua
```diff
@@ -137,9 +139,16 @@ function M.detect_selection_mode(query_string, keymap_mode)
...
+  -- the value returned from vim.fn.mode()
+  -- ends with "v" e.g. {'nov', 'noV', 'v', 'V'}
   local visual_mode = vim.fn.mode(1)
   visual_mode = visual_mode:sub(#visual_mode)
-  selection_mode = visual_mode == "o" and selection_mode or visual_mode
+
+  local valid_chars = { "o", "v", "V" }
+  if not vim.tbl_contains(valid_chars, visual_mode) then
+    selection_mode = visual_mode
+  end
```
kiyoon, line 151: I still don't get this logic. If the current mode is not visual, visual line or operator-pending, the selection mode becomes the current mode. Wouldn't selection mode mostly be normal mode in this case?

PR #445, opened 2023-04-29
fix: Use ts compat module for get_query

lucario387:
Closes nvim-treesitter/nvim-treesitter#4731

### DIFF
--- a/lua/nvim-treesitter-textobjects.lua
+++ b/lua/nvim-treesitter-textobjects.lua
@@ -1,15 +1,11 @@
 local configs = require "nvim-treesitter.configs"
 local utils = require "nvim-treesitter.utils"
+local ts = require "nvim-treesitter.compat"
 
 local M = {}
 
 M.has_textobjects = function(lang)
-  if vim.treesitter.query.get_files then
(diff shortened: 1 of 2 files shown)


kiyoon:
Thanks! In order to close the issue, don't you have to also fix the function in nvim-treesitter directly? The deprecation warning may happen in other plugins that use this

lucario387:
> Thanks! In order to close the issue, don't you have to also fix the function in nvim-treesitter directly? The deprecation warning may happen in other plugins that use this

It's already in used for other plugins, including nvim-treesitter itself as far as I'm aware. So far no deprecation notice has been popping up while using it (to me at least)

kiyoon:
https://github.com/nvim-treesitter/nvim-treesitter/blob/1ca9b938c1c3d8b2c64a168f6a5419b8f3870410/lua/nvim-treesitter/query.lua#L108

This is where the problem came in. So you can also submit a PR there to replace it to compat

kiyoon:
Sorry, it's actually using the compat module. I don't know why the issue has been raised then. Maybe the author didn't update nvim-treesitter.

PR #447, opened 2023-05-04
Add R assignment textobjects

fbearoff:
This adds the assignment text objects for the R language. `@assignment.outer` encompasses the whole statement so the lhs can be selected from the rhs using similar logic to #386


kiyoon:
Thank you!

First of all, sorry for not having clear instructions on contributing.

The `@assignment.inner` should select either LHS or RHS. I think right now it only selects one of them. It should select based on where your cursor is.

In my experience, putting `@assignment.inner` twice in one query doesn't work. So you should make a separate query.

Secondly, you don't have to update README manually. It will be done automatically. I'm saying this for the future but you can leave that as is for now.

fbearoff:
I think I figured it out now, thanks for the guidance.

kiyoon:
Thank you!

PR #451, opened 2023-05-20
fix(dot-repeat): Use '<cmd>...<cr>' style for mapping to enable proper dot repeat

phgz:
At the moment, when using a direct lua function for the key mappings, the "dot-repeat" action does not trigger as expected. For example, `dap` which will  delete a 3 chars parameter, when repeated with `.`, will simply delete 3 chars again, instead of deleting the parameter the cursor is on.

**edit**: Locally, when I try `dVac` (`ac` is "@conditional.outer") then `.` with the old method, I get stuck in an apparent infinite loop. This behaviour is fixed with the new method. I don't know if it is reproducible or if it's due to my config.


kiyoon:
Thanks!

PR #454, opened 2023-05-22
feat: update readme filter languages

kiyoon:
Only show languages with at least one textobject defined


kiyoon:
I tested on my forked repo and it works

PR #458, opened 2023-05-28
queries: add support for norg (with query copied from neorg's own folds.scm

laher:
Hi. Please consider this small addition for the `.norg` filetype. `.norg` is the filetype created for the [Neorg](https://github.com/nvim-neorg/neorg) project, and Neorg is a `neovim` plugin for note taking and task/project management. It's similar to Emacs' `org-mode`, but the syntax is quite different. Neorg uses treesitter to parse and manipulate `.norg` files.

Being a structured markup language, `.norg`'s tree features nested `heading[1-6]`s. These headings are already referenced in `fold` queries, and it would be straightforward & valuable to be able to reuse these queries for text objects.

It's a very simple use case, but it's very handy to be able to yank headings (and their content) at any level like this. It's working well for me.

For reference:
* The treesitter parser for `.norg` is here: https://github.com/nvim-neorg/tree-sitter-norg
* The `.norg` specification is here: https://github.com/nvim-neorg/norg-specs
* The provided query came directly from this file in the `neorg` project: https://github.com/nvim-neorg/neorg/blob/main/queries/norg/folds.scm

Thanks

### DIFF
--- a/queries/norg/textobjects.scm
+++ b/queries/norg/textobjects.scm
@@ -1,0 +1,13 @@
+(ranged_verbatim_tag
+    name: (tag_name) @_name
+    (#eq? @_name "document.meta")
+) @fold
+
+[
+    (heading1)
+    (heading2)
+    (heading3)
+    (heading4)
+    (heading5)
+    (heading6)
+] @fold


kiyoon:
Hi, I think this is not really for the textobjects. You can open a PR at the `nvim-treesitter`

kiyoon:
The textobjects doesn't support `@folds` query in other languages. If you want that, you can just add the file in your `~/.config/nvim/queries/norg/textobjects.scm` file.

laher:
OK, I'll do that. I didn't know about this `nvim/queries` directory, thanks. I'll close this now.

FWIW I think `@fold` seems like a pretty nice `textobject` (I'm using `["z"] = "@fold"` in my `select` keymaps), because you don't have to remember different mappings for different language constructs, and it seems to handle nesting very nicely. I only tried it for `norg` so far, but I'll try it out for a couple of actual programming languages.

Thanks for your time.

laher:
> `["z"] = {query = "@fold", query_group = "folds"}`
> 
> and it will work for all languages.

Aah. That's great, now I get it.
So I don't need to explicitly load this template either, it's already loaded by neorg.

I have passed on the info to the neorg folks.
Thanks for the explanation.

PR #459, opened 2023-06-01
fix(go): update definition of `@parameter`

theHamsta:
Fixes https://github.com/nvim-treesitter/nvim-treesitter-textobjects/issues/460

Definition `parameter_declaration`: https://github.com/tree-sitter/tree-sitter-go/blob/7cccc30535a40da07092626031120ae6b36202d7/grammar.js#L236

### DIFF
--- a/queries/go/textobjects.scm
+++ b/queries/go/textobjects.scm
@@ -76,12 +76,12 @@
  (#make-range! "parameter.outer" @parameter.inner @_end))
 
 (parameter_declaration
-  (identifier)
-  (identifier) @parameter.inner)
+  name: (identifier)
+  type: (_) @parameter.inner)
 
 (parameter_declaration
-  (identifier) @parameter.inner
-  (identifier))
+  name: (identifier) @parameter.inner
+  type: (_))
 
 (parameter_list
   "," @_start .


theHamsta:
Hmm, this does not handle the case multiple identifiers, same type

queries/go/textobjects.scm
```diff
@@ -76,12 +76,12 @@
  (#make-range! "parameter.outer" @parameter.inner @_end))
 
 (parameter_declaration
-  (identifier)
-  (identifier) @parameter.inner)
```
theHamsta, line 80: this is no longer an `(indentifier)` but can be any `_type`

kiyoon:
> Hmm, this does not handle the case multiple identifiers, same type: `a, b int` will count as one `@parameter.outer`. I suppose that is ok?
> 
> Or should it be `a,` and `b int`?

Has it been like this all the time (before this PR)? I don't know about Go but I think it should be separated. But if people didn't complain about it probably there are reasons why they were okay with it.

theHamsta:
> > Hmm, this does not handle the case multiple identifiers, same type: `a, b int` will count as one `@parameter.outer`. I suppose that is ok?
> > Or should it be `a,` and `b int`?
> 
> Has it been like this all the time (before this PR)? I don't know about Go but I think it should be separated. But if people didn't complain about it probably there are reasons why they were okay with it.

It shouldn't change behavior. I think `a, b int` is somewhat rare and you probably don't need parameter.outer that often for that

theHamsta:
Can we merge as a hotfix?

PR #463, opened 2023-06-07
add glimmer support

yads:
Support for glimmer parser


kiyoon:
I'm sorry for the delay. I'll approve this for now, but since this is a very special case and I don't know if defining the mustache block to be a `@function.*`. If other users may ask for a change and if alternative sounds more intuitive, the behaviour can change any time.

yads:
That's fair. Thanks!

PR #470, opened 2023-06-26
feat: add on_not_selected callback

woodgear:


kiyoon:
Thank you. What do you think is the use case for this one?

woodgear:
to distinguish between an internal error of Neovim/Treesitter and a mismatch between the current buffer and the query string .
for example
```lua
  require("xxx").setup {
    | (where the cursor)
  }
```
i thought the setup is "function.inner",but in fact is not.
it take me sometimes to figure out, the treesitter-textobject works prepectly. it just has not to match the query string

clason:
Thank you, but that is not a compelling argument for adding a callback. (Builtin textobjects also don't give any feedback if no valid object can be found.)

PR #474, opened 2023-07-23
[Scala] Add queries for case and enumeration

ghostbuster91:


theHamsta (approved):
This looks good to me (not familiar yet with the new captures).

Let's wait for https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/475 to fix CI.

theHamsta:
@ghostbuster91 could you rebase on master to fix CI?

kiyoon:
Is it a good idea to add a very language specific queries like this, rather than merging it to something existing (like parameters)?

Also, when you introduce new queries you need to add that to `CONTRIBUTING.md`

theHamsta (changes requested):
@kiyoon is right `@case` probably is `@conditional`. Also `@enumerator` does not exist. I was assuming they were added when I was away for vacation.

ghostbuster91:
I see, I was not aware of that. I am not sure if `@case` should fall into `@conditional`.  
It is kind of condition but it is also something else than regular if statement. 

In scala case expression can happen in two scenarios:
1. as a part of partial function
   ```scala
    list.collect { case elemen:Type => element }
   ```
2. as a part of match expression
   ```scala
   value match {
     case "some pattern" => do_something()
   }
   ```

Regarding enumerator. I am not sure if such things are in other languages. 
In scala we have for comprehensions e.g.
```scala
for {
 a <- fun1()
 b = fun2(a)
} yield result
```

In this case the for-comp expression contains two enumerators:
`a <- fun1()`
and 
`b = fun2(a)`

Do you know about any existing category that this should fall into?

kiyoon:
@ghostbuster91 If we allow too many language-specific captures, there will be hundreds of captures in the end. Users can configure it as they want, but it requires an understanding of each capture for each language. The goal of this plugin is to provide a straightforward and intuitive configuration that most people will agree on. Thus, I support putting `case` into `@conditional`.

About the enumerator, I don't really understand what it does. Maybe `@loop.inner`?

Also, note that not every capture has to be included in this plugin. If most people don't need this capture, it could be in your nvim config separately and this plugin can register any custom captures.

ghostbuster91:
Yeah, I think that the enumerator might be too specific. I will remove it from the PR and put the `case` into `conditional`.

theHamsta (approved):

PR #495, opened 2023-10-01
feat(go): add assignment textobjects

nikklassen:


queries/go/textobjects.scm
```diff
@@ -97,3 +97,25 @@
...
+(assignment_statement
+  left: (_) @assignment.lhs
+  right: (_) @assignment.rhs @assignment.inner) @assignment.outer
+(var_spec
+  name: (_) @assignment.lhs
+  value: (_) @assignment.rhs @assignment.inner)
+(var_spec
+  name: (_) @assignment.inner
+  type: (_))
+(var_spec) @assignment.outer
```
ObserverOfTime, line 114: Instead of querying it a third time, just add the capture to the other two (same for `(const_spec)`).

nikklassen: Yup, makes sense

ObserverOfTime (approved):

PR #511, opened 2023-10-24
Fix/ js inner without curly braces

ValJed:
Following [some posts](https://github.com/nvim-treesitter/nvim-treesitter-textobjects/issues/112) I found about removing the curly braces of inner selections.
Like for functions and methods it feels logical that it would behave the same for other blocks.

I updated the ecma text objects to select only the inner part for loops and if statements.
I tested locally it seems working nicely, let me know if I made some mistakes, I basically reused code I found. (not familiar with the syntax).

### DIFF
--- a/queries/ecma/textobjects.scm
+++ b/queries/ecma/textobjects.scm
@@ -52,8 +52,8 @@
   (#make-range! "loop.inner" @_start @_end))) @loop.outer
 
 (if_statement
-  consequence: (_)? @conditional.inner
-  alternative: (_)? @conditional.inner) @conditional.outer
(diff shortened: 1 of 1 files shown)


theHamsta (approved):
LGTM to me! Some of the other maintainers want to have a look @kiyoon  @amaanq ?

kiyoon:
It looks very good!

By the way, did the behaviour in Python change? Maybe the python parser has some updates? The python test is failing maybe because of that

theHamsta:
The Python tests are failing already since a long time. I did not investigate the reasons for that.

PR #512, opened 2023-10-25
feat: add astro textobjects

virchau13:
Astro's textobjects are the same as HTML's textobjects, so it's just inherited from HTML.

Fixes https://github.com/virchau13/tree-sitter-astro/issues/19.

### DIFF
--- a/queries/astro/textobjects.scm
+++ b/queries/astro/textobjects.scm
@@ -1,0 +1,1 @@
+; inherits: html


queries/astro/textobjects.scm
```diff
@@ -0,0 +1 @@
+; inherits: html
```
devgioele, line 1:
```suggestion
; inherits: html,tsx
```

Wdyt about inheriting also the textobjects of tsx?
An astro file does not contain html only.
Do you think the two are in conflict?

devgioele (changes requested):

queries/astro/textobjects.scm
virchau13: The Astro tree-sitter parser parses only modified HTML; TypeScript and TSX are injected, so the textobjects for TS/TSX should already be handled.

theHamsta (approved):
Thank you!

Issue #503, opened 2023-10-08
Refactor into standalone plugin


clason:
[Nvim-treesitter 1.0](https://github.com/nvim-treesitter/nvim-treesitter/issues/4767) (expected later this year) will completely remove the module system in order to significantly reduce complexity.

This means that nvim-treesitter-textobjects needs to be refactored into a standalone plugin, similarly to https://github.com/nvim-treesitter/nvim-treesitter-context or the [`indent.lua` module in 1.0](https://github.com/nvim-treesitter/nvim-treesitter/blob/main/lua/nvim-treesitter/indent.lua). (There's no need to hook into `nvim-treesitter`; textobjects should simply work if a suitable parser and queries are in runtimepath. This should also allow simplifying the `setup` as it no longer needs to be shoehorned into the nvim-treesitter setup structure.)

As a bonus, the `incremental-selection` (also removed in 1.0) feature could be integrated here as a `in`/`an` node object.

TheLeoP:
Would a PR be welcome?

clason:
Of course!

TheLeoP:
Would the PR also have to remove the dependency on nvim-treesitter?

clason:
Yes, that is the whole point!

ilan-schemoul:
> [Nvim-treesitter 1.0](https://github.com/nvim-treesitter/nvim-treesitter/issues/4767) (expected later this year) will completely remove the module system in order to significantly reduce complexity

For reference : nvim treesitter roadmap to 1.0 now reads we'll wait nvim 0.11 before Treesitter 1.0. nvim 0.10 just got released after a very long time and now we need to wait 0.11 then (hopefully) Treesitter will be 0.10 then (hopefully) this package will be 1.0

All this could take a year or two so I find it not a great thing my PR get halt because we have to wait 1.0 https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/588#issuecomment-2047918166

clason:
I consider this as completed by #523 and the follow-up fixes (thanks all!). Feel free to test it (if **and only if** you are testing nvim-treesitter `main`!) and report (or, better yet, fix) any regressions you find.

PR #523, opened 2023-11-18
refactor: remove nvim-treesitter as a dependency

TheLeoP:
## DONE:

- [x] refactor
    - [x] copy and refactor functions previously owned by nvim-treesitter and no longer present on their `main` branch
    - [x] move away from the nvim-treesitter modules based architechture
    - [x] expose lua functions instead of creating keymaps ourselves
    - [x] introduce a TSTextObjects.Range abstraction to deal with: TSNode, range4, range6, lsp_range, vim_range and fake nodes created using the metadata from multiple captures (i.e. `@*.outer` captures)
- [x] update README
    - [x] show new keymaps examples using `vim.keymap.set` directly
- [x] update docs
    - [x] same as README
- [x] update tests
- [x] update scripts
- [x] add autoformat queries to CI (has also already been done on the `master` branch)

This addresses #503

## Moved to a follow up PR (TODO)
- [ ] cache query results
- [ ] reimplement consistency tests using Lua ([python was previously used for this](https://github.com/kiyoon/nvim-treesitter-textobjects-tests))
- [ ] better define the user interface of the plugin
  - [ ] some mappings rhs have the form `some_exposed_func` while others have the form `function() some_exposed_func() end`
  - [ ] should other interfaces like the previously used user commands be considered?
- [ ] maybe include incremental selection as `an`, `in` textobject


clason:
Thank you for tackling this! I think the best course here is to not merge this to master but make this a new parallel branch `main` (which will become the default branch when nvim-treesitter itself switches). The goal here is to make this work (only) in concert with the `main` branch of nvim-treesitter; mixing and matching `main` and `master` is not supported.

(This also means we don't have to complete the rewrite in a single PR.)

> refactor the code again (to not only used the copied functions from nvim-treesitter)

💯 

> check if autocmds on filetype are the best option

Probably not; I would enable this globally with a check for parsers and query that short-circuits them.

Also, it's perfectly fine (and often preferable) to have some global startup code in `/plugin`; the standard assumption is that the plugin should work without explicitly calling `setup` if you're fine with default options.

> add some kind of on_attach like nvim-treesitter-context (?)

I consider `nvim-treesitter-context` to be the gold standard right now for a standalone treesitter plugin; I'd try to follow its lead as much as possible. (@lewis6991)

>  refactor attach.lua

Yes, `attach`/`detach` *delendam esse*. These are obsolete concepts. The assumption is that people enable these different functions themselves, however makes most sense for the functionality.

 should vim.treesitter.query.get be cached?
> delete configuration fields for keymaps and let users define the keymaps themselves (?)

Yes, definitely! We don't need yet another DSL for setting keymaps. (Although we could talk about reasonable default mappings as opt-in.)

Although possibly this is a special case here; maybe we could discuss instead exposing a single (Lua!) function that is easy to map?

> nvim-treesitter is only being used inside of scripts (check-queries, minimal_init and update_readme), is this ok?

That is OK; you need `nvim-treesitter` (`main`!) to set up the test environment.

lua/nvim-treesitter-textobjects/shared.lua
```diff
@@ -1,12 +1,291 @@
...
+  else
+    filetype = vim.split(filetype, "%.")[1]
+    assert(filetype, string.format("No language for filetype %s", filetype))
+    return ts.language.get_lang(filetype) or filetype
+  end
+end
+
+---@param bufnr integer
+---@return string lang
+function M.get_buf_lang(bufnr)
```
theHamsta, line 26: If I remember correctly this might be possible even with pure nvim functions

clason: `vim.treesitter.language.get_lang(vim.bo.ft)`

lewis6991:
> should vim.treesitter.query.get be cached?

Correct me if I'm wrong, but doesn't core already cache this via a weak table?

theHamsta (commented):
Are there any plans to keep the commands? We've added them because some users requested them because they wanted to use them in ext mode or for vinyl remappings. Not a strong requirement since all functionality is available via Lua functions, just being curious on what your idea for this PR was.

theHamsta:
> > should vim.treesitter.query.get be cached?
> 
> Correct me if I'm wrong, but doesn't core already cache this via a weak table?

I believe you're right, they should be cached. Nvim-treesitter-textobjects might be doing additional caching of the query results if they are using still an own slow algorithm to aggregate query results

clason:
> Are there any plans to keep the commands? We've added them because some users requested them because they wanted to use them in ext mode or for vinyl remappings. Not a strong requirement since all functionality is available via Lua functions, just being curious on what your idea for this PR was.

I think we should take this opportunity to revisit the user API here, with all the lessons learned since. (Not necessarily in this PR.)

My preference would be to start with a clean slate and (re-)add stuff when it is requested with a compelling use case. (User commands are trivial to add in a `plugin` file, as we do in nvim-treesitter now.)

What I _don't_ want to see again is a custom module for creating keymaps and user commands through `setup()`.

TheLeoP:
> Thank you for tackling this! I think the best course here is to not merge this to master but make this a new parallel branch `main` (which will become the default branch when nvim-treesitter itself switches).

How should I tackle this? I see that I can change which branch the PR targets, but I can't select the `main` branch since it does not exit yet. @clason

clason:
Now it does :)

lucario387:
So `make-range` was never needed? 😱

lucario387:
Looks like that

clason:
Nice; want to make a PR to remove it from the `main` branch? (After local testing, which -- at least for indents -- you're in the best position to do.)

lucario387:
Sadly, my previous comment was incorrect. 

`vim.treesitter.query.{get,parse}` is cached, but the result of it is used for memoize of locals of indents, so neovim core's caching is for the parsing, while nvim-treesitter's caching is for the handling of such parse

clason:
So we will need the same sort of memoization here for the textobjects query results, right?

lucario387:
> So we will need the same sort of memoization here for the textobjects query results, right?

Would be nice, I don't think there's one right now though

TheLeoP:
> So `make-range` was never needed? 😱

I didn't understand the codebase enought to know what `make-range` does, so I deleted it. I just added it again (because I couldn't make the test pass because `parameter.outer` in Python uses `make-range`), sorry :c.

lucario387:
@TheLeoP basically we create a new node from the two node that is specified for `make-range`,

clason:
> @clason Is there a specific part of the code left that you would like to be refactored? I still haven't give too much atention to test, docs or scripts, I wanted to focus on the plugin functionality first.


I think you are right that we should first make sure everything works with nvim-treesitter `main` and that all tests pass; afterwards it's much easier to make changes one by one (e.g., going through the backported utility functions to see if they can be replaced by Neovim core APIs).

Also, [we have an autoformatter for queries](https://github.com/nvim-treesitter/nvim-treesitter/blob/main/scripts/format-queries.lua) now, which could be added to CI here ;)

ObserverOfTime (commented):
Can you keep the keymap configuration in setup?

> What I _don't_ want to see again is a custom module for creating keymaps and user commands through `setup()`.

Disregard this comment. :^)

clason:
No, this comment stands. I want to see the module framework gone. If you want to set keymaps (through this plugin), use the Neovim API directly.

ObserverOfTime:
Yes, I didn't mean it should remain a module.

clason:
@TheLeoP this looks like it's in a good place -- can you take care of the remaining items?

TheLeoP:
@clason. Will do. Question: are the current vimdocs generated? Or do they have to be manually updated?

clason:
They are manually written. It's OK to have a minimal (but correct) version for the initial PR.

scripts/minimal_init.lua
```diff
@@ -19,193 +9,379 @@ require("nvim-treesitter.configs").setup {
...
+    set_jumps = true,
+  },
+}
 
-  textobjects = {
-    select = {
-      enable = true,
+local select = require "nvim-treesitter-textobjects.select"
+for _, mode in ipairs { "x", "o" } do
+  vim.keymap.set(mode, "am", function()
```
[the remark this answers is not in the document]
clason: For `vim.keymap.set`, yes; but some of these functions have weird multiple dispatch. (I've removed such magic in nvim-treesitter; we should remove it here as well. Explicit is better.)

TheLeoP:
Is it expected to document the lua API being used for keymaps on the vimdocs/README?

README.md
```diff
@@ -232,23 +231,27 @@ Alternatively, you could use a repeatable movement managing plugin such as [nvim
...
 local peek_definition_code = require "nvim-treesitter-textobjects.lsp_interop".peek_definition_code
 -- You can use the capture groups defined in `textobjects.scm`
-vim.keymap.set({"n", "x"}, "<leader>df", peek_definition_code("@function.outer", "textobjects"))
-vim.keymap.set({"n", "x"}, "<leader>dF", peek_definition_code("@class.outer", "textobjects"))
+vim.keymap.set({"n", "x"}, "<leader>df", function()
+  peek_definition_code("@function.outer", "textobjects")
+end)
+vim.keymap.set({"n", "x"}, "<leader>dF", function()
+  peek_definition_code("@class.outer", "textobjects")
+end)
```
clason: Not quite clear what the point of this change is? This is just additional indirection without any benefit (since you `require` the module outside of the mapped function).

clason:
It's a matter of taste whether you consider the help file or the README as the primary documentation. It certainly doesn't hurt to have it in the README (and shows that we really need to rework the interface here... follow-up PR, though).

README.md
TheLeoP: I can move all the `require`s inside of the called functions. The change was mainly to avoid long lines.

clason: Yes, you _need_ to move the `require` inside to have any benefits. Otherwise you can just dispense with the `function() ... end` wrapper.

README.md
```diff
@@ -89,21 +91,10 @@ Define your own mappings to swap the node under the cursor with the next or prev
...
-        ["<leader>A"] = "@parameter.inner",
-      },
-    },
-  },
-}
-EOF
+-- keymaps
+local swap = require("nvim-treesitter-textobjects.swap")
+vim.keymap.set("n", "<leader>a", swap.swap_next("@parameter.inner"))
+vim.keymap.set("n", "<leader>A", swap.swap_next("@parameter.outer"))
```
wookayin, line 97: It is not clear `swap.swap_next` is a keymap (dot-repeatable) rhs or the command/action itself that can programmatically called. I don't find it's consistent with `move.goto_next_start`, etc. There is even `swap_textobject` (a local function) that is not a keymap.  Would you think having some module name like `mapping` or `keymap` somewhere would be a good idea?

README.md
TheLeoP: Just to be clear and fully understand you. The purpose of the `function() ... end` wrapper is to invoke a lua function as callback of the keymap. I can't directly put `peek_definition_code("@class.outer", "textobjects")` because I would be putting an integer (this is what `peek_definition_code` returns) as a parameter of `vim.keymap.set`.

I'm interpreting your original message and your last message as if you thought that the purpose of the `function() ... end` wrapper was to lazily call the function. Is this correct? Maybe I'm just not understanding you right now. If that's the case, I'm sorry.

Originally, I thought that you didn't liked the indirection of requiring the function outside of the keymap and then calling it inside it, that's why I made the comment about being able to move all the `require`s inside the keymaps if that was preferred.

README.md
TheLeoP: I agree that a consistent interface would be preferred, but I don't have any strong opinions about how to implement it

mehalter:
Hey @TheLeoP this is an amazing PR and is looking great! I pulled it up and started testing it out and I noticed one weirdness so far when I do something like a selection in a filetype that doesn't support a query. Say I do like `vic` that I configured for visually selecting within a class in a filetype that doesn't support the `@class` query like Lua it gives me the notification that the filetype is not supported but it leaves me in visual mode which I don't expect. Could we have it on a failed attempt to also put it back in normal mode or something? I'm not sure if that makes sense.

Another thing is the notification you get is "This filetype is not supported by nvim-treesitter-textobjects" which is a bit misleading. It would be nice to give a more contextual warning like "The `lua` parser does not support the `@class` query" or something like that.

README.md
clason: Well, the function would return an integer either way? (The only difference is _when_ the integer is computed.)

But I trust you that the current code works; I'd still move the require inside so you only pay the cost when you're actually using a mapping; it's just good practice we should reinforce.

mehalter:
They are sort of regressions from the current `master`. The way the current `master` works is it only creates mappings that are supported so you don't run into this case of a binding that isn't supported. The addition of the error message I mentioned is in this pull request.

ObserverOfTime:
> The way the current `master` works is it only creates mappings that are supported so you don't run into this case of a binding that isn't supported.

Proof that we need to keep the keymap config. ;)

clason:
Yeah, yeah. (But not the old system, that needs to die in a fire.)

TheLeoP:
Currently, the approach is to check (by using `require'nvim-treesitter-textobjects.shared'.check_support`) whether a filetype, query file, and capture are supported every time a function intended for mapping (i.e. `require "nvim-treesitter-textobjects.select".select_textobject("@function.outer", "textobjects", mode)`) is called, so the user has to create only one keymap.

Another option could be to instructs users into creating keymaps inside a `filetype` autocmd that uses `check_support` before creating a buffer scoped keymap. This would be similar to how LSP keymaps are usually created inside an `LspAttach` autocmd and would be consisten with the current behavior.

scripts/minimal_init.lua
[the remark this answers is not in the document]
TheLeoP: @clason could you help me with a link to a PR/commit in nvim-treesitter with this kind of changes to get a better idea of how this should work?

clason:
I think your approach sounds good. My point is to have a mergable MVP as quick as possible to unblock the nvim-treesitter transition; we can then add QoL in follow-up PRs. (My second point is to not get stuck in outdated patterns out of convenience.)

scripts/minimal_init.lua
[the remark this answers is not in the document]
clason: Not really, since I just removed everything ;) Here we need the mappings. This was just a general comment, though; if there's no magic and @ObserverOfTime's suggestion works, just do that.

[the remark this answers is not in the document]
clason: Or, conversely, if the function needs to be passed the `mode` because it can't work otherwise (or find out itself), then you need to keep the loop.

TheLeoP:
All the TODOs (beside the caching, which I'm not sure how to handle because of things like `get_capture_matches_recursively`), have been adressed. Is there something else obvious that I may be missing?

queries/supercollider/textobjects.scm
```diff
@@ -1,51 +1,60 @@
...
+      (_))) @parameter.inner)
+  (#make-range! "parameter.outer" @_start @parameter.inner))
+
+(((symbol)?
+  (identifier)
+  .
+  ((method_name)
+    (function_call
+      (_))) @parameter.inner.","? @_end)
+  (#make-range! "parameter.outer" @parameter.inner @_end))
```
lucario387, line 60: It trips again ahhh

Can you manually reformat this specific query? It looks like a certain amount of parens can be kicked

clason:
Thank you very much! I think this looks good for the first (most important) step; the rest can be done in follow-up PRs as the need arises (memoization of queries and further refactors to simplify code paths; convenience functions for setting up keymaps).

What remains here is just to fix the failing tests. I am not too familiar with the tests here, but the first steps should be to check the `.github/workflows` to make sure
1. they use the `nightly` tag for Neovim;
2. they install nvim-treesitter from the `main` branch instead of the default `master`;
3. use the provided scripts for installing parsers (see `./scripts/install-parsers.lua` there).

mehalter:
Out of curiosity @clason which part of this PR depends on Neovim nightly and `nvim-treesitter` `main` branch? I have been testing the PR on both Neovim 0.9 and Neovim nightly with `nvim-treesitter` master branch and the only line that is not compatible with Neovim 0.9 is the use of `vim.list_contains` in `shared.lua` on line 710. This can also be replaced with `vim.tbl_contains` and it would have full support for neovim non-nightly and nightly as well as treesitter master and main.

Would it be useful for that single function call change to happen to maintain that sort of compatibility?

lua/nvim-treesitter-textobjects/shared.lua
```diff
@@ -0,0 +1,754 @@
...
+  end
+
+  if queries then
+    if vim.tbl_isempty(queries) then
+      return false
+    end
+
+    local available_textobjects = M.available_textobjects(buf_lang, query_group)
+    for _, query in ipairs(queries) do
+      if not vim.list_contains(available_textobjects, query:sub(2)) then
```
mehalter, line 710:
```suggestion
      if not vim.tbl_contains(available_textobjects, query:sub(2)) then
```

I believe this is the only change necessary to maintain support for non-neovim nightly

clason:
Let me be clear that the main branches will not consider compatibility with Nvim 0.9. 

So textobjects main will require nvim-treesitter main, which very much requires the latest Nvim master (and will continue to do so aggressively).

mehalter:
Yeah, I am just curious what part of this new refactor requires `nvim-treesitter` main. I am currently using it with `nvim-treesitter` master and it seems to work totally fine, no? I also could be confused and it just be sheer luck that it works?

clason:
My comment was about the tests, which rely on nvim-treesitter. 

Obviously, the plugin itself doesn't, which is the whole point of the exercise. But it's important to set expectations; if anything changes on Neovim master, the main branch will adapt instantly (including removing things that are upstreamed or become unnecessary) without any warning.

mehalter:
I see, I was wondering based on the comment regarding updating the `README` since that seems to recommend to users that if they want to use this plugin they should use neovim nightly. This discussion may be off topic and should happen elsewhere so feel free to mark it as such if necessary.

If it is the case that all development on `main` will adapt to neovim `main` in an "as soon as possible" manner. What is the plan to actually support users? Will there be a much stronger adherence to version tagging/semantic versioning and maintaining which parts of the repository can actually be used by users who aren't using the least stable version of Neovim? It seems like if there is no plan to support anything less than the latest commit on neovim it will very quickly burn users who are using stable releases

clason:
Please refer to the nvim-treesitter tracking issue, which details the rationale.

mehalter:
Yeah I definitely meant in relation to this plugin rather than `nvim-treesitter`. I didn't realize this also had a `main` vs `master` thing going on as well. Was mainly curious about getting this refactor out as a release to users who aren't using Neovim nightly would be nice. I haven't really been talking about `nvim-treesitter` at all since this PR fully separated this plugin and that so they have no relation

mehalter:
Appreciate your time and effort supporting this huge project ❤️

queries/supercollider/textobjects.scm
TheLeoP: When running the new formatting script, this ends up as

```scheme
(((symbol)?
  (identifier)
  .
  ((method_name)
    (function_call
      (_))) @parameter.inner.
  ","? @_end)
```

Which gives an error for the Neovim query diagnostics on nightly because `@parameter.inner.` is interpreted as the name of the capture (instead of the last `.` being interpreted as an anchor). Is this intended?

.github/workflows/check_query_files.yml
```diff
@@ -19,7 +19,7 @@ jobs:
 
       - name: Prepare
         env:
-          NVIM_TAG: stable
+          NVIM_TAG: ${{ matrix.nvim_tag }}
```
clason, line 22:
```suggestion
          NVIM_TAG: nightly
```

clason, line 22:
```suggestion
          NVIM_TAG: nightly
```

clason:
You can use `nvim -l` to run Lua scripts. To install all parsers, use (an adapted copy of) https://github.com/nvim-treesitter/nvim-treesitter/blob/main/scripts/install-parsers.lua

queries/supercollider/textobjects.scm
lucario387: It's not intended for sure.

doc/nvim-treesitter-textobjects.txt
```diff
@@ -107,76 +101,98 @@ or previous function.
...
+  end)
+
+  -- Go to either the start or the end, whichever is closer.
+  -- Use if you want more granular movements
+  -- Make it even more gradual by adding multiple queries and regex.
+  vim.keymap.set({ "n", "x", "o" }, "]d", function()
+    require("nvim-treesitter-textobjects.move").goto_end("@conditional.outer", "textobjects")
+  end)
+  vim.keymap.set({ "n", "x", "o" }, "[d", function()
+    require("nvim-treesitter-textobjects.move").goto_end("@conditional.outer", "textobjects")
```
kiyoon, line 167: aren't the names should be `goto_next` and `goto_previous`?

README.md
```diff
@@ -113,56 +101,74 @@ This is similar to `]m`, `[m`, `]M`, `[M` Neovim's mappings to jump to the next
...
+end)
+
+-- Go to either the start or the end, whichever is closer.
+-- Use if you want more granular movements
+-- Make it even more gradual by adding multiple queries and regex.
+vim.keymap.set({ "n", "x", "o" }, "]d", function()
+  require("nvim-treesitter-textobjects.move").goto_end("@conditional.outer", "textobjects")
+end)
+vim.keymap.set({ "n", "x", "o" }, "[d", function()
+  require("nvim-treesitter-textobjects.move").goto_end("@conditional.outer", "textobjects")
```
kiyoon, line 164: Again I think this is a typo of `goto_next` and `goto_previous`

queries/supercollider/textobjects.scm
lucario387: Ok it's my fault :)

Please run again.

lucario387: I think I fixed it now (for real). Please drop the formatting commits again and re-format the whole thing :bowing_man:

TheLeoP:
The query files test is working again c:, but the python tests seem to be broken indeed and I don't know a lot of python to look into them.

I wouldn't like to keep blocking this PR. @clason would it be ok to remove the python tests, as you mentioned, and move native lua test to a follow up PR?

README.md
TheLeoP: `make_dot_repeatable` returns a function, so the wrapping isn't needed

doc/nvim-treesitter-textobjects.txt
```diff
@@ -185,35 +201,22 @@ Overrriding or extending text objects~
...
+    end)
+  end
 <
 Where '@custom-capture' can be defined in a scheme file
->
-  ```scm
-  -- after/queries/python/textobjects.scm
-  (function_definition) @custom-capture
-  ```
+>scheme
```
lucario387, line 216: tree sitter query is shipped with nightly so this can be 
```suggestion
>query
```

clason:
Can you squash your commits and update the PR/commit message (what's changed, what's still todo)? Then I'll merge.

TheLeoP:
I've squashed all my commits and updated the PR and commit messages. Let me know if there is something else I may be missing

clason:
You may need to rebase on `main` as well (`rebase -i main` and drop all but your commit), otherwise  the merge won't be clean.

clason:
Guess we need to rebase on master. Let me try to do this properly, you don't have to worry about it.

clason:
Ok, CI is green.

@TheLeoP Before you delete (or pull!) your branch, please check the diff here against your local changes to make sure I did not force-push some of them away (in which case, please make a follow-up PR from latest `main`). Thank you so much for tackling this and for your patience through the (too) long process!

I'm looking forward to follow-up PRs (by you and other interested parties) cleaning up the old legacy design and adding a proper keymap setup (@ObserverOfTime).

@kiyoon @ribru17 Note that this means that `master` is essentially frozen:
* no further big changes to the plugin (bugfixes are fine, but only if they can be rebased on cleanly -- otherwise they should go to `main` directly)
* query changes still go to `master` as usual, but remember to regularly `git rebase master` on `main`; `git checkout --theirs` for non-query files will deal with conflicts.

ribru17:
Awesome, thank you so much for your work @TheLeoP and @clason! Is there a roadmap (or general idea) for what must happen before `master` is updated to reflect these changes?

clason:
Yes: hell must freeze over.

clason:
Less facetiously: these changes will _never_ make it to `master`; what will happen is that the `main` branch will become the default branch that receives all future updates and `master` becomes essentially a `legacy` branch for people unable to move to the shiny new world. This will happen together with the same change for `nvim-treesitter` (and other plugins we control), so is described by [its roadmap](https://github.com/nvim-treesitter/nvim-treesitter/issues/4767).

ribru17:
Makes sense, thank you

PR #528, opened 2023-12-02
exclude loop.inner braces for rust and cpp

alex-tdrn:


alex-tdrn:
It seems this PR breaks some test cases, and I must admit I cannot figure out how the test suite works. I found https://github.com/kiyoon/nvim-treesitter-textobjects-tests/tree/master but I cannot figure out how to get from the CI output to a specific test file to see what I broke 🤔

kiyoon:
Don't worry about the failing test. It's been broken from before and it's not your fault.

PR #534, opened 2023-12-13
feat(ecma): `@call` object for constructor calls

ribru17:
Provides appropriate `@call.inner` and `@call.outer` text objects for object constructors, e.g. `new MyObject(5, 'hi')`


kiyoon:
Can you rebase it to the master so that the tests pass? It is required to pass before merging and the older master commit was failing. Sorry and thank you!

PR #535, opened 2023-12-13
feat(rust): Add `macro_invocation` as a type of `@call.outer`.

kylechui:
To my understanding, macro invocations should be considered a type of function call. Would also resolve [this issue](https://github.com/kylechui/nvim-surround/issues/290#issuecomment-1854629989).

### DIFF
--- a/queries/rust/textobjects.scm
+++ b/queries/rust/textobjects.scm
@@ -71,6 +71,8 @@
 
 ;; calls
 (macro_invocation) @call.outer
+(macro_invocation (token_tree . "(" . (_) @_start (_)? @_end . ")"
+  (#make-range! "call.inner" @_start @_end)))
 (call_expression) @call.outer
 (call_expression
   arguments: (arguments . "(" . (_) @_start (_)? @_end . ")"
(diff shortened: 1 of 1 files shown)


queries/rust/textobjects.scm
```diff
@@ -70,6 +70,7 @@
 (unsafe_block (_)? @block.inner) @block.outer
 
 ;; calls
+(macro_invocation) @call.outer
```
ribru17, line 73: Would it be possible to also add the corresponding `@call.inner` group? Maybe something like:

```scheme
(macro_invocation (token_tree . "(" . (_) @_start (_)? @_end . ")"
  (#make-range! "call.inner" @_start @_end)))
```

kylechui: Some basic testing using the query editor seems to validate this code; added in a commit.

ribru17 (approved):

kiyoon:
Can you rebase it to the master so that the tests pass? It is required to pass before merging and the older master commit was failing. Sorry and thank you!

kylechui:
Woops accidentally merged instead, hopefully that's fine. Sorry!

PR #539, opened 2023-12-22
fix(select.lua): Fix condition for 'visual_mode' selection

### LINKED ISSUE #515: selection_modes not working
This is my config:
``` lua
textobjects = {
	select = {
		enable = true,

		-- Automatically jump forward to textobj, similar to targets.vim
		lookahead = true,

		keymaps = {
			-- You can use the capture groups defined in textobjects.scm
			["af"] = "@function.outer",
			["if"] = "@function.inner",
			["ac"] = "@class.outer",
			["ic"] = "@class.inner",
			["aa"] = "@parameter.outer",
			["ia"] = "@parameter.inner",
			["ai"] = "@conditional.outer",
			["ii"] = "@conditional.inner",
			["al"] = "@loop.outer",
			["il"] = "@loop.inner",
			["ab"] = "@block.outer",
			["ib"] = "@block.inner",
			["ar"] = "@return.outer",
			["ir"] = "@return.inner",
			-- -- You can optionally set descriptions to the mappings (used in the desc parameter of
			-- -- nvim_buf_set_keymap) which plugins like which-key display
			-- ["ic"] = { query = "@class.inner", desc = "Select inner part of a class region" },
			-- -- You can also use captures from other query groups like `locals.scm`
			-- ["as"] = { query = "@scope", query_group = "locals", desc = "Select language scope" },
		},
		-- You can choose the select mode (default is charwise 'v')
		--
		-- Can also be a function which gets passed a table with the keys
		-- * query_string: eg '@function.inner'
		-- * method: eg 'v' or 'o'
		-- and should return the mode ('v', 'V', or '<c-v>') or a table
		-- mapping query_strings to modes.
		selection_modes = {
			["@parameter.outer"] = "v", -- charwise
			["@function.outer"] = "V", -- linewise
			["@class.outer"] = "<c-v>", -- blockwise
			["@conditional.outer"] = "V", -- linewise
			["@loop.outer"] = "V", -- linewise
		},
		-- If you set this to `true` (default is `false`) then any textobject is
		-- extended to include preceding or succeeding whitespace. Succeeding
		-- whitespace has priority in order to act similarly to eg the built-in
		-- `ap`.
		--
		-- Can also be a function which gets passed a table with the keys
		-- * query_string: eg '@function.inner'
		-- * selection_mode: eg 'v'
		-- and should return true of false
		include_surrounding_whitespace = false,
	},
	move = {
		enable = true,
		set_jumps = true, -- whether to set jumps in the jumplist
		goto_next_start = {
			["]f"] = "@function.outer",
			["]i"] = "@conditional.outer",
		},
		goto_next_end = {
			["]F"] = "@function.outer",
			["]I"] = "@conditional.outer",
		},
		goto_previous_start = {
			["[f"] = "@function.outer",
			["[i"] = "@conditional.outer",
		},
		goto_previous_end = {
			["[F"] = "@function.outer",
			["[I"] = "@conditional.outer",
		},
		-- Below will go to either the start or the end, whichever is closer.
		-- Use if you want more granular movements
		-- Make it even more gradual by adding multiple queries and regex.
		-- goto_next = {
		-- 	["]i"] = "@conditional.outer",
		-- 	["]f"] = "@function.outer",
		-- },
		-- goto_previous = {
		-- 	["[i"] = "@conditional.outer",
		-- 	["[f"] = "@function.outer",
		-- },
	},
},

```
And when i press `vaf` its not linewise:

Screenshot_20231101_231444

xvzc:
The value of `config.selection_modes` is only considered when we are in `operation pending mode`. The example below describes the issue that this PR can solve.

1. Set `config.selection_modes["@function.outer"] = "V"`
2. Type `"vaf"` in normal mode.
3. We get pure `"v"` mode enabled instead of `"V"`

In my opinion, we should also consider the value of `config.selection_modes` in the case of `{"v", "V", "<C-v>"}`. Please kindly review the description, and I would really appreciate it if I could get any feedback.

Fixes #515

### DIFF
--- a/lua/nvim-treesitter/textobjects/select.lua
+++ b/lua/nvim-treesitter/textobjects/select.lua
@@ -139,7 +139,7 @@ function M.detect_selection_mode(query_string, keymap_mode)
   -- last set `selection_mode`
   local visual_mode = vim.fn.mode(1)
   visual_mode = visual_mode:sub(#visual_mode)
-  selection_mode = visual_mode == "o" and selection_mode or visual_mode
+  selection_mode = visual_mode == "o" and visual_mode or selection_mode
 
   if selection_mode == "n" then
     selection_mode = "v"
--- a/lua/nvim-treesitter/textobjects/select.lua
+++ b/lua/nvim-treesitter/textobjects/select.lua
@@ -139,11 +139,8 @@ function M.detect_selection_mode(query_string, keymap_mode)
   -- last set `selection_mode`
   local visual_mode = vim.fn.mode(1)
   visual_mode = visual_mode:sub(#visual_mode)
-  selection_mode = visual_mode == "o" and visual_mode or selection_mode
-
-  if selection_mode == "n" then
-    selection_mode = "v"
-  end
+  vim.notify(visual_mode)
+  selection_mode = visual_mode == "n" and "v" or selection_mode
 
   return selection_mode
 end
--- a/lua/nvim-treesitter/textobjects/select.lua
+++ b/lua/nvim-treesitter/textobjects/select.lua
@@ -134,12 +134,9 @@ function M.detect_selection_mode(query_string, keymap_mode)
     selection_mode = selection_modes or "v"
   end
 
-  -- According to "mode()" mapping, if we are in operator pending mode or visual mode,
-  -- then last char is {v,V,<C-v>}, exept for "no", which is "o", in which case we honor
-  -- last set `selection_mode`
-  local visual_mode = vim.fn.mode(1)
-  visual_mode = visual_mode:sub(#visual_mode)
-  selection_mode = visual_mode == "n" and "v" or selection_mode
+  if selection_mode == "n" then
+    selection_mode = "v"
+  end
 
   return selection_mode
 end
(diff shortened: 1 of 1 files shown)


kiyoon:
Is there any reason why you'd want to enforce `"v"` mode and not use `selection_mode` in normal mode?

To me, it seems more logical to use the `selection_mode` which is `"v"` by default, and if it ever happens to be `"n"` we can set it to `"v"` by default, so you shouldn't remove that statement.  
I don't know when that would occur but I think it is there for a reason.

xvzc:
Ah, okay i agree with you too. Let me add it back.

kiyoon:
I'd like to merge this, and it does fix the problem. However, the required check loading of syntax files test must pass before I can merge. You need to rebase it to the master branch and force update, then it will pass.

xvzc:
I see `Tests / Run tests (ubuntu-latest, gcc, stable) (pull_request)` is still failing. Is there any problem with my commit?

phgz:
This commit causes a regression. We can no longer force a mode for selection, i.e. `yVaf` to yank a function in linewise fasion only yanks it charwise.

phgz:
Does not respect either `Vaf`...

kiyoon:
@phgz Thanks for the feedback. I will revert it and good to know such test cases for future reference.

@xvzc I think we need to use the selection mode only when the current mode is a `v` mode and a default operator-pending mode (not something like `yV`). If you can address this, I'll review it.

PR #540, opened 2023-12-22
feat(c_sharp): treat expression body in methods (arrow function) as `function`

bartoszluka:


kiyoon:
Hi, I have not much clue about C#. Can you elaborate more and show me examples what it covers?

bartoszluka:
@kiyoon if the function in C# only has a `return` statement, which looks like this:

```csharp
int Foo()
{
    return 42;
}
```

it can be rewritten to look like this

```csharp
int Foo() => 42;
```

in the treesitter the second version uses `arrow_expression_clause` as body of the method. This wasn't matched as a `@function.inner` or `@function.outer` so doing a motion like "daf" skipped this kinds of methods.

kiyoon:
OK! That makes sense. Can you rebase it to the master so that the tests pass? It is required to pass before merging and the older master commit was failing. Sorry and thank you!

PR #544, opened 2023-12-27
Add assignment textobjects to CSS

pockata:
These are the property/value textobjects for CSS.

Selecting property (key): 
image

Selecting value:
image

However there's an edge case which I don't know how to properly handle. It has to do with some properties having multiple values. With the current implementation, only one of the values is selected (the closest one).

`background-image`:
image

`transition`:
image

As a workaround we can make `@assignment.rhs` and `@assignment.inner` (which seem to be synonyms in most language queries) to differ slightly and have `@assignment.rhs` capture all the values and `@assignment.inner` capture one value only.

Having the following keymaps works very intuitively for this setup:
``` lua
['ik'] = '@assignment.lhs', -- select the key
['ak'] = '@assignment.outer', -- select the whole assignment
['iv'] = '@assignment.inner', -- select the value (a single value, if multiple are available)
['av'] = '@assignment.rhs', -- select the value (the whole value, if multiple are available)
```

This query is very close, but I struggle to exclude the `:` and `;` characters from the range via `#offset!`.
```scm
(declaration
  (property_name) @assignment.lhs
  (":") @_start (#offset! @_start 0 1 0 1)
  (_) @assignment.inner
  (";") @_end (#offset! @_end 0 -1 0 -1)
  (#make-range! "assignment.rhs" @_start @_end)
) @assignment.outer
```

### DIFF
--- a/queries/css/textobjects.scm
+++ b/queries/css/textobjects.scm
@@ -3,3 +3,8 @@
   (float_value)
   (color_value)
 ] @number.inner
+
+(declaration
+  (property_name) @assignment.lhs
+  (_) @assignment.inner @assignment.rhs
+) @assignment.outer
--- a/queries/css/textobjects.scm
+++ b/queries/css/textobjects.scm
@@ -6,5 +6,11 @@
 
 (declaration
   (property_name) @assignment.lhs
-  (_) @assignment.inner @assignment.rhs
+  . ":" . (_) @_start (_)? @_end . ";"
+  (#make-range! "assignment.inner" @_start @_end)
+  (#make-range! "assignment.rhs" @_start @_end)
 ) @assignment.outer
+
+(declaration
+  (property_name) @assignment.inner
+)


kiyoon:
Note: `assignment.inner` selects both rhs and lhs depending on the cursor position in all other languages, so it should be consistent with CSS

pockata:
Ah, I see. This is also fine.

The query in the PR is consistent with the other `@assignment` queries so it should be good to go. Let me know if you need anything more from me.

kiyoon:
In order for `@assignment.inner` to work properly, we need one more query like this:

```
(declaration
  (property_name) @assignment.inner
)
```

In my opinion the entire value should be selected, because in other languages like Python which support multiple assignment values, we don't separate it.

pockata:
> In order for `@assignment.inner` to work properly, we need one more query like this:
> 
> ```
> (declaration
>   (property_name) @assignment.inner
> )
> ```
I'm a bit confused about this. All other instances of assignment in the queries I could find had `assignment.inner` set to the value of the assignment and not the name. The only exception being value-less assignments, e.g. `var name`.

https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/e7234c57d7a1dfc98d56ef84b1e38dbb79a875bd/queries/go/textobjects.scm#L101-L119

Maybe I'm missing something?

  &nbsp;
  
> In my opinion the entire value should be selected, because in other languages like Python which support multiple assignment values, we don't separate it.

I agree. I've updated the query to select the whole value and behave exactly as other assignment captures that are currently implemented. 

The query is a bit complicated because the nodes in the treesitter tree don't provide left/right or name/value field names like in Go for example.

Perhaps we can add name/value field names in the CSS grammar upstream and later simplify this query?

kiyoon:
It's because, you can't use the same capture name twice in one query. Thus in all other languages they have a separate query that captures lhs part too.

In the example of Go,

```
 (const_spec 
   name: (_) @assignment.lhs 
   value: (_) @assignment.rhs @assignment.inner) @assignment.outer 
 (const_spec 
   name: (_) @assignment.inner 
   type: (_))
```

This is exactly selecting either lhs or rhs. We don't need repeating `@assignment.outer` though, so I removed the second one here for illustration.

pockata:
Oh, I see. I think I understand now.
I've rebased against the current master and updated the queries to reflect the proposed changes.

kiyoon:
This is perfect, except you don't need to repeat the `@assignment.outer`. It is redundant as both queries will capture the same range.

Just make the second query like this:

```
(declaration
  (property_name) @assignment.inner
)
```

pockata:
@kiyoon I see. I was trying to match the style of the existing queries.
I've updated the PR with your feedback. It should be good to go.

kiyoon:
Great, thank you!

PR #546, opened 2023-12-28
fix(select): patch #539 regression

phgz:
@xvzc if you could review. I tested on my side and it works for #515.

### DIFF
--- a/lua/nvim-treesitter/textobjects/select.lua
+++ b/lua/nvim-treesitter/textobjects/select.lua
@@ -138,7 +138,7 @@ function M.detect_selection_mode(query_string, keymap_mode)
   -- then last char is {v,V,<C-v>}, exept for "no", which is "o", in which case we honor
   -- last set `selection_mode`
   local mode = vim.fn.mode(1)
-  local is_normal_or_charwise_v = #mode == 1 and (mode == "n" or mode == "v")
+  local is_normal_or_charwise_v = mode == "n" or mode == "v"
 
   if is_normal_or_charwise_v then
     return mode == "n" and "v" or mode == "v" and selection_mode
--- a/lua/nvim-treesitter/textobjects/select.lua
+++ b/lua/nvim-treesitter/textobjects/select.lua
@@ -134,18 +134,19 @@ function M.detect_selection_mode(query_string, keymap_mode)
     selection_mode = selection_modes or "v"
   end
 
-  -- According to "mode()" mapping, if we are in operator pending mode or visual mode,
-  -- then last char is {v,V,<C-v>}, exept for "no", which is "o", in which case we honor
-  -- last set `selection_mode`
+  local ret_value = selection_mode
   local mode = vim.fn.mode(1)
   local is_normal_or_charwise_v = mode == "n" or mode == "v"
 
-  if is_normal_or_charwise_v then
-    return mode == "n" and "v" or mode == "v" and selection_mode
+  if not is_normal_or_charwise_v then
+    -- According to "mode()" mapping, if we are in operator pending mode or visual mode,
+    -- then last char is {v,V,<C-v>}, exept for "no", which is "o", in which case we honor
+    -- last set `selection_mode`
+    local visual_mode = mode:sub(#mode)
+    ret_value = visual_mode == "o" and selection_mode or visual_mode
   end
 
-  local visual_mode = mode:sub(#mode)
-  return visual_mode == "o" and selection_mode or visual_mode
+  return ret_value == "n" and "v" or ret_value
 end
 
 M.keymaps_per_buf = {}
--- a/lua/nvim-treesitter/textobjects/select.lua
+++ b/lua/nvim-treesitter/textobjects/select.lua
@@ -134,11 +134,18 @@ function M.detect_selection_mode(query_string, keymap_mode)
     selection_mode = selection_modes or "v"
   end
 
-  if selection_mode == "n" then
-    selection_mode = "v"
+  -- According to "mode()" mapping, if we are in operator pending mode or visual mode,
+  -- then last char is {v,V,<C-v>}, exept for "no", which is "o", in which case we honor
+  -- last set `selection_mode`
+  local mode = vim.fn.mode(1)
+  local is_normal_or_charwise_v = #mode == 1 and (mode == "n" or mode == "v")
+
+  if is_normal_or_charwise_v then
+    return mode == "n" and "v" or mode == "v" and selection_mode
   end
 
-  return selection_mode
+  local visual_mode = mode:sub(#mode)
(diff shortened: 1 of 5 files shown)


xvzc:
Could you share your configuration?

lua/nvim-treesitter/textobjects/select.lua
```diff
@@ -134,11 +134,18 @@ function M.detect_selection_mode(query_string, keymap_mode)
...
-  if selection_mode == "n" then
-    selection_mode = "v"
+  -- According to "mode()" mapping, if we are in operator pending mode or visual mode,
+  -- then last char is {v,V,<C-v>}, exept for "no", which is "o", in which case we honor
+  -- last set `selection_mode`
+  local mode = vim.fn.mode(1)
+  local is_normal_or_charwise_v = mode == "n" or mode == "v"
+
+  if is_normal_or_charwise_v then
+    return mode == "n" and "v" or mode == "v" and selection_mode
```
kiyoon, line 144: I would just write it case by case like:

```lua
if mode == "n" then
  return "v"
end

if mode == "v" then
  return selection_mode
end
```

However, what's the reason we'd use `v` mode in normal mode? I think either way it should be `selection_mode`

xvzc: I think we should first set the selection_mode to visual_mode?

phgz: > However, what's the reason we'd use `v` mode in normal mode? I think either way it should be `selection_mode`

It's you who added that haha (#420).

xvzc: Also, assigning a parameter to `vim.fn.mode()` function seems to be no longer working in `nvim nightly`. It just returns on character regardless of the existence of parameter

phgz:
I use no special config. I tried with
```lua
selection_modes = {
["@function.outer"] = "V", -- linewise
}
```
and
```lua
selection_modes = {}
```

lua/nvim-treesitter/textobjects/select.lua
xvzc: What about we do something like this? 
```lua
  -- ...
  local selection_mode = vim.fn.mode()
  if type(selection_modes) == "table" then
    selection_mode = selection_modes[query_string] or selection_mode
  elseif selection_modes then
    selection_mode = selection_modes
  end

  if selection_mode == "n" then
    selection_mode = "v"
  end

  return selection_mode
```

kiyoon: > > However, what's the reason we'd use `v` mode in normal mode? I think either way it should be `selection_mode`
> 
> It's you who added that haha (#420).

That logic is different because it is stopping from accidentally returning "n" as a selection mode. In your logic, you're looking at the current vim mode, but not `selection_mode` configured in this plugin.

xvzc:
@phgz So you wanna respect the selection_mode only when the mode is "v"?

lua/nvim-treesitter/textobjects/select.lua
phgz: Oh, my bad. Here is almost 1AM. Will check it back tomorrow morning when I have rested 😬.

kiyoon:
@xvzc I think that makes sense, and in the default operator-pending mode. (if we can detect the `yV` case, we should prioritise the user's choice.)

lua/nvim-treesitter/textobjects/select.lua
phgz: Although it seems to work, because if `selection_mode` is `n`, so will be `mode`, no? I tried `:TSTextobjectSelect @function.outer` and it selected correctly. In what other case would it fail?

kiyoon: > Oh, my bad. Here is almost 1AM. Will check it back tomorrow morning when I have rested 😬.

Take good care of yourself! :)

phgz:
@xvzc `selection_mode` is respected either if `v` or not in operator pending with forced-motion.

xvzc:
Lets say we have following configuration.
```lua
selection_modes = {
["@function.outer"] = "V", -- linewise
}
```

"vaf" selects "V-Line" mode, but should we select "V-block" mode when we "<C-v>af? what do you guys think?

phgz:
It is already the case.

kiyoon:
@xvzc yeah, if it is something other than `vaf` it should respect the user's choice. And from my understanding this PR addresses this.

kiyoon:
Let's fix 2 things:

1. `is_normal_or_charwise_v` then `ret_value = selection_mode`
2. before returning, `if selection_mode == "n" then return "v"`

xvzc:
I agree with this. Its way more intuitive.

```lua
if mode == "n" then
  return "v"
end

if mode == "v" then
  return selection_mode
end
```

xvzc:
Anyway, my bad, guys. I apologize.

kiyoon:
We should also add some test cases. It's very easy

1. change `scripts/minimal_init.lua` to add `selection_mode` for one type of textobject.
2. In `tests/select/python_spec.lua`, add a line like this
```lua
run:equal_cmds("new_test_file.py", { row = 1, col = 10, cmds = { "dVil", "Vial", "cVil" } })
```

we need to add a `new_test_file.py` which contains a class, for example.

phgz:
> Let's fix 2 things:
> 
> 1. `is_normal_or_charwise_v` then `ret_value = selection_mode`
> 2. before returning, `if ret_value == "n" then return "v"`

We can also simply `if is_normal_or_charwise_v then return selection_mode end`, since it will take user preference into account as well for the `TSTextobjectSelect` command.

lua/nvim-treesitter/textobjects/select.lua
phgz: > Also, assigning a parameter to vim.fn.mode() function seems to be no longer working in nvim nightly. It just returns on character regardless of the existence of parameter

I just downloaded the lastest nightly and it works. Try something like `y:echo mode(1)`.

kiyoon:
I think in some weird cases the `selection_mode` value could be "n", so we need to add that logic before returning it.

xvzc:
> Also, assigning a parameter to `vim.fn.mode()` function seems to be no longer working in `nvim nightly`. It just returns on character regardless of the existence of parameter

By the way, no worries about this. it was just a type warning. it's happening on `nvim 0.10.0` but the fixed version will soon be released with neovim/neovim#26776

phgz:
There seems to be something flaky in one test, but I think it's just a matter of rerunning the failed job (For an oscure reason, I can't see the button to rerun, maybe only "Collaborator" can).

kiyoon:
Thanks a lot for your work! This is perfect :)

PR #548, opened 2024-01-08
Handle comment nodes within text objects

qstrahl:
While I was editing my config for this very plugin, commenting and uncommenting lines in the config table, I tried to make use of the `@parameter` text objects and noticed some unexpected behaviour around comment nodes. The relevant query, it turns out, naively assumes that table constructors consist of single nodes separated by commas. In reality, there could be any number of nodes between those commas: one `field`, and any number of `comment`.

Unfortunately, due to some limitations of treesitter queries (in particular, the fact that the dot anchor ignores anonymous nodes) I was unable to find a simple modification to the existing queries that would account for all cases. Instead, two new queries had to be added, in addition to some slight modifications to the existing queries. I would really like to discover that I've missed an obvious, more succinct solution.

I suspect there are many other queries across all supported languages that have similar room for improvement, but I don't have the time or energy to fix them all, hence the draft PR. It's my intention to continue improving queries as I encounter more edge cases, and my hope that others can apply a similar approach across other queries in the project.


qstrahl:
In the interest of not letting perfect be the enemy of good, and in light of the recent movement on this issue in #549, I'm opening this for review.

PR #550, opened 2024-01-10
fix(swap): preserve swapped nodes' range metadata

ribru17:
Right now the `adjacent` node to be swapped does not have its range metadata preserved properly. This results in weird asymmetrical behavior when swapping nodes that have their own capture ranges, such as the following:
```scheme
((list_item) @swappable
  (#trim! @swappable))
```
Currently, such swaps will include a blank new line at the end, but only when swapping from the second to last list item to the last list item (the reverse direction is unaffected; this also only applies when a new line follows the list). This PR fixes that by preserving the range.

### DIFF
--- a/lua/nvim-treesitter/textobjects/shared.lua
+++ b/lua/nvim-treesitter/textobjects/shared.lua
@@ -332,7 +332,7 @@ function M.next_textobject(node, query_string, query_group, same_parent, overlap
 
   local next_node = queries.find_best_match(bufnr, query_string, query_group, filter_function, scoring_function)
 
-  return next_node and next_node.node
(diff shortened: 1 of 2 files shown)


ribru17:
In case I explained this too confusingly, note that the following weird behavior is observed (and fixed with this PR):

For markdown, with:

```scm
((list_item) @swappable
  (#trim! @swappable))
```

```md
### begin

- swap from me
- to me

### end
```

results in the following:

```md
### begin

- to me

- swap from me
### end
```

whereas swapping from bottom to top gives the following:

```md
### begin

- to me
- swap from me

### end
```

kiyoon:
Thank you!

PR #551, opened 2024-01-12
fix(latex): proper function call text objects

ribru17:
Gives `@call.(outer|inner)` text objects to LaTeX functions, to closer match the highlights. Extends the objects to text functions, and their inner content.

### DIFF
--- a/queries/latex/textobjects.scm
+++ b/queries/latex/textobjects.scm
@@ -16,7 +16,28 @@
 
 [
   (generic_command)
-] @statement.outer
+  (text_mode)
+] @call.outer
+
+(text_mode
+  (curly_group
+    "{"
+    .
+    (_) @_start
+    (_)? @_end
+    .
+    "}")
+  (#make-range! "call.inner" @_start @_end))
+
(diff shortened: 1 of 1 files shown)


kiyoon:
Hi, could you explain with some examples of what issues it used to have and how your PR addresses those?

Thank you!

ribru17:
Sure! The first part is mainly preference, changing the `generic_command` from `@statement.outer` to `@call.outer`. It also adds the appropriate inner groups, so something like this can now be selected/deleted/etc.:
```latex
\frac{5}{2}
%     ^--^---- @call.inner captures
% everything still captured as @call.outer
```
It also gives the same to any text commands (which are parsed as a different named node from `generic_command`:
```latex
\text{I can now manipulate this inner text easily with @call.inner!}
\text{As well as this entire object (including the \text) with @call.outer}
```
Currently there is no way to work with the inner text of a generic command, nor the inner or outer objects of a `\text` command.

kiyoon:
Sounds good to me! Thanks for your contribution :)

PR #558, opened 2024-02-01
chore(ecma): update queries from upstream

amaanq:


amaanq:
nvm, approval not needed here

PR #564, opened 2024-02-09
Support `php_only`

przepompownia:


ribru17:
What is this necessary for? Sorry, I don't know much about the PHP Tree-sitter ecosystem, but it looks like in the `nvim-treesitter` repo, `php_only` is just used as a baseline for the main `php` files to inherit from (and use with the html parser). Does this give any functionality that doesn't exist yet?

przepompownia:
Without it this extension simply does not work if I have `php_only` alias without `php`.

przepompownia:
No, its the matter of a parser alias, not of the filetype.

Merged equivalents:
https://github.com/nvim-treesitter/nvim-treesitter-context/pull/391
https://github.com/Wansmer/treesj/pull/148

ribru17:
I see. Thank you for the explanation!

przepompownia:
Thank you too!

PR #570, opened 2024-02-22
queries/fennel: update queries for the new grammar

fsouza:
Updating queries to account for the new parser introduced in nvim-treesitter/nvim-treesitter#6132.

### DIFF
--- a/queries/fennel/textobjects.scm
+++ b/queries/fennel/textobjects.scm
@@ -1,52 +1,34 @@
-
 ; https://fennel-lang.org/reference
 
 (comment) @comment.outer
 
 (_ . "(" ")" .) @statement.outer
 
-; functions
-((fn . name: (_)? . (parameters) . docstring: (_)? . (_) @_start . (_)* . (_)? @_end .)
- (#make-range! "function.inner" @_start @_end)) @function.outer
-
-((lambda . name: (_)? . (parameters) . docstring: (_)? . (_) @_start . (_)* . (_)? @_end .)
- (#make-range! "function.inner" @_start @_end)) @function.outer
-
-(hashfn ["#" "hashfn"] @function.outer.start (_) @function.inner) @function.outer
-
-; parameters
-(parameters (_) @parameter.inner)
-(parameters (_) @parameter.outer)
+; functions & parameters
+(list .
+  (symbol) @_fn 
+  (symbol)? @_fn_name 
+  (sequence
+    (_) @parameter.inner) 
+  (_)* @function.inner
+  (#any-of? @_fn "fn" "lambda" "λ")) @function.outer 
 
 ; call
-((list . [(multi_symbol) (symbol)] @_sym . (_) @_start . (_)* . (_)? @_end .)
- (#make-range! "call.inner" @_start @_end)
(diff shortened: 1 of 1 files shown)


queries/fennel/textobjects.scm
```diff
@@ -1,52 +1,34 @@
...
-(parameters (_) @parameter.inner)
-(parameters (_) @parameter.outer)
+; functions & parameters
+(list .
+  (symbol) @_fn 
+  (symbol)? @_fn_name 
+  (sequence
+    (_) @parameter.inner) 
+  (_)* @function.inner
+  (#any-of? @_fn "fn" "lambda" "λ")) @function.outer
```
davidmh, line 14: I was trying to solve the same problem, but I couldn't figure out how to solve a case that is still not being picked up by your changes.

Look at these three cases:

```fennel
(fn happy [name]
  (vim.notify (.. "hello " name "!")))

(fn sad1 []
  (vim.notify "this works!"))

(fn sad2 []
  (print "will this work?")
  (vim.notify "no 😭"))
```

We can make `sad1` pass by making the params optional:

```suggestion
(list .
  (symbol) @_fn 
  (symbol)? @_fn_name 
  (sequence
    (_)* @parameter.inner) 
  (_)* @function.inner
  (#any-of? @_fn "fn" "lambda" "λ")) @function.outer 
```

But `sad2` has a false positive on `@function.outer` because it considers every statement in the body of the function as an individual match:

Screenshot 2024-02-21 at 22 43 14

I'm not super familiar with the query syntax, do you know how can we group adjacent lists into a single match?

davidmh: I think I found it:

```suggestion
(list .
  (symbol) @_fn
  (symbol)? @_fn_name
  (sequence (_)? @parameter.inner .) @parameter.outer
  (_) @_start ((_)* (_)?) @_end
  (#make-range! "function.inner" @_start @_end)
  (#any-of? @_fn "fn" "lambda" "λ")) @function.outer
```

Not sure if there's a better way of doing it, but that covers all cases.

fsouza: Sorry things got a bit busy on my end, but I'll try to come back to this PR later today.

One thing I did to get myself unblocked was to copy these queries to my dotfiles and play with them (alternatively you can checkout this PR locally). To be honest, my first reaction was to delete all the queries so I could stop getting errors when I opened neovim hahah and then I tried to fix the queries.

nvim-treesitter will give priority to the queries defined in your local config.

davidmh: Yeah, that's what I've been doing for now: https://github.com/davidmh/dot-files/blob/af1b67e2a149c8513202a9f6a9275097d6d69016/nvim/queries/fennel/textobjects.scm

I also added support for the [hash function shorthand](https://fennel-lang.org/reference#hash-function-literal-shorthand)

alexmozaidze: I don't use textobjects myself, so I can't really recommend anything. I'm working on updating the parser to include all the forms in order to compensate for the limitations of the query language. See https://github.com/nvim-treesitter/nvim-treesitter/pull/6190 for more details.

kiyoon:
I will merge this because it is breaking at the moment. But feel free to open a new PR if you think more work needs to be done.

PR #571, opened 2024-02-22
Queries for odin

xb-bx:
Hello. 
This PR introduces queries for [odin](https://odin-lang.org/)

Thx for great plugin!

### DIFF
--- a/queries/odin/textobjects.scm
+++ b/queries/odin/textobjects.scm
@@ -1,0 +1,141 @@
+; procedures
+(procedure_declaration
+  (_
+    (block
+      .
+      "{"
+      .
+      (_) @_start @_end
+      (_)? @_end
+      .
+      "}"
+      (#make-range! "function.inner" @_start @_end)))) @function.outer
+
+; returns
+(return_statement
+  (_)? @return.inner) @return.outer
+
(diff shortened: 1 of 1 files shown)


theHamsta (approved):
Thank you! Looks good to me. I didn't have the time yet to test personally but might have on weekend. But merging would be also fine to me.

ribru17:
Would you mind rebasing on the latest `master` and running the formatter? I think then this will be good to merge. Thank you for the contribution!

xb-bx:
Okay, but what formatter should i use?

ribru17:
It should be the one found [here](https://github.com/nvim-treesitter/nvim-treesitter/blob/main/scripts/format-queries.lua): this is the one used by the main `nvim-treesitter` repo. It's probably enough to just do `nvim -l /path/to/nvim-treesitter/scripts/format-queries.lua /path/to/nvim-treesitter-textobjects/queries/odin/textobjects.scm` (if you have an up-to-date version of the `nvim-treesitter` plugin)

ribru17:
Thank you for the change and feedback! I will get on that :+1:

ribru17 (approved):

PR #574, opened 2024-02-26
Add hcl queries

rwblokzijl:
Add support for HCL:

```
@assignment.inner
@assignment.lhs
@assignment.outer
@assignment.rhs
@call.inner
@call.outer
@conditional.inner
@conditional.outer
@loop.inner
@loop.outer
@number.inner
@parameter.inner
@parameter.outer
```

Additionally added (curious what you think):
```
@conditional.condition
@conditional.lhs
@conditional.rhs

@loop.control
@loop.lhs
@loop.rhs
@loop.condition
```


ribru17:
Thank you for the PR! Would you mind running the formatter script found [here](https://github.com/nvim-treesitter/nvim-treesitter/blob/main/scripts/format-queries.lua) (and also maybe rebasing on the latest master)? Thank you so much for the contribution!

rwblokzijl:
> Thank you for the PR! Would you mind running the formatter script found [here](https://github.com/nvim-treesitter/nvim-treesitter/blob/main/scripts/format-queries.lua?rgh-link-date=2024-04-10T15%3A51%3A08Z) (and also maybe rebasing on the latest master)? Thank you so much for the contribution!

Done and done.

ribru17:
Thank you! Also sorry for the late review but I think it would be best to avoid the non-documented capture groups, at least for now. There is a more experimental branch `main` which may be more receptive to adding new groups so you could try to petition for them there, but I personally think something like `@loop.(lhs|rhs)` may not make much sense in many languages. Also, I think that the role of `@conditional.condition` is usually filled by `@conditional.inner` (and I think we should use this for the `@loop.condition` as well). What do you think?

rwblokzijl:
I think you're right. I think it would be cool to have the `lhs` and `rhs` for conditionals, however i understand how this is experimental. I have updated the PR to remove all the non-documented ones and update them with inner.

ribru17:
Thank you for the changes and contribution in general; and yes I saw don't worry :) I was waiting on this one, and I will merge it right after :+1:

ribru17 (approved):

PR #577, opened 2024-03-04
fix: nil index when swapping at first/last child

ribru17:
Apologies!! The previous PR caused a nil index error when swapping out of the node ranges (e.g. swapping with a previous node when already at the first node). This should fix it.

Lua is quirky...


kiyoon:
Can you rebase it? The CI has to pass and it's breaking so I can't merge this. The master should fix the issue.

ribru17:
Done :+1:

PR #579, opened 2024-03-05
Patch 1

archibate:
In `M.next_textobject`:
```lua
next_node and next_node.node, next_node.metadata
```
Is actually:
```lua
(next_node and next_node.node), next_node.metadata
```
`next_node.metadata` would be `nil.metadata` when next_node is nil, resulting in `attempt to index local 'next_node' (a nil value)` when I trying to use swap_next at the last argument.
`M.previous_textobject` has the same issue too. We'd better use an early return instead.


kiyoon:
Thank you, but this is already fixed with #577. Sorry for the late merging

Issue #585, opened 2024-03-18
Assignment operator do not exist for c/cpp


ilan-schemoul:
**Describe the bug**
TSTextobjectSelect do not work for assignement for c/cpp. It works for python.

**To Reproduce**
1. Go to a cpp file with any assignment such as `int ae = 42;`
2. Type command `TsTextobjectSelect @assignment.inner`
4. Nothing happens

**Expected behavior**
Inner assignment should be selected

**Output of `:checkhealth nvim-treesitter`**

<details>
<code>

==============================================================================
nvim-treesitter: require("nvim-treesitter.health").check()

Installation ~
- WARNING `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
- OK `node` found v18.19.1 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
- OK `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
  Version: cc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
- OK Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

OS Info:
{
  machine = "x86_64",
  release = "5.15.146.1-microsoft-standard-WSL2",
  sysname = "Linux",
  version = "#1 SMP Thu Jan 11 04:09:03 UTC 2024"
} ~

Parser/Features         H L F I J
  - c                   ✓ ✓ ✓ ✓ ✓
  - cpp                 ✓ ✓ ✓ ✓ ✓
  - diff                ✓ . . . .
  - lua                 ✓ ✓ ✓ ✓ ✓
  - markdown            ✓ . ✓ ✓ ✓
  - markdown_inline     ✓ . . . ✓
  - python              ✓ ✓ ✓ ✓ ✓
  - query               ✓ ✓ ✓ ✓ ✓
  - regex               ✓ . . . .
  - rust                ✓ ✓ ✓ ✓ ✓
  - vim                 ✓ ✓ ✓ . ✓
  - vimdoc              x . . . ✓

  Legend: H[ighlight], L[ocals], F[olds], I[ndents], In[j]ections
         +) multiple parsers found, only one will be used
         x) errors found in the query, try to run :TSUpdate {lang} ~

The following errors have been detected: ~
[...]

PR #587, opened 2024-03-30
feat(c): assignment operator for c

ilan-schemoul:
Problem : #585 
Solution : 
- Assignment inner/outer/left/right now fully works on declaration such as `int a = 3` and definitions such as `int a = 10`
- It does also work on expression_statement such as `a = 0` (notice there is no type here)

### DIFF
--- a/queries/c/textobjects.scm
+++ b/queries/c/textobjects.scm
@@ -116,3 +116,8 @@
 (declaration
   type: (primitive_type)
   declarator: (_) @assignment.inner)
+
+(expression_statement
+  (assignment_expression
+    left: (_) @assignment.lhs
+    right: (_) @assignment.rhs) @assignment.inner) @assignment.outer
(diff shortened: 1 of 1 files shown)


ribru17:
Thank you for the PR! Would you mind rebasing on the latest `master`? This should hopefully fix the test failures

ilan-schemoul:
@theHamsta @ribru17 do you know why run tests keep failing on nightly. It cannot create swap file but I don't see how it can be related to me ?

theHamsta:
> @theHamsta @ribru17 do you know why run tests keep failing on nightly. It cannot create swap file but I don't see how it can be related to me ?

That seems to be related to caching in the test CI. I haven't followed this repo recently so I don't know whether this has been occurring since a while.

theHamsta:
Re-running the workflow fixed the problem. Seems to be intermittent (perhaps workflow concurrency with shared cache?)

PR #588, opened 2024-03-30
Lookbehind local keymap setting

ilan-schemoul:
Problem : #156 we may want a global setting (e.g: lookahead) for most keymaps and then override that setting for one keymap (e.g: lookbehind)
You can now add a local lookbehind/lookahead in the keymap that takes precedence :

```lua
      textobjects = {
        select = {
          enable = true,
          lookahead = true,
          keymaps = {
            ["t="] = { query = "@type", desc = "Select right hand side of an assignment", lookbehind = true }, 
          }
        }
      }

```


kiyoon:
The development is now done in `main` and `master` is frozen. Can you open a new PR with this feature?

lua/nvim-treesitter/textobjects/select.lua
```diff
@@ -94,10 +94,18 @@ local val_or_return = function(val, opts)
   end
 end
 
-function M.select_textobject(query_string, query_group, keymap_mode)
+function M.select_textobject(query_string, query_group, keymap_mode, local_lookahead, local_lookbehind)
```
kiyoon, line 97: I think the name of the args could be `lookahead` and `lookbehind`. If they are not defined just use the global. Or I would name it like `override_lookahead` etc.

PR #590, opened 2024-04-02
feat(roc): add roc queries

nat-418:


clason:
Please do not merge master into your branch; always rebase instead.

PR #591, opened 2024-04-02
fix(v): reflect new comment structure in parser

ribru17:
This change fixes the CI issue related to V


clason:
We really need some better integration CI between the different repos...

ribru17:
100%

PR #593, opened 2024-04-03
feat(roc): add roc queries

nat-418:
@clason I made a new clean PR

### DIFF
--- a/queries/roc/textobjects.scm
+++ b/queries/roc/textobjects.scm
@@ -1,0 +1,35 @@
+(anon_fun_expr
+  (expr_body) @function.inner
+) @function.outer
+
+(argument_patterns
+  ((_) @parameter.inner . ","? @parameter.outer) @parameter.outer
+)
+
+(function_type
+  ((_) @parameter.inner . ","? @parameter.outer) @parameter.outer(#not-eq? @parameter.inner "->")
+)
+
+(function_call_expr
+  .
+  (_)
+  (parenthesized_expr (expr_body) @parameter.inner) @parameter.outer
+)
+
+(function_call_expr
+  .
+  (_) ((_) @parameter.inner) @parameter.outer
+)
+
+[
+  (annotation_type_def ) @class.inner
+  (alias_type_def ) @class.inner
+  (opaque_type_def ) @class.inner
+] @class.outer
+
+(apply_type_arg) @parameter.inner
+
+(line_comment) @comment.outer
+(doc_comment) @comment.outer
+
+


queries/roc/textobjects.scm
```diff
@@ -0,0 +1,35 @@
+(anon_fun_expr
+  (expr_body) @function.inner
+) @function.outer
+
+(argument_patterns
+  ((_) @parameter.inner . ","? @parameter.outer) @parameter.outer
```
ribru17, line 6: This query makes it so that `@parameter.outer` only matches the comma sometimes (maybe every time?). Can you change this to only have the second outer query which captures everything? Or if that doesn't work you may need `#make-range!`

queries/roc/textobjects.scm
```diff
@@ -0,0 +1,35 @@
...
+)
+
+(function_type
+  ((_) @parameter.inner . ","? @parameter.outer) @parameter.outer(#not-eq? @parameter.inner "->")
```
ribru17, line 10: Same as above I believe. Also we don't enforce style (yet?) on this repo but would you mind putting the predicate on a new line, just some easier reading?

queries/roc/textobjects.scm
```diff
@@ -0,0 +1,35 @@
...
+[
+  (annotation_type_def ) @class.inner
+  (alias_type_def ) @class.inner
+  (opaque_type_def ) @class.inner
+] @class.outer
+
+(apply_type_arg) @parameter.inner
+
+(line_comment) @comment.outer
+(doc_comment) @comment.outer
```
ribru17, line 33: Would you mind adding the appropriate `@comment.inner` queries (with `line_comment` offset by one character from the left and `doc_comment` by two)? If you disagree with this that's fine

queries/roc/textobjects.scm
```diff
@@ -0,0 +1,35 @@
...
+(function_call_expr
+  .
+  (_) ((_) @parameter.inner) @parameter.outer
+)
+
+[
+  (annotation_type_def ) @class.inner
+  (alias_type_def ) @class.inner
+  (opaque_type_def ) @class.inner
+] @class.outer
```
ribru17, line 28: I think this query is a bit misleading, it basically just applies `@class.inner` and `@class.outer` to all elements in this list. Would you mind either moving it to be `[ ... ] @class.inner @class.outer` or doing something else you feel appropriate if you meant there to be a difference between these queries? I don't know this language at all so do what you think is best

queries/roc/textobjects.scm
```diff
@@ -0,0 +1,35 @@
...
+
+(function_call_expr
+  .
+  (_)
+  (parenthesized_expr (expr_body) @parameter.inner) @parameter.outer
+)
+
+(function_call_expr
+  .
+  (_) ((_) @parameter.inner) @parameter.outer
```
ribru17, line 21: These parentheses are also redundant; if this was really meant to do what it is doing (apply inner and outer to the same thing) you can just do `(_) @parameter.inner @parameter.outer`

queries/roc/textobjects.scm
clason: We should add the query linter from `nvim-treesitter`; I don't think we should support a different style here. (You can just download/copy the file from that repo; it's a standalone script.)

queries/roc/textobjects.scm
nat-418: I don't understand what the problem is here. A Roc function looks like this:

```
\x -> x
```

or with multiple params:

```
\x, y -> x
```

queries/roc/textobjects.scm
nat-418: I added the query linter and ran it. I am still not sure what the issue is with this query though. If someone can explain to me in more detail I will make whatever change necessary.

https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/593#discussion_r1554554398

ribru17: Oh, thank you for doing that! Sorry I should have mentioned this was done by me a few days ago; but I think your issue is that the linter expects the output to match the `main` branch formatter: I am guessing the one you added is from `master`?  (I also did this at first!) You can just remove that file from this PR and then run your original queries through the `main` branch's formatter and it should stop complaining :)

queries/roc/textobjects.scm
ribru17: Yes; if you try it yourself, you'll find that operating on `@parameter.outer` selects only the comma itself, when it should be the entirety of, say, `x,` (with or without trailing whitespace). This is why you may need `#make-range!`. You can see some examples of other textobject files doing this if you need, this type of query is pretty common across different languages.

queries/roc/textobjects.scm
```diff
@@ -0,0 +1,39 @@
...
+
+[
+  (annotation_type_def)
+  (alias_type_def)
+  (opaque_type_def)
+] @class.inner @class.outer
+
+(apply_type_arg) @parameter.inner
+
+((#offset! line_comment 0 1 0 -1)) @comment.inner
```
ribru17, line 33: This results in an error: try looking [here](https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/f5183cea0fda26126e22e789382c208e7b1120f4/queries/v/textobjects.scm#L71) for an example of how this should be done

ribru17: Also, if you happen to be using [lazy](https://github.com/folke/lazy.nvim), you can just put your textobject queries in `~/.local/share/nvim/lazy/nvim-treesitter-textobjects/queries/roc/textobjects.scm` and restart your neovim instance and they should apply locally, if you'd like to be able to do some quick debugging.

queries/roc/textobjects.scm
nat-418: thanks for explaining. this is my first foreay into tree-sitter

clason:
Closing as stale. A fresh PR against `main` (including tests) would be welcome.

PR #599, opened 2024-04-10
rust: add self_parameter query

daver32:
Adds support for the `self` param.


ribru17:
Would you mind running the formatter to pass the linting check? (Or just manually touch it up by removing the empty line after the parameters comment and removing the trailing whitespace from the other new empty line.)

ribru17 (approved):

PR #601, opened 2024-04-10
Remove undocumented captures, fix CI (and touch up CONTRIBUTING.md)

ribru17:
Mentions the new formatter check, gives a few touch-ups.

EDIT: It also fixes the capture group CI check, which was dependent upon the `h3` heading of `Textobjects` having the same name (case insensitive) as the query group (what??? hopefully this is addressed in `main` :joy:)


clason:
> EDIT: It also fixes the capture group CI check, which was dependent upon the h3 heading of Textobjects having the same name (case insensitive) as the query group (what??? hopefully this is addressed in main 😂)

No, that's the same way it works in nvim-treesitter (which has to test multiple different queries). We can talk about pulling the queries into the main repo, but only if we know they're stable (otherwise adjusting the queries to plugin changes becomes impossible).

ribru17:
Ah, makes sense. I will amend this to remove all undocumented captures to "stabilize" things; they're query changes, so would it be fine to target `master`? Or do you think it should be `main` since it's a bit breaking maybe?

clason:
No, query changes should target master. (Nvim-treesitter had a similar "consolidation" phase recently.)

ribru17:
Oh, right. Thanks!

clason:
Alternatively, if `@function.outer.start` is useful, we could add it to the documentation and other queries. What does Helix do?

ribru17:
That sounds fine. I personally didn't think it looked too useful, and helix only has `@function.(inside|around)`

clason:
That's fine, then.

clason (approved):

PR #602, opened 2024-04-10
fix(julia): Update queries

savq:


mehalter:
@savq Should this be marked as ready for review now that the updated julia queries have been merged into `nvim-treesitter`?

savq:
Yeah, this should've been merged after the changes in nvim-treesitter. sorry about that!

PR #605, opened 2024-04-18
fix(elixir): correctly match call arguments

calops:
Function parameters weren't correctly matched in all situations before. The previous queries seemed to be a bit complex to me and I fail to identify why, but maybe I'm missing something?

### DIFF
--- a/queries/elixir/textobjects.scm
+++ b/queries/elixir/textobjects.scm
@@ -51,19 +51,19 @@
     right: (body) @function.inner)) @function.outer
 
 (call
+  [
+    (arguments
+      (_) @parameter.inner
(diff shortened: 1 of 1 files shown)


ribru17:
I don't know if this is the right change; the elixir parser is very weird sadly (I don't know why): even in something like the following:
```ex
if a === true do
   IO.puts "Variable a is true!"
end
```
this entire block is parsed as a `(call)` node (what?)! Perhaps there is a very good reason that I am not aware of but either way it makes it difficult to just label anything in a `(call)` as a `@parameter`

clason:
Closing as stale.

PR #606, opened 2024-04-22
feat(fennel): update queries for new parser

alexmozaidze:
Updated queries to support the new parser.


ObserverOfTime (approved):

alexmozaidze:
I wonder, is it faster/better to do this
```scm
(_
  .
  "("
  ")" .) @statement.outer
```
Or this?
```scm
(_
  open: "("
  close: ")") @statement.outer
```

PR #607, opened 2024-04-22
[main] Fix: regressions after refactor

TheLeoP:
The change to `TsTextObjects.Range` leaved some nonsensical code after the refactor, this has been adressed.

Also, `TsTextObjects.Range.is_in_range` wasn''t working as expected, this has been fixed.


przepompownia:
For the record I copy the repro from the source discussion

`init.lua`:

```lua
local thisInitFile = debug.getinfo(1).source:match('@?(.*)')
local configDir = vim.fs.dirname(thisInitFile)

vim.env['XDG_CONFIG_HOME'] = configDir
vim.env['XDG_DATA_HOME'] = vim.fs.joinpath(configDir, '.xdg', 'data')
vim.env['XDG_STATE_HOME'] = vim.fs.joinpath(configDir, '.xdg', 'state')
vim.env['XDG_CACHE_HOME'] = vim.fs.joinpath(configDir, '.xdg', 'cache')
local stdPathConfig = vim.fn.stdpath('config')

vim.opt.runtimepath:prepend(stdPathConfig)
vim.opt.packpath:prepend(stdPathConfig)

local function gitClone(url, installPath, branch)
  if vim.fn.isdirectory(installPath) ~= 0 then
    return
  end

  local command = {'git', 'clone', '--', url, installPath}
  if branch then
    table.insert(command, 3, '--branch')
    table.insert(command, 4, branch)
  end
  local sysObj = vim.system(command, {}):wait()
  if sysObj.code ~= 0 then
    error(sysObj.stderr)
  end
  vim.notify(sysObj.stdout)
  vim.notify(sysObj.stderr, vim.log.levels.WARN)
end

local pluginsPath = 'nvim/pack/plugins/opt'
vim.fn.mkdir(pluginsPath, 'p')
pluginsPath = vim.uv.fs_realpath(pluginsPath)

--- @type table<string, {url:string, branch: string?}>
local plugins = {
--  ['nvim-treesitter'] = {url = 'https://github.com/nvim-treesitter/nvim-treesitter'},
  ['nvim-treesitter-textobjects'] = {url = 'https://github.com/nvim-treesitter/nvim-treesitter-textobjects', branch = 'main'}
}

for name, repo in pairs(plugins) do
  local installPath = vim.fs.joinpath(pluginsPath, name)
  gitClone(repo.url, installPath, repo.branch)
  vim.cmd.packadd({args = {name}, bang = true})
end

local tsSelectKeymaps = {
  ['ap'] = '@parameter.outer',
  ['ip'] = '@parameter.inner',
}

for input, capture in pairs(tsSelectKeymaps) do
  vim.keymap.set({'x', 'o'}, input, function ()
    require 'nvim-treesitter-textobjects.select'.select_textobject(capture, 'textobjects')
  end)
end
```

```sh
nvim --clean -u init.lua init.lua
```

Try `vip` on the second parameter of `gitClone` declaration (line 13), third param in line 21 (`branch`) and not only.


One obvious exception: the branch is from this PR. 

I still get 
```
E5108: Error executing lua: ...from-src/share/nvim/runtime/lua/vim/treesitter/query.lua:788: No handler for make-range!
stack traceback:
        [C]: in function 'error'
        ...from-src/share/nvim/runtime/lua/vim/treesitter/query.lua:788: in function 'apply_directives'
        ...from-src/share/nvim/runtime/lua/vim/treesitter/query.lua:959: in function 'matches'
        ...r-textobjects/lua/nvim-treesitter-textobjects/shared.lua:42: in function '(for generator)'
        ...r-textobjects/lua/nvim-treesitter-textobjects/shared.lua:147: in function 'fn'
        ...c/share/nvim/runtime/lua/vim/treesitter/languagetree.lua:488: in function 'for_each_tree'
        ...r-textobjects/lua/nvim-treesitter-textobjects/shared.lua:144: in function 'get_capture_ranges_recursively'
        ...r-textobjects/lua/nvim-treesitter-textobjects/shared.lua:379: in function 'textobject_at_point'
        ...r-textobjects/lua/nvim-treesitter-textobjects/select.lua:186: in function 'select_textobject'
        ...ctgx/nvim/envs/nvim-treesitter-textobjects-main/init.lua:54: in function <...ctgx/nvim/envs/nvim-treesitter-textobjects-main/init.lua:53>
```

TheLeoP:
@przepompownia Thanks for the report and the repro. Could you try this PR again?

przepompownia:
@TheLeoP now on the repro I get correct parameter selection. Thanks for fixing.

towry:
```
E5108: Error executing lua: .../nvim-treesitter-textobjects/plugin/query_predicates.lua:15: #make-range! does not supp
ort captures on multiple nodes
stack traceback:
        [C]: in function 'assert'
        .../nvim-treesitter-textobjects/plugin/query_predicates.lua:15: in function 'handler'
        /usr/local/share/nvim/runtime/lua/vim/treesitter/query.lua:791: in function 'apply_directives'
        /usr/local/share/nvim/runtime/lua/vim/treesitter/query.lua:959: in function 'matches'
        ...r-textobjects/lua/nvim-treesitter-textobjects/shared.lua:44: in function '(for generator)'
        ...r-textobjects/lua/nvim-treesitter-textobjects/shared.lua:177: in function 'find_best_range'
        ...ter-textobjects/lua/nvim-treesitter-textobjects/move.lua:136: in function 'move_fn'
        ...ects/lua/nvim-treesitter-textobjects/repeatable_move.lua:75: in function 'move_repeatable'
        ...ter-textobjects/lua/nvim-treesitter-textobjects/move.lua:208: in function <...ter-textobjects/lua/nvim-tree
sitter-textobjects/move.lua:207>
        /Users/towry/.config/nvim/lua/user/plugins/lang.lua:88: in function </Users/towry/.config/nvim/lua/user/plugin
s/lang.lua:86>
```

towry:
I have a attempt fix that works.

```lua
  local end_node_id = predicate[4]

  local start_nodes = match[start_node_id]
  local end_nodes = match[end_node_id]

  -- if start_nodes then
  --   assert(#start_nodes <= 1, "#make-range! does not support captures on multiple nodes")
  -- end
  if end_nodes then
    if #end_nodes > 1 then
      # select last, works for select @function.inner and @function.outer
      end_nodes = { end_nodes[#end_nodes] }
    end
  end
```

TheLeoP:
This is happening because, with the previous behaviour of `iter_matches`, if multiple nodes where captured under the same name, only the last node would be taken into account. Queries as the one used for typescript/javascript rely in this previous behaviour. For example:

```scheme
(function_declaration
  body: (statement_block
    .
    "{"
    .
    (_) @_start @_end
    (_)? @_end
    .
    "}"
    (#make-range! "function.inner" @_start @_end)))
```

The implementation of `make-range!` already defaults the end of the range to the start node if there is no end node (and viceversa). So, unless I'm missing something, updating the captures to no longer have this fallback behaviour that results in multiple nodes being captured sometimes would also be a solution. I would prefer this option because I don't think it makes sense to handle multiple nodes being captured on `make-range!`, but I may be missing some edge case.

@clason is there a preferred approach for solving this?

clason:
To be honest, the preferred approach is to avoid the fake `#make-range!` directive in the first place... But that would be a much larger refactor.

For other predicates, we have deliberately added `any-` variants to distinguish this, but that doesn't make much sense here.

clason:
> As a side note. Do you know if shome recent change in nvim-treesitter may have cause the install-parsers.lua´ script to stop working properly?. It looks like it's hanging on html-tags` because neovim is asking if the parsers should be reinstalled or not.

Yes, there was a _huge_ breaking change (which I'm still fiddling with). It _will_ reinstall dependencies multiple times (can't really be helped due to async, but it's something I am looking at), but it shouldn't hang. Can you try the current version?

TheLeoP:
Thanks, that was it c:

ribru17:
Is this ready to be merged?

TheLeoP:
Yes. I just squashed all the commits

ribru17:
Thanks!!

PR #609, opened 2024-05-02
feat: add Tact queries

novusnota:
Adding `textobjects.scm` into `queries/tact/` (formatted according to [format-queries.lua](https://github.com/nvim-treesitter/nvim-treesitter/blob/main/scripts/format-queries.lua)).

Tree-sitter grammar/parser for Tact lives here: https://github.com/tact-lang/tree-sitter-tact


queries/tact/textobjects.scm
```diff
@@ -0,0 +1,278 @@
...
+    (_) @_start
+    (_)? @_end
+    ")")
+  (#make-range! "call.inner" @_start @_end)) @call.outer
+
+(instance_expression
+  arguments: (instance_argument_list
+    "{"
+    (_) @_start
+    (_)? @_end
```
ribru17, line 216: Right now this query will match at most two inner argument nodes (same with the other similar ones): usually some sort of `(_)*` is needed; but if the argument list is always surrounded by brackets would it be possible to just use `#offset! 0 1 0 -1` here?

queries/tact/textobjects.scm
```diff
@@ -0,0 +1,278 @@
...
+; ----------------------
+; statis
+(static_function
+  body: (_)) @function.outer
+
+(static_function
+  body: (function_body
+    "{"
+    (_) @_start @_end
+    (_)? @_end
```
ribru17, line 12: This has the same issue, the pattern only matches the start and then an optional node following it. You may need anchors to ensure that the final optional `@_end` node is actually the end node: something like `. "{" . (_) @_start (_)? @_end . "}"` (also the `@_end` capture on the `@_start` capture is not needed)

queries/tact/textobjects.scm
```diff
@@ -0,0 +1,278 @@
...
+
+(field
+  name: (_) @assignment.lhs
+  value: (_) @assignment.inner @assignment.rhs) @assignment.outer
+
+; comment.inner & outer
+; ---------------------
+(comment) @comment.inner
+
+(comment)+ @comment.outer
```
ribru17, line 274: Unfortunately quantified captures aren't supported here yet (will change soon after Neovim 0.10) so maybe best to comment this query out for now and just tack on `@comment.outer` to the inner query too

queries/tact/textobjects.scm
```diff
@@ -0,0 +1,278 @@
...
+(if_statement) @conditional.outer
+
+; block.inner & outer
+; -------------------
+(_
+  (block_statement) @block.inner) @block.outer
+
+; call.inner & outer
+; ------------------
+(method_call_expression
```
ribru17, line 196: Just a nit but it looks like the parser allows a wide node range for `(method_call_expression)`, might be best to form the `@call.(outer|inner)` queries out of just 
```
name: (identifier)
arguments: (argument_list)
```
 pairs?

novusnota:
@ribru17 Thanks for your suggestions, I'll look into them only in couple of hours :)

ribru17:
Thanks! No rush :)

novusnota:
So, I've looked how JS/TS and Rust did their textobjects (they're both somewhat syntactically similar) and fine-tuned my ones, while applying your suggestions. Hope it's ok now. Also, there's a weird CI/CD error with Neovim swap file failures, I'm not sure it is caused by my parser :)

queries/tact/textobjects.scm
novusnota: Hmm, I don't know for sure here. For now went with a more involved approach, but if that becomes a bottleneck somehow, then it could be simplified to your form

queries/tact/textobjects.scm
```diff
@@ -0,0 +1,357 @@
...
+; static
+(static_function
+  body: (_)) @function.outer
+
+(static_function
+  body: (function_body
+    .
+    "{"
+    .
+    (_) @_start @_end
```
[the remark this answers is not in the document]
novusnota: Sure!

queries/tact/textobjects.scm
```diff
@@ -0,0 +1,357 @@
...
+; number.inner
+; ------------
+(integer) @number.inner
+
+; parameter.inner & outer
+; -----------------------
+(parameter_list
+  ((_) @parameter.inner
+    .
+    ","? @parameter.outer) @parameter.outer)
```
ribru17, line 326: Sorry one last comment and I think it should be good to merge: I think here you will need to use `#make-range!` (you can see some rust examples for this maybe?)- right now (this will also be fixed after 1.0 release :sweat_smile:) the `@parameter.outer` query will only capture the comma itself. I think you will need to combine the ranges for it to work properly

Also not a huge deal but I think these queries don't apply to something like the `(receive_function)` node

novusnota: Yup, fixed those queries

ribru17:
Sorry, looks like the haskell queries need to be updated. Will merge soon :+1:

ribru17:
Ok would you mind rebasing? I have merged the Haskell changes that hopefully fix the CI and allow me to merge this one

novusnota:
Now it's C# :)

ribru17:
:man_facepalming: So sorry, let me take a look...

ribru17:
Alright, fingers crossed that a rebase will turn CI green now...

ribru17:
Thank you so much for your patience :pray:

novusnota:
@ribru17 Sure thing, and massive thanks to you for pushing the progress forward and helping out with queries here and for Haskell/C#!

PR #610, opened 2024-05-04
feat(inko): add queries for Inko

yorickpeterse:
This adds text object support for [Inko](https://inko-lang.org/).

I wasn't too sure which query file to use as a reference since they differ a bit, so I based the implementation on the queries file used for Rust.


queries/inko/textobjects.scm
```diff
@@ -0,0 +1,262 @@
...
+    (#make-range! "class.inner" @_start @_end)))
+
+(reopen_class) @class.outer
+
+(reopen_class
+  body: (reopen_class_body
+    .
+    "{"
+    .
+    (_) @_start @_end
```
ribru17, line 50: Would you mind removing the `@_end` captures from the `@_start` captures? Sorry this is something we are going to remove across the repo soon

queries/inko/textobjects.scm
```diff
@@ -0,0 +1,262 @@
...
+    (_)? @_end
+    .
+    "}"
+    (#make-range! "loop.inner" @_start @_end))) @loop.outer
+
+; Conditionals
```
ribru17, line 93: Would you mind adding something like
```
(while
  condition: (_) @conditional.inner)
```
(or an equivalent that you deem most appropriate)?

queries/inko/textobjects.scm
```diff
@@ -0,0 +1,262 @@
...
+    "{"
+    .
+    (_) @_start @_end
+    (_)? @_end
+    .
+    "}"
+    (#make-range! "class.inner" @_start @_end)))
+
+; Methods
+(method) @function.outer
```
ribru17, line 57: Should these be given to `(closure)`?

ribru17:
Thank you!

queries/inko/textobjects.scm
```diff
@@ -0,0 +1,278 @@
...
+(method
+  body: (block
+    .
+    "{"
+    .
+    (_) @_start
+    (_)? @_end
+    .
+    "}"
+    (#make-range! "function.inner" @_start)))
```
ribru17, line 68: Looks like this directive and many (all?) others are missing the second parameter: should be (for this one)
```query
(#make-range! "function.inner" @_start @_end)
```

yorickpeterse: @ribru17 I removed this in response to your request in https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/610#discussion_r1595773398. Did you mean to only remove `@_end` from certain nodes (and if so, which ones)?

ribru17: Sorry, I should have been clearer. Basically, the `@_end` should be removed in `(_) @_start @_end`, but not in the `#make-range!` predicate. That way the make range still uses the end capture, but makes sure it always uses the last end capture (not the one that was also applied to the start capture)

yorickpeterse: @ribru17 Ah gotcha, this should be fixed now :smiley:

ribru17: Thank you!

ribru17 (approved):

PR #612, opened 2024-05-10
feat: support quantified captures

ribru17:
This commit also removes all uses of `#make-range!` as it is no longer necessary.


kiyoon:
Good to see this change as I've been waiting for this for quite a while! I guess this requires nightly nvim (0.10)?

ribru17:
It does sadly, hopefully we can merge the main branch soon after 0.10 releases!

kiyoon:
Maybe some people use custom captures or captures from other plugins. Should we deprecate the use of `make-range!` but remove the code a little later so it won't break any?

ribru17:
Good point: up to @clason I think. This branch is already pretty breaking so removing the directive might not matter (or maybe we should just add the fake `function() end` placeholder directive to prevent errors?) I don't mind either way

kiyoon:
Since we are going to keep the queries in sync with the master, the main branch will likely have `#make-range!` directives for some time. And merging this will break this syncronisation.

clason:
> It does sadly, hopefully we can merge the main branch soon after 0.10 releases!

No, we will definitely not do that! The `main` branch will become the default, but `master` will remain (possibly after renaming) as a legacy branch. This will happen together with the corresponding switch for nvim-treesitter (and nvim-treesitter-refactor) so people can upgrade to `main` + `main` or stay on `master` + `master` (but not get updates).

ribru17:
> No, we will definitely not do that!

Right, sorry :sweat_smile: I remembered you saying that after I left that comment...

> But now that @TheLeoP converted this into a proper directive, this is much less urgent.

I see: so should that directive code be kept around on `main` for a bit, like @kiyoon says?

clason:
I think so, yes. I do like the change, but keeping queries in sync (for now) makes developing both branches in parallel much easier.

clason:
In fact, I would make this change on `master` (in general, PRs should _never_ target `main` unless they're related to the upstream refactor) as soon as we can require Nvim 0.10 (probably sometime after the first bugfix release, to give people time to update).

The "support matrix" for Nvim in the nvim-treesitter org is
* latest release
* latest commit on HEAD

(That doesn't mean that we'll aggressively drop support for older versions, but we don't make _any_ long-term support guarantees. Delaying updates is just common courtesy.)

clason:
Although I'm not 100% sure quantified captures are fully equivalent to `make-range`? I _think_ the latter could create a range that spans _other_ captures as well, which might be useful?

ribru17:
I think it would be difficult to target `master`: or at least, the changes would have to be applied to the regular `nvim-treesitter` iterator utils which this repo uses, then we could apply just the query changes to this repo.

> Although I'm not 100% sure quantified captures are fully equivalent to `make-range`?

Everything I have tested (which, granted, wasn't every single query change but *was* most unique types of queries that  query files use) has shown the behavior to be identical. Hopefully it just always works, but not 100% positive about some wacky edge cases

clason:
> I think it would be difficult to target master: or at least, the changes would have to be applied to the regular nvim-treesitter iterator utils which this repo uses, then we could apply just the query changes to this repo.

If this would help the transition, I'd be willing to merge a PR *iff* 1. it only changes stuff that is dropped on `main` anyway and 2. is unlikely to be breaking for other users.

> Everything I have tested (which, granted, wasn't every single query change but was most unique types of queries that query files use) has shown the behavior to be identical. Hopefully it just always works, but not 100% positive about some wacky edge cases

I was thinking more that `make-range` is (much?) more general, even if current queries don't use that (yet). I'm also not unsure if other plugins use it (but that's not necessarily a problem for changes in nvim-treesitter-textobjects).

ribru17:
Sounds good, I'll mark as draft and sit on it until 0.10 :+1:

ribru17:
I'm wondering if it is still better to target this on `master`. Doing some rebasing now and I'm realizing that the capture iteration logic resides in the `nvim-treesitter` utils that this repo inherits. I was able to change it easily on `main` since it was all standalone logic. For `master`, I think there would have to be some API changes on `nvim-treesitter`'s end to allow for `{ all = true }`. Does that sound possible? Or should this be left to target `main`?

kiyoon:
From what I understand we would target that on master and rebase it on main when nvim-0.10.1 is out.

clason:
I think there are a number of possible simplifications by relying on Nvim core APIs rather than nvim-treesitter utils left on the table; these would be worthwhile in any case.

I see two options:

1. Merge this now to  `main` and essentially freeze the code on `master` to prevent rebase conflicts (any possible conflicts to queries are fairly easy to resolve manually); this is how I handle nvim-treesittter.
2. Change the logic in nvim-treesitter `master` and merge this to `master` (and make followup changes to `main`).

I am not averse to 2. in principle, but the decision would depend on the exact size of changes (conflicts when rebasing, dropping 0.9.x support or extensive comaptibility shims etc.)

kiyoon:
I also don't mind freezing `master` and focusing on updating `main` which won't break much for the existing users and even encourage them to migrate.

clason:
I wouldn't encourage people to migrate ... yet.

ribru17:
I'll opt for solution 1. then?

clason:
That certainly works; if you want my opinion on option 2, I would need more details ;)

ribru17:
I think this is good to go now. I have removed the `make-range` predicate definition; I know we had a discussion that ended with deciding to keep it but I think if plugins use it that means they are also using the `nvim-treesitter` iteration utils which account for `make-range`. So I think the only backwards-compatibility this would break would be for user `textobjects` queries that have `make-range`. Should I still keep the `make-range` definition for this case?

clason:
My concern was that `make-range` seems strictly more powerful; whether that power is used currently I don't know.

(And whether the implementation actually works correctly...)

clason:
>  right now the range is calculated as . Same logic as make-range

Ok, then there's really no reason. I was thinking -- as has been requested a few times -- that it's $\overline{\mathrm{co}} (\textbf{first-captured-node} \cup \textbf{last-captured-node})$ (so includes _everything_ between the two captured nodes).

ribru17:
Oops, sorry. That is how it works, everything in between the first and last. My (lack of) mathematical notation knowledge fails me :sweat_smile:

clason:
Ok, so there is worth in the directive; the question is whether we need it _here_.

ribru17:
`make-range` also included everything between ranges; the logic [here](https://github.com/nvim-treesitter/nvim-treesitter/blob/c5cbd3ec74f6f5ddbac939e6f24b99fe78262b4c/lua/nvim-treesitter/tsrange.lua#L30) is what I referenced for this PR. So I believe the behavior is identical and thus the directive is not needed here. But it may be worthwhile to add a `make-range` stub here that warns users that it is deprecated and they should use newer query methods.

clason:
The idea is that make-range could capture the range between different nodes (and not just for text objects -- which is out of scope for this repo, of course).

clason:
@ribru17 do we want to go through with this? if so, it needs a (big) rebase. 

(I worry a bit about reusing `make-range` in an incompatible fashion, but maybe we can keep the "make range between" implementation compatible in some way. That would be a follow-up, though, and probably better discussed in core.)

clason:
Actually, now that Nvim 0.10 has been out a while and had its first bugfix release, I feel much more comfortable bumping the minimum version on `master`. So how about this:

1. on `master`, require Nvim 0.10 and update the queries but leave the directive untouched;
2. rebase `main` to pull in the new queries and then remove `make-range` (and other simplifications).

It's a bit more work but strikes me as the least painful approach? @ribru17 @kiyoon

ribru17:
This sounds good to me, I can rebase today. I was even going to ask your thoughts on using such a large PR to migrate queries to the main nvim-treesitter repo? Maybe this is breaking and should wait till 1.0 (or never :() but it seemed like a decent opportunity with this PR, and would help maintainability a lot imo (relating to CI)

clason:
We've discussed that a lot, but never came to a good conclusion. On the one hand, having the queries in nvim-treesitter makes it easier to prevent breakage when updating parsers (which is a _big_ plus). On the other hand, it makes it harder to make bigger changes to the plugin itself that requires changing the captures. It also goes counter to the goal of making these plugins stand-alone and moving nvim-treesitter away from being a mandatory single point of failure... 

I feel that _if_ we do this, we should also (only!) move the query handling part into nvim-treesitter (similar to what `locals.lua` does) and leave only the user-facing parts (selection, movement, swap -- if that remains) here. It would essentially mean `master` is frozen, though, since I don't want to manually port updates between here and nvim-treesitter.

In any case, that would only affect `main`, and the move to quantified captures on `master` would still be good.

clason:
Pro tip: `rebase -i main` and then just drop the first commit(s).

clason:
(You might want to pull in the changes from my CI PR, at least the ones regarding the query file check -- I got the cache working again ;))

clason:
Do I understand this correctly that here quantified captures get _extended_ (i.e., you get the convex hull of the union of the individual node ranges) -- differently from how they behave for highlights and injections? (And folds?)

ribru17:
I believe it is the same as folds, but yeah different from the other ones. If we think of the `make-range` functionality as a block box, it is identical to that. Only difference is instead of passing nodes via predicate arguments, nodes are passed by <firstnode>, <lastnode>, for a given quantified capture group

clason:
I think that's fine (although personally I always prefer explicitness over convenience), but I think this should then be documented. (Also for folds, but that's a separate upstream issue.)

(I do worry a bit about the potential issue showing a situation where you would want to have separate textobjects, but we can burn that bridge when we get to it.)

ribru17:
``

> (I do worry a bit about the potential issue showing a situation where you would want to have separate textobjects, but we can burn that bridge when we get to it.)

Wdym? afaics the behavior is identical to make-range; yes the calling convention is more explicit but I don't think there are any cases where the predicate could solve something that the quantified style couldn't

clason:
I mean the opposite: a situation where you _don't_ want a range, and previously could have just omitted the directive (but now can't since the behavior is builtin).

ribru17:
Ah I see what you mean. I thought about that as well. But I think this isn't an issue (it just might mandate slightly more explicit queries). E.g.:
```
(func
  ; this no longer works bc its combined
  (param) @parameter.outer
  (type_param) @parameter.outer)
```
We can get the original behavior with:

```
(func
  (param) @parameter.outer)

(func
  (type_param) @parameter.outer)
```
separate queries will not merge ranges. And of course, if we are dependent on the sibling interaction:
```
(func
  (param)
  (type_param) @parameter.outer)

(func
  (type) @parameter.outer
  (type_param))
```
So again, more verbose but at least the original functionality can still be achieved. And I think situations like these would be rare-ish anyway

clason:
Yes, I didn't think this was a blocker; just mentioned it for the sake of completeness.

ofseed:
Is `make-range` already not taking any effect? Is it safe to delete reading the metadata it created in the code?

ofseed:
Can we make this PR target `master` and move on it? If we keep `plugin/query_predicates.lua`, there should not be any breaking change on `master`. Users can still use `#make-range` in their custom queries. The cost should only be this plugin no longer supports nvim version under 0.10.

clason:
Would need to document the new requirement and a tag of the last version compatible with Nvim 0.9, but otherwise no opposition.

ribru17:
I think this was discussed before; even though I would love to target master (less rebases), it seems not as simple as that since as of now we iterate over captures with ts_utils instead of the core api. So that would have to be migrated as well... I wonder if we could do it easily and without breaking anything such that that change could also target master? And then we get a partial jump start on syncing master and main...

ofseed:
`main` removed some features such as lsp_interop, and there's a lot of refactoring after that, so I think we've lost the chance to partially migrate some of `main`'s code into `master` already...except for queries, these two branches may have been completely forked, so we may have to develop master separately, or freeze it directly and switch to main.

The purpose of wanting this PR target master is to ensure that queries(which use these quantified captures) are still available to master in the future, as we'll drop `#make-range!` in `main`. I still think it is worth having a try, but I don't know whether it could be easily done too.

clason:
> we may have to develop master separately

That is not the plan; we don't have the resources for that. Master is essentially frozen as a plugin, but it is still the primary target for query updates; _that_ is what will change. (And since query updates are _necessary_ to keep the plugin from working with current parsers, that means basically abandoning `master`.)

So, yes, the _only_ goal here would be to keep the queries in sync for longer. If that is not feasible, we should just switch -- but then the mapping setup needs to be fixed first; that's a hard prerequisite (and the last one).

> So that would have to be migrated as well... I wonder if we could do it easily and without breaking anything such that that change could also target master?

It's worth spending maybe twenty minutes looking into that and seeing if it's feasible (meaning, changing _just_ that without triggering a bigger refactor that gets us halfway to main). The goal would be the minimal change to allow quantified queries with Nvim 0.10. If it doesn't look that way, we should just pull the trigger and start the switchover by decoupling the queries (and effectively freezing `master` completely).

ribru17:
> It's worth spending maybe twenty minutes looking into that and seeing if it's feasible (meaning, changing _just_ that without triggering a bigger refactor that gets us halfway to main). The goal would be the minimal change to allow quantified queries with Nvim 0.10. If it doesn't look that way, we should just pull the trigger and start the switchover by decoupling the queries (and effectively freezing `master` completely).GG

I spent a couple hours looking at this and the change seems pretty annoying :/. Also there would be several conflicts with main I'm sure. But if someone else wants to take a stab at it they are more than welcome

clason:
Nah, then fuck it. Let's go all in on `main`.

ribru17:
Man `main` is still in a rough shape it looks like... seems like `outer` selections dont work for operator pending modes, and inners dont work for selections :laughing: After this commit, they still wont. But quantified captures will! :upside_down_face: I also fixed up a bit of the `shared` code, now range metadata is captured properly (it was not before). Let me know if anything else needs work, I want to try and get this merged this as soon as possible

clason:
Yeah, it's a mess; test coverage is... partial. (Most of the breakage is probably my fault.)

Feel free to merge; as long as things improve, no sense to hold back.

clason:
But please make a note of exactly what doesn't work in your tests (exact steps) so we can add them to the list for at least manual checking.

lua/nvim-treesitter-textobjects/shared.lua
```diff
@@ -75,20 +76,26 @@ local get_query_matches = memoize(function(bufnr, query_group, root, root_lang)
   local matches = {}
   local start_row, _, end_row, _ = root:range()
   -- The end row is exclusive so we need to add 1 to it.
-  for pattern, match, metadata in query:iter_matches(root, bufnr, start_row, end_row + 1) do
+  for pattern, match, metadata in query:iter_matches(root, bufnr, start_row, end_row + 1, { all = true }) do
```
clason, line 79: That will no longer be necessary in the near future; we should target nightly on this branch.

ribru17: Sure, I'll remove it and hopefully we can merge after https://github.com/neovim/neovim/pull/30193 lands?

clason: Yeah, it won't take long. I can take a stab at docs tomorrow; in any case, that shouldn't block it so I might just YOLO merge.

ribru17: Nevermind I read your prior comment

ribru17:
Is there a list for manual testing scenarios or would you like me to comment them here?

ofseed:
I noticed there's something wrong with the metadata after you pointed out we could not handle `#offset!`, so there is https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/676, but it looks like your fix is ​​more comprehensive.

If it doesn't work, please tell me the specific scenario and I will check what's going on. in #676?

ribru17:
Oh thanks! Sorry I didn't see your PR. Yours looks fine too, but I wonder if add_bytes is the better way to go about things because I don't know if e.g. offset! changes the byte section of the range as well?...

ofseed:
> changes the byte section of the range as well?...

It does not, and I think it never does. The range (modified by #offset!) is separately stored in the past, only when functions do not need the bytes info, it will be used, otherwise will always use the range provided by `node:range(true)`, which does not apply the offsets.

lua/nvim-treesitter-textobjects/shared.lua
clason: Neovim PR is merged; feel free to remove now (need to wait till tomorrow for nightlies -- and thus CI -- to have that, but we don't necessarily need to care about this on `main`).

PR #613, opened 2024-05-10
fix(haskell): update queries for new parser

ribru17:
The parser had a rewrite giving lots of invalid nodes. This fixes all the errors (but maybe too lazily in some cases? not 100% confident about these changes so hopefully someone can confirm they are valid)

### DIFF
--- a/queries/haskell/textobjects.scm
+++ b/queries/haskell/textobjects.scm
@@ -1,6 +1,6 @@
-((exp_apply
+((apply
   .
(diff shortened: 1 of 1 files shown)


mrcjkb:
Thanks @ribru17

I've just tested them and they seem to be working fine. Only the infix one I couldn't find a use for.
I've added some small tweaks in #623 if you want to check them out :smile:

PR #619, opened 2024-05-18
Added islist compatibility function for Neovim 0.12+

ColinKennedy:
See: https://neovim.io/doc/user/deprecated.html#deprecated-0.10

`Deprecated vim.tbl_islist() Use vim.islist() instead.`


theHamsta (approved):
Thank you!

ColinKennedy:
@theHamsta should be good now

PR #621, opened 2024-05-18
ci: remove consistency_tests from CI

theHamsta:
The consistency tests have been failing for a while. I personally, don't exactly know how they are working and find the failing tests therefore not helpful but confusing.


kiyoon:
I don't mind removing it if you wish, but when has it been failing?

clason:
I'm afraid I agree; this test setup is a bit outside my comfort zone. Neovim has a standard test runner (`busted`), and adding other ecosystems into the mix (especially Python...) simply adds too many points of failure. If these tests are testing something useful that is not covered in our busted tests, we should move them there.

PR #622, opened 2024-05-21
fix: builtin find not working with dot repeat

### LINKED ISSUE #519: Mapping repeating movements as recommended breaks some repeated commands
**Describe the bug**
I've mapped repeat motions as described in the readme:
```
-- Optionally, make builtin f, F, t, T also repeatable with ; and ,
vim.keymap.set({ "n", "x", "o" }, "f", ts_repeat_move.builtin_f)
vim.keymap.set({ "n", "x", "o" }, "F", ts_repeat_move.builtin_F)
vim.keymap.set({ "n", "x", "o" }, "t", ts_repeat_move.builtin_t)
vim.keymap.set({ "n", "x", "o" }, "T", ts_repeat_move.builtin_T)
```
However now, when I want to repeat an action, such as `dt<space>` using `.`,
instead of repeating that action, I'm somehow in operator-pending mode
after pressing the dot. So to repeat `dt<space>` I have to press `.<space>`
instead of just `.`

Commenting these lines out fixed the issue for me as a workaround.

**To Reproduce**
Set the repeat movement binds [recommended in the Readme](https://github.com/nvim-treesitter/nvim-treesitter-textobjects#text-objects-move).
Try to repeat an action that includes these commands.
You'll find that the repeat command doesn't correctly reply
the specifier of the movement key (until which letter should it apply the action).

**Expected behavior**
Repeat the last action, without forgetting the last specifier.

**Output of `:checkhealth nvim-treesitter`**

<details>
<code>

Installation ~
- WARNING `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
- OK `node` found v21.1.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
- OK `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
  Version: cc (GCC) 13.2.1 20230801
- OK Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

OS Info:
{
  machine = "x86_64",
  release = "6.5.9-arch2-1",
  sysname = "Linux",
  version = "#1 SMP PREEMPT_DYNAMIC Thu, 26 Oct 2023 00:52:20 +0000"
} ~

Parser/Features         H L F I J
  - bash                ✓ ✓ ✓ . ✓
  - c                   ✓ ✓ ✓ ✓ ✓
  - cpp                 ✓ ✓ ✓ ✓ ✓
  - diff                ✓ . . . .
  - dockerfile          ✓ . . . ✓
  - lua                 ✓ ✓ ✓ ✓ ✓
  - markdown            ✓ . ✓ ✓ ✓
  - python              ✓ ✓ ✓ ✓ ✓
  - query               ✓ ✓ ✓ ✓ ✓
  - rust                ✓ ✓ ✓ ✓ ✓
  - toml                ✓ ✓ ✓ ✓ ✓
  - vim                 ✓ ✓ ✓ . ✓
  - vimdoc              ✓ . . . ✓
  - yaml                ✓ ✓ ✓ ✓ ✓

         x) errors found in the query, try to run :TSUpdate {lang} ~

</code>
</details>

**Output of `nvim --version`**
```
NVIM v0.9.4
Build type: Release
LuaJIT 2.1.1697887905

```

kiyoon:
fixes #519 

The current implementation of builtin `f`, `F`, `t`, `T` is a re-implementation to match the behaviour of the nvim upstream. Although it works with `;` and `,` movement repeat, unfortunately it probably wouldn't work with complex combinations (e.g. `dfe` and `.` repeat wouldn't work).

Instead, we can make `builtin_f_expr`, `builtin_F_expr`, ... just returns `"f"`, `"F"`, `"t"`, `"T"` character so you can use it to map with `{expr = true}`. See the implementation below. This will almost likely resemble the upstream behaviour in most of the situations.

```lua
M.builtin_f_expr = function()
  M.last_move = {
    func = "f",
    opts = { forward = true },
    additional_args = {},
  }
  return "f"
end
```

Before returning the string, note that it also maps the last movement so the `;` and `,` mapping can handle it as well.

Important: users must change their keybindings to include `{ expr = true }`. I updated the README.md as follows:

```lua
-- Optionally, make builtin f, F, t, T also repeatable with ; and ,
vim.keymap.set({ "n", "x", "o" }, "f", ts_repeat_move.builtin_f_expr, { expr = true })
vim.keymap.set({ "n", "x", "o" }, "F", ts_repeat_move.builtin_F_expr, { expr = true })
vim.keymap.set({ "n", "x", "o" }, "t", ts_repeat_move.builtin_t_expr, { expr = true })
vim.keymap.set({ "n", "x", "o" }, "T", ts_repeat_move.builtin_T_expr, { expr = true })
```

The original functions `builtin_f` etc. are left unchanged, except it will notify the deprecation warning once.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,10 +174,10 @@ vim.keymap.set({ "n", "x", "o" }, ",", ts_repeat_move.repeat_last_move_previous)
 -- vim.keymap.set({ "n", "x", "o" }, ",", ts_repeat_move.repeat_last_move_opposite)
 
 -- Optionally, make builtin f, F, t, T also repeatable with ; and ,
-vim.keymap.set({ "n", "x", "o" }, "f", ts_repeat_move.builtin_f)
-vim.keymap.set({ "n", "x", "o" }, "F", ts_repeat_move.builtin_F)
-vim.keymap.set({ "n", "x", "o" }, "t", ts_repeat_move.builtin_t)
-vim.keymap.set({ "n", "x", "o" }, "T", ts_repeat_move.builtin_T)
+vim.keymap.set({ "n", "x", "o" }, "f", ts_repeat_move.builtin_f, { expr = true })
+vim.keymap.set({ "n", "x", "o" }, "F", ts_repeat_move.builtin_F, { expr = true })
+vim.keymap.set({ "n", "x", "o" }, "t", ts_repeat_move.builtin_t, { expr = true })
+vim.keymap.set({ "n", "x", "o" }, "T", ts_repeat_move.builtin_T, { expr = true })
 ```
 
 You can even make a custom repeat behaviour.
--- a/README.md
+++ b/README.md
@@ -174,10 +174,10 @@ vim.keymap.set({ "n", "x", "o" }, ",", ts_repeat_move.repeat_last_move_previous)
 -- vim.keymap.set({ "n", "x", "o" }, ",", ts_repeat_move.repeat_last_move_opposite)
 
 -- Optionally, make builtin f, F, t, T also repeatable with ; and ,
-vim.keymap.set({ "n", "x", "o" }, "f", ts_repeat_move.builtin_f, { expr = true })
-vim.keymap.set({ "n", "x", "o" }, "F", ts_repeat_move.builtin_F, { expr = true })
-vim.keymap.set({ "n", "x", "o" }, "t", ts_repeat_move.builtin_t, { expr = true })
-vim.keymap.set({ "n", "x", "o" }, "T", ts_repeat_move.builtin_T, { expr = true })
+vim.keymap.set({ "n", "x", "o" }, "f", ts_repeat_move.builtin_f_expr, { expr = true })
+vim.keymap.set({ "n", "x", "o" }, "F", ts_repeat_move.builtin_F_expr, { expr = true })
+vim.keymap.set({ "n", "x", "o" }, "t", ts_repeat_move.builtin_t_expr, { expr = true })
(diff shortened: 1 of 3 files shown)


kiyoon:
In order to prevent from breaking, we should change the name to `builtin_f_expr` and keep the original functions?

sahinakkaya (approved):
I tested this and can confirm that it works as expected.

kiyoon:
@sahinakkaya Thanks for testing! Would it be possible to keep it in your config and try it for some time? Because the last thing we want to do is to ask all users to change their config only to see another issue that we have to roll back.

sahinakkaya:
Yeah, sure. I was already planning to keep it until this is merged.

kiyoon:
@ribru17 I've also been using this and had no problem so far. Do you think it's ready to merge?

clason:
If it applies cleanly to the `main` branch.

kiyoon:
@clason Yes it does, it's separated from the module structure.

I can make a separate PR for the main branch if it's needed. I don't know if it can be just rebased as main branch has a little bit of refactoring.

clason:
Rebasing would be preferred (also to keep `main` up-to-date with respect to query changes). This will require some manual conflict resolution, but hopefully nothing too involved.

kiyoon:
I can rebase and make the PR.

clason:
No, no need to make a PR -- just `git checkout main; git rebase master`, resolve any conflicts, then `git rebase --continue`. Lather, rinse, repeat, and then just `git push -f`.

PR #623, opened 2024-05-23
fix(haskell): tweak textobject queries

mrcjkb:
A follow-up to #613.


clason:
I trust that this is still needed and correct?

PR #626, opened 2024-05-26
ci(lint): use query parser from nvim-treesitter

ribru17:
Adapts the linter to use the query parser in the nvim-treesitter repo, rather than the bundled Neovim one. This prevents breaking parser changes from messing up the formatting lint.

See https://github.com/nvim-treesitter/nvim-treesitter/commit/d3acd105bd48b6529cb5dc02e73d96990862e196

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -34,16 +34,20 @@ jobs:
   format-queries:
     name: Lint queries
     runs-on: ubuntu-latest
+    env:
+      NVIM_TAG: stable
     steps:
       - uses: actions/checkout@v4
       - name: Prepare
         run: |
-          wget https://github.com/neovim/neovim/releases/download/nightly/nvim-linux64.tar.gz
-          tar -zxf nvim-linux64.tar.gz
-          sudo ln -s "$PWD"/nvim-linux64/bin/nvim /usr/local/bin
+          bash ./scripts/ci-install-ubuntu-latest.sh
           wget -P scripts/ https://raw.githubusercontent.com/nvim-treesitter/nvim-treesitter/main/scripts/format-queries.lua
+          mkdir -p ~/.local/share/nvim/site/pack/nvim-treesitter/start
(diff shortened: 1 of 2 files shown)


tests/repeatable_move/common.lua
```diff
@@ -11,7 +11,7 @@ function M.run_builtin_find_test(file, spec)
   assert.are.same(1, vim.fn.filereadable(file), string.format('File "%s" not readable', file))
 
   -- load reference file
-  vim.cmd(string.format("edit %s", file))
+  pcall(vim.cmd.edit, file)
```
kiyoon, line 14: Why `pcall` instead of `vim.cmd.edit(file)`? Won't it silence the error when loading fails?

ribru17: I'm not sure, I just added it to be overkill; I didn't think that the `vim.cmd.edit` form of the function would handle the error gracefully if `vim.cmd('edit..` didn't

kiyoon: But silencing the error is worse than failing, don't you think?

ribru17: True, so then maybe we need another different solution. But right now I think there are errors unrelated to the actual tests (caused by this function, *I think*) which is why some checks pass randomly and other times fail. I don't understand the test code well enough, but could it be the case that silencing this error will still allow the tests to run successfully? Seems like due to the nature of the error it just can't create a file because it already exists, so if we silence that we can just continue normally?

kiyoon: It's not creating a file, rather it is reading a file. The tests read sample files and try multiple keystrokes to see if they have desirable behaviour. For example, it tests if `dia`, `viad`, `cia` results in the same buffer. Note that there were issues where it fails with certain combination that I don't really use but some people found it annoying.

So, if tests failed randomly it could be that we didn't wait for the keystrokes to finish (I thought they were synchronous but maybe not).

PR #629, opened 2024-06-03
fix(odin): update `has-type` renamed to `kind-eq`

mehalter:
Treesitter predicate changed upstream to align with Helix: https://github.com/nvim-treesitter/nvim-treesitter/commit/a80fe081b4c5890980561e0de2458f64aaffbfc7


clason:
Thanks!

PR #634, opened 2024-06-18
docs: fix the position of the `include_surrounding_whitespace`

ofseed:
`include_surrounding_whitespace` should be inside the `select`.


ofseed:
Almost forgot there's another tiny fix for `main` too, @clason Could you please do it again?

PR #637, opened 2024-06-19
feat(c_sharp): add class.outer support for records, enums, interfaces

marcinjahn:
Extends support for recognizing class.outer for the following types:

- classes with empty declarations

```cs
public class MyClass {}
```

- structs with empty declarations

```cs
public struct MyStruct {}
```

- records

```cs
public record MyRecord1();
public record MyRecord2() {}
public record MyRecord3 {}
```

- interfaces

```cs
public IMyInterface {}
```

- enums

```cs
public enum MyEnum {}
```

Before these changes, only structs and classes (both needed to have at least 1 member) had been supported.

Note: Ideally, there should also be support for structs, interfaces, classes and enums without curly braces (e.g. `class MyClass;` or `class MyClass();`), but it seems that tree-sitter does not support that syntax yet, tree inspection shows errors.

### DIFF
--- a/queries/c_sharp/textobjects.scm
+++ b/queries/c_sharp/textobjects.scm
@@ -9,6 +9,14 @@
     "}"
     (#make-range! "class.inner" @_start @_end))) @class.outer
 
+(class_declaration
+  body: (declaration_list
+    .
+    "{"
+    .
+    "}"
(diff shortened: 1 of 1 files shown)


marcinjahn:
Any love for C#?

PR #645, opened 2024-07-03
Update to support the complete re-writing of the R parser

PMassicotte:
The R grammar (https://github.com/r-lib/tree-sitter-r) was recently rewrote, which introduced breaking changes in `nvim-treesitter`. See https://github.com/nvim-treesitter/nvim-treesitter/commit/5373c90c55a33408604d17713edaa9560306b83d.

These breaking changes were addressed in these two commits:

1. https://github.com/nvim-treesitter/nvim-treesitter/commit/3e535e826da265748b875b18dbfb361bd1a1701c
2. https://github.com/nvim-treesitter/nvim-treesitter/commit/f0e3b5c5fe38d0012c63368db90017fef87c85a2

However, the changes in `tree-sitter-r` also broke `nvim-treesitter-textobjets` because node names were change (https://github.com/nvim-treesitter/nvim-treesitter/issues/6849):

```
Error detected while processing BufNewFile Autocommands for "*":                                                                                                                                                    
Error executing lua callback: ...bob/nightly/nvim-linux64/share/nvim/runtime/filetype.lua:36: BufNewFile Autocommands for "*"..FileType Autocommands for "*": Vim(append):Error executing lua callback: ...-linux64/
share/nvim/runtime/lua/vim/treesitter/query.lua:252: Query error at 12:2. Invalid node type "if":                                                                                                                   
(if                                                                                                                                                                                                                 
 ^                                                                                                                                                                                                                  
                                                                                                                                                                                                                    
stack traceback:                                                                                                                                                                                                    
        [C]: in function '_ts_parse_query'                                                                                                                                                                          
        ...-linux64/share/nvim/runtime/lua/vim/treesitter/query.lua:252: in function 'fn'                                                                                                                           
        ...vim-linux64/share/nvim/runtime/lua/vim/func/_memoize.lua:58: in function 'fn'                                                                                                                            
        ...vim-linux64/share/nvim/runtime/lua/vim/func/_memoize.lua:58: in function 'get_query'                                                                                                                     
        ...r-textobjects/lua/nvim-treesitter/textobjects/shared.lua:99: in function 'available_textobjects'                                                                                                         
        ...r-textobjects/lua/nvim-treesitter/textobjects/select.lua:173: in function 'attach'                                                                                                                       
        ...vim/lazy/nvim-treesitter/lua/nvim-treesitter/configs.lua:509: in function 'attach_module'                                                                                                                
        ...vim/lazy/nvim-treesitter/lua/nvim-treesitter/configs.lua:532: in function 'reattach_module'                                                                                                              
        ...vim/lazy/nvim-treesitter/lua/nvim-treesitter/configs.lua:133: in function <...vim/lazy/nvim-treesitter/lua/nvim-treesitter/configs.lua:132>                                                              
        [C]: in function 'nvim_cmd'                                                                                                                                                                                 
        ...bob/nightly/nvim-linux64/share/nvim/runtime/filetype.lua:36: in function <...bob/nightly/nvim-linux64/share/nvim/runtime/filetype.lua:35>                                                                
        [C]: in function 'pcall'                                                                                                                                                                                    
        vim/shared.lua: in function <vim/shared.lua:0>                                                                                                                                                              
        [C]: in function '_with'                                                                                                                                                                                    
        ...bob/nightly/nvim-linux64/share/nvim/runtime/filetype.lua:35: in function <...bob/nightly/nvim-linux64/share/nvim/runtime/filetype.lua:10>                                                                
stack traceback:                                                                                                                                                                                                    
        [C]: in function '_with'                                                                                                                                                                                    
        ...bob/nightly/nvim-linux64/share/nvim/runtime/filetype.lua:35: in function <...bob/nightly/nvim-linux64/share/nvim/runtime/filetype.lua:10>   
```

This PR is my attempt to fix these issues. I am by no mean a `treesitter` expert, but I have tested it and it seems to be working fine.

### DIFF
--- a/queries/r/textobjects.scm
+++ b/queries/r/textobjects.scm
@@ -74,19 +74,9 @@
 (float) @number.inner
 
 ; assignment
-("<-"
-  name: (_) @assignment.lhs
-  value: (_) @assignment.inner @assignment.rhs) @assignment.outer
-
-("<-"
-  name: (_) @assignment.inner)
-
-("->"
-  value: (_) @assignment.inner @assignment.lhs
-  name: (_) @assignment.rhs) @assignment.outer
-
-("->"
-  name: (_) @assignment.inner)
+(binary_operator
+  lhs: (_) @assignment.inner @assignment.lhs
+  rhs: (_) @assignment.rhs) @assignment.outer
 
 ((comma)
   name: (_) @assignment.lhs
@@ -94,17 +84,6 @@
 
 ((comma)
   name: (_) @assignment.inner)
-
-("<<-"
-  name: (_) @assignment.lhs
-  value: (_) @assignment.inner @assignment.rhs) @assignment.outer
-
-("<<-"
-  name: (_) @assignment.inner)
-
-("->>"
-  value: (_) @assignment.inner @assignment.lhs
-  name: (_) @assignment.rhs) @assignment.outer
 
 (binary_operator)
-  name: (_) @assignment.inner
+   name: (_) @assignment.inner
--- a/queries/r/textobjects.scm
+++ b/queries/r/textobjects.scm
@@ -77,13 +77,3 @@
 (binary_operator
   lhs: (_) @assignment.inner @assignment.lhs
   rhs: (_) @assignment.rhs) @assignment.outer
-
-((comma)
-  name: (_) @assignment.lhs
-  value: (_) @assignment.inner @assignment.rhs) @assignment.outer
-
-((comma)
-  name: (_) @assignment.inner)
-
-(binary_operator)
-   name: (_) @assignment.inner
--- a/queries/r/textobjects.scm
+++ b/queries/r/textobjects.scm
@@ -9,68 +9,56 @@
 (comment) @comment.outer
 
 ; conditional
-(if
+(if_statement
   condition: (_)? @conditional.inner) @conditional.outer
 
 ; function
-[
-  (function_definition)
-  (lambda_function)
-] @function.outer
-
 (function_definition
   [
     (call)
-    (binary)
-    (brace_list)
-  ] @function.inner) @function.outer
-
-(lambda_function
-  [
-    (call)
-    (binary)
-    (brace_list)
+    (binary_operator)
+    (braced_expression)
   ] @function.inner) @function.outer
 
 ; loop
 [
-  (while)
-  (repeat)
-  (for)
+  (while_statement)
+  (repeat_statement)
+  (for_statement)
 ] @loop.outer
 
-(while
+(while_statement
   body: (_) @loop.inner)
 
-(repeat
+(repeat_statement
   body: (_) @loop.inner)
 
-(for
+(for_statement
   body: (_) @loop.inner)
 
 ; statement
-(brace_list
+(braced_expression
   (_) @statement.outer)
 
 (program
   (_) @statement.outer)
 
 ; parameter
-((formal_parameters
-  "," @_start
+((parameters
+  (comma) @_start
   .
   (_) @parameter.inner)
   (#make-range! "parameter.outer" @_start @parameter.inner))
 
-((formal_parameters
+((parameters
   .
   (_) @parameter.inner
   .
-  ","? @_end)
+  (comma)? @_end)
   (#make-range! "parameter.outer" @parameter.inner @_end))
 
 ((arguments
-  "," @_start
+  (comma)? @_start
   .
   (_) @parameter.inner)
   (#make-range! "parameter.outer" @_start @parameter.inner))
@@ -79,44 +67,44 @@
   .
   (_) @parameter.inner
   .
-  ","? @_end)
+  (comma)? @_end)
   (#make-range! "parameter.outer" @parameter.inner @_end))
 
 ; number
 (float) @number.inner
 
 ; assignment
-(left_assignment
+("<-"
   name: (_) @assignment.lhs
   value: (_) @assignment.inner @assignment.rhs) @assignment.outer
 
-(left_assignment
+("<-"
   name: (_) @assignment.inner)
 
-(right_assignment
+("->"
   value: (_) @assignment.inner @assignment.lhs
   name: (_) @assignment.rhs) @assignment.outer
 
-(right_assignment
+("->"
   name: (_) @assignment.inner)
 
-(equals_assignment
+((comma)
   name: (_) @assignment.lhs
   value: (_) @assignment.inner @assignment.rhs) @assignment.outer
 
-(equals_assignment
+((comma)
   name: (_) @assignment.inner)
 
-(super_assignment
+("<<-"
   name: (_) @assignment.lhs
   value: (_) @assignment.inner @assignment.rhs) @assignment.outer
 
-(super_assignment
+("<<-"
   name: (_) @assignment.inner)
 
-(super_right_assignment
+("->>"
   value: (_) @assignment.inner @assignment.lhs
   name: (_) @assignment.rhs) @assignment.outer
 
-(super_right_assignment
-  name: (_) @assignment.inner)
+(binary_operator)
+  name: (_) @assignment.inner


milanglacier:
I tried this patch, I notice there is an incorrect behavior:

In the following example

```R
for (i in 1:20) {
    if (i > 20) {
        p<cursorPosition>rint(i)
    }
}

```

select the `@inner.condition` should select `print(i)`, but instead it selects `i > 20`

PMassicotte:
Thank you very much!

milanglacier:
Any updates on things that are needed to do to merge this PR?  Currently R users have to completely disable this plugin or rollback to an older version in able to write R in Neovim, otherwise they can not even save the file.

queries/r/textobjects.scm
```diff
@@ -79,44 +67,44 @@
...
 
-(super_right_assignment
+("->>"
   value: (_) @assignment.inner @assignment.lhs
   name: (_) @assignment.rhs) @assignment.outer
 
-(super_right_assignment
-  name: (_) @assignment.inner)
+(binary_operator)
+  name: (_) @assignment.inner
```
milanglacier, line 110: should this be

```lisp
(binary_operator
   name: (_) @assignment.inner)
```

? I am not super familiar with treesitter query grammer. Forgive me if I am actually wrong.

queries/r/textobjects.scm
```diff
@@ -9,68 +9,56 @@
...
   .
   (_) @parameter.inner
   .
-  ","? @_end)
+  (comma)? @_end)
   (#make-range! "parameter.outer" @parameter.inner @_end))
 
 ((arguments
-  "," @_start
+  (comma)? @_start
```
milanglacier, line 61: Is the `?`  needed at here?

queries/r/textobjects.scm
PMassicotte: I think this is ok like this. If I try your suggestion, I have an error.

queries/r/textobjects.scm
PMassicotte: I think it if for case like 

```r
plot(1, )
```

queries/r/textobjects.scm
milanglacier: > I think this is ok like this. If I try your suggestion, I have an error.

Yes, tried and see the error. Just ignore my suggestion.

milanglacier:
The following patch works for me on `main branch`:

```diff
diff --git a/plugin/query_predicates.lua b/plugin/query_predicates.lua
index b92f43d..9ba6d93 100644
--- a/plugin/query_predicates.lua
+++ b/plugin/query_predicates.lua
@@ -10,7 +10,9 @@ ts.query.add_directive("make-range!", function(match, pattern, buf, predicate, m
   local start_nodes = match[start_node_id]
   local end_nodes = match[end_node_id]
 
-  assert(#start_nodes == 1, "#make-range! does not support captures on multiple nodes")
+  if start_nodes then
+    assert(#start_nodes == 1, "#make-range! does not support captures on multiple nodes")
+  end
   if end_nodes then
     assert(#end_nodes == 1, "#make-range! does not support captures on multiple nodes")
   end
diff --git a/queries/r/textobjects.scm b/queries/r/textobjects.scm
index b14a37e..5cce11b 100644
--- a/queries/r/textobjects.scm
+++ b/queries/r/textobjects.scm
@@ -9,68 +9,56 @@
 (comment) @comment.outer
 
 ; conditional
-(if
+(if_statement
   condition: (_)? @conditional.inner) @conditional.outer
 
 ; function
-[
-  (function_definition)
-  (lambda_function)
-] @function.outer
-
 (function_definition
   [
     (call)
-    (binary)
-    (brace_list)
-  ] @function.inner) @function.outer
-
-(lambda_function
-  [
-    (call)
-    (binary)
-    (brace_list)
+    (binary_operator)
+    (braced_expression)
   ] @function.inner) @function.outer
 
 ; loop
 [
-  (while)
-  (repeat)
-  (for)
+  (while_statement)
+  (repeat_statement)
+  (for_statement)
 ] @loop.outer
 
-(while
+(while_statement
   body: (_) @loop.inner)
 
-(repeat
+(repeat_statement
   body: (_) @loop.inner)
 
-(for
+(for_statement
   body: (_) @loop.inner)
 
 ; statement
-(brace_list
+(braced_expression
   (_) @statement.outer)
 
 (program
   (_) @statement.outer)
 
 ; parameter
-((formal_parameters
-  "," @_start
+((parameters
+  (comma) @_start
   .
   (_) @parameter.inner)
   (#make-range! "parameter.outer" @_start @parameter.inner))
 
-((formal_parameters
+((parameters
   .
   (_) @parameter.inner
   .
-  ","? @_end)
+  (comma)? @_end)
   (#make-range! "parameter.outer" @parameter.inner @_end))
 
 ((arguments
-  "," @_start
+  (comma)? @_start
   .
   (_) @parameter.inner)
   (#make-range! "parameter.outer" @_start @parameter.inner))
@@ -79,44 +67,44 @@
   .
   (_) @parameter.inner
   .
-  ","? @_end)
+  (comma)? @_end)
   (#make-range! "parameter.outer" @parameter.inner @_end))
 
 ; number
 (float) @number.inner
 
 ; assignment
-(left_assignment
+("<-"
   name: (_) @assignment.lhs
   value: (_) @assignment.inner @assignment.rhs) @assignment.outer
 
-(left_assignment
+("<-"
   name: (_) @assignment.inner)
 
-(right_assignment
+("->"
   value: (_) @assignment.inner @assignment.lhs
   name: (_) @assignment.rhs) @assignment.outer
 
-(right_assignment
+("->"
   name: (_) @assignment.inner)
 
-(equals_assignment
+((comma)
   name: (_) @assignment.lhs
   value: (_) @assignment.inner @assignment.rhs) @assignment.outer
 
-(equals_assignment
+((comma)
   name: (_) @assignment.inner)
 
-(super_assignment
+("<<-"
   name: (_) @assignment.lhs
   value: (_) @assignment.inner @assignment.rhs) @assignment.outer
 
-(super_assignment
+("<<-"
   name: (_) @assignment.inner)
 
-(super_right_assignment
+("->>"
   value: (_) @assignment.inner @assignment.lhs
   name: (_) @assignment.rhs) @assignment.outer
 
-(super_right_assignment
-  name: (_) @assignment.inner)
+(binary_operator)
+  name: (_) @assignment.inner

```


This following patch is required:

```lua
-  assert(#start_nodes == 1, "#make-range! does not support captures on multiple nodes")
+  if start_nodes then
+    assert(#start_nodes == 1, "#make-range! does not support captures on multiple nodes")
+  end
```

Otherwise will get the following error:

```lua
E5108: Error executing lua: .../nvim-treesitter-textobjects/plugin/query_predicates.lua:13: attempt to get length of local 'start_nodes' (a nil value)
```

PMassicotte:
Anything that needs to be done on my side at this point?

milanglacier:
> Anything that needs to be done on my side at this point?

I think you don’t need to do anything?

 The `main` branch will supersede the `master` branch in the future (similar to what happens in the `nvim-treesitter` repo). Since the two branch are not compatible… Maybe we can wait until this PR is merged in the master branch and back ported, and then to see what is the next step in the `main` branch?

PMassicotte:
> Maybe we can wait until this PR is merged in the master branch and back ported, and then to see what is the next step in the `main` branch?

I think so. I am not sure if anything is missing for that PR to be merged.

Tal500:
Thanks! that fixes my issues

Fixes r-lib/tree-sitter-r#131
Fixes nvim-treesitter/nvim-treesitter#7001

clason:
Checks are failing, so that's an obvious thing that should be fixed.

clason:
And ignore the main branch, the master branch is still the default.

PMassicotte:
The failed check seems to be related to another grammar:

```
Checking tact textobjects
Check failed:
...bQDB/usr/share/nvim/runtime/lua/vim/treesitter/query.lua:252: Query error at 5:2. Invalid node type "static_function":
(static_function
```

clason:
Ugh. This repo is in worse shape than I thought.

@lucario387 can you take a look at the query format failure?

queries/r/textobjects.scm
```diff
@@ -79,44 +67,44 @@
...
   name: (_) @assignment.inner)
 
-(super_right_assignment
+("->>"
   value: (_) @assignment.inner @assignment.lhs
   name: (_) @assignment.rhs) @assignment.outer
 
-(super_right_assignment
-  name: (_) @assignment.inner)
+(binary_operator)
```
lucario387, line 109: Failure is indeed intended because there is a superfluous `)` here.

queries/r/textobjects.scm
lucario387, line 110:
```suggestion
(binary_operator
  name: (_) @assignment.inner)
```

lucario387:
> @lucario387 can you take a look at the query format failure?

Oops did not handle this :) I'll add a PR later

queries/r/textobjects.scm
```diff
@@ -79,44 +67,44 @@
...
-  ","? @_end)
+  (comma)? @_end)
   (#make-range! "parameter.outer" @parameter.inner @_end))
 
 ; number
 (float) @number.inner
 
 ; assignment
-(left_assignment
+("<-"
```
lucario387, line 77: Okay this changes the syntax of these queries to be no longer parent-child but rather sequential nodes. Is this intended? @PMassicotte

What I mean by the above is
Before, the query would look for `name/value` field that are direct children of `left_assignment`, but now it looks for `name/value` field that have the same direct parent as `<-`

queries/r/textobjects.scm
PMassicotte: This is in the current version, right? `super_right_assignment` does not exist anymore.

queries/r/textobjects.scm
lucario387: Then this is not exactly what you want then :). Please remove it if it's unnecessary

queries/r/textobjects.scm
PMassicotte: > Okay this changes the syntax of these queries to be no longer parent-child but rather sequential nodes. Is this intended? @PMassicotte

I am not a treesitter expert by no mean, but this is my understanding.

queries/r/textobjects.scm
```diff
@@ -79,44 +67,44 @@
...
   value: (_) @assignment.inner @assignment.rhs) @assignment.outer
 
-(super_assignment
+("<<-"
   name: (_) @assignment.inner)
 
-(super_right_assignment
+("->>"
   value: (_) @assignment.inner @assignment.lhs
   name: (_) @assignment.rhs) @assignment.outer
```
lucario387, line 107: Reading the grammar, I have yet to find a place where this pattern is matched. I think all of these `{left,super,right,equals}_assignment` can be replaced with

```suggestion
(binary_operator
  lhs: (_) @assignment.inner @assignment.lhs
  rhs: (_) @assignment.rhs) @assignment.outer
```

queries/r/textobjects.scm
clason: Yes, the format of the _current_ version is an incorrect s-expression. The way you've written it, the `name` is _outside_ the pattern for the `(binary_operator)` node, which makes no sense. (Compare with the paren placement in the old version.)

queries/r/textobjects.scm
lucario387: You still need to remove these 2 lines. They don't make sense.

queries/r/textobjects.scm
```diff
@@ -79,44 +67,23 @@
...
-  name: (_) @assignment.inner)
-
-(super_assignment
+((comma)
   name: (_) @assignment.lhs
   value: (_) @assignment.inner @assignment.rhs) @assignment.outer
 
-(super_assignment
+((comma)
   name: (_) @assignment.inner)
```
lucario387, line 86: In the parser, comma never goes with a `name` field, only `parameter`. Please remove this as well

lucario387:
Please remove the emojis from the commit message :). All three of them

PMassicotte:
> Please remove the emojis from the commit message :). All three of them

Done.

lucario387 (approved):
Tact issues should be fixed in a different PR

PMassicotte:
Thank you.

Tal500:
Thanks!

Issue #638, opened 2024-06-21
Swap @function.outer doesn't work from within decorated functions in Python


Hubro:
**Describe the bug**

Swapping `@function.outer` in Python doesn't work when inside a decorated function. Take the example:

```python
def test():
    print("This is a test")


@some_decorator
def test2():
    print("This is a test")


def test3():
    print("This is a test")
```

When in either of the un-decorated functions, swapping with the decorated function works just fine. When in the decorated function, nothing happens when you try to swap, unless the cursor is on the decorator line.

Given the queries:

```query
(function_definition
  body: (block)? @function.inner) @function.outer

(decorated_definition
  (function_definition)) @function.outer
```

My guess is that this happens because when inside the decorated function, the non-decorated `@function.outer` match is the closest match and is selected for swapping, but since it's inside a `(decorated_definition ...`, it doesn't have any siblings to swap with:

![image](https://github.com/nvim-treesitter/nvim-treesitter-textobjects/assets/597206/4ebccdd8-e53c-4a3e-a48a-b02fd0e8fb86)

So nothing happens.

**To Reproduce**

(See above)

I'm using these functions:

```lua
local ts_swap = require("nvim-treesitter.textobjects.swap")

ts_swap.swap_next("@function.outer")
ts_swap.swap_previous("@function.outer")
```

**Expected behavior**

If the current cursor position matches both a function_definition and a decorated_definition, the latter should have higher priority.

**Output of `:checkhealth nvim-treesitter`**

<details>
<code>
nvim-treesitter: require("nvim-treesitter.health").check()

Installation ~
- WARNING `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
- OK `node` found v20.11.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
- OK `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
  Version: gcc (GCC) 13.2.0
- OK Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

OS Info:
{
  machine = "x86_64",
  release = "6.9.3",
  sysname = "Linux",
  version = "#1-NixOS SMP PREEMPT_DYNAMIC Thu May 30 07:45:04 UTC 2024"
} ~

Parser/Features         H L F I J
  - bash                ✓ ✓ ✓ . ✓
[...]

PR #648, opened 2024-07-17
make @function.outer and @class.outer for python always use the decorated version

feakuru:
See #638 for context.

This is a minor adjustment to the Python queries that I find to produce a more useful behavior: I don't ever need to select a function without its decorators, because when I select a whole function, it's to move it, delete it or copy it in whole, and in all cases, I almost never need to leave the decorators out of my operation. If I do sometimes need to leave them out, I can manually adjust the selection later in a multitude of ways. I admit that this might be a personal preference issue, but I don't have reason to believe that my work differs from the average significantly in this instance.

In any case, please tell me if I need to adjust something else or otherwise, as I am not very experienced with treesitter yet.

upd: seeing the failing tests, will look into those later, feel free to post any advice for those


kiyoon:
I like this too. Can you make it the same for the class?

kiyoon:
But the tests are failing.. Do you know why? You don't have to care about the consistency tests because the functionality changed anyway

kiyoon:
Another problem is that if you just remove the textobject, then it won't grab when the function doesn't have decorators..

feakuru:
> Can you make it the same for the class?

i'll take a look at the class queries, but i will make no promises since i am not very experienced with this 😄

queries/python/textobjects.scm
```diff
@@ -1,5 +1,5 @@
 (function_definition
-  body: (block)? @function.inner) @function.outer
+  body: (block)? @function.inner)
```
ribru17, line 2: I think the `?` can be removed since there are no other captures

feakuru: i found a way to write out the only definition we need instead of two different ones, and it seems to work, so i left the `?` in. but i will say that i don't really understand why it is needed. is it for when a function has no body? is such a thing valid? or is this to handle the situation where the user has not finished typing the function?

feakuru:
> i'll take a look at the class queries

so i did take a look and found that the query for class could be adjusted in the same way. did just that, seems to pass the tests 

> it won't grab when the function doesn't have decorators..

now it seemingly works for both cases, i basically just marked the `decorated_definition` as optional with a `?` and voila 🎉 thanks for spotting this!

queries/python/textobjects.scm
ribru17: Thanks, in this case I think it's fine. I don't even think the `?` is needed because the grammar looks like it [always gives function definitions a body](https://github.com/tree-sitter/tree-sitter-python/blob/0dee05ef958ba2eae88d1e65f24b33cad70d4367/grammar.js#L409C1-L423C7) . But I think it's fine as is

kiyoon: There might be edge cases like python type stub files probably don't have a body etc.

kiyoon:
Although I haven't tested it, I like the approach. Interestingly the consistency test still pass, and I thought it would fail. Maybe I didn't add the `@function.outer` tests?

kiyoon (approved):

kiyoon:
> from what I understand from reading the consistency tests script, they only check a big `loss.py` example file which does not seem to contain any decorators

You're right! Thanks for looking into it. 

I'll merge this when the query problem gets fixed

kiyoon:
Can you rebase it to master?

feakuru:
> Can you rebase it to master?

done 🙃

PR #658, opened 2024-08-02
feat(latex): improved block/frame and sections

simonmandlik:
This PR fixes and improves LaTeX block and frame queries:

- arbitrary length blocks/frames are supported
- `@frame` works now
- `@class.inner` supported


simonmandlik:
@kiyoon Could you please try rerunning the tests here?

Issue #649, opened 2024-07-20
[main] moving between functions does not work


przepompownia:
`require('nvim-treesitter-textobjects.move').goto_next_start('@function.outer', 'textobjects')` seems to not work. 

With `init.lua`
```lua
local thisInitFile = debug.getinfo(1).source:match('@?(.*)')
local cwd = vim.fs.dirname(thisInitFile)
local appname = vim.env.NVIM_APPNAME or 'nvim'

vim.env.XDG_CONFIG_HOME = cwd
vim.env.XDG_DATA_HOME = vim.fs.joinpath(cwd, '.xdg', 'data')
vim.env.XDG_STATE_HOME = vim.fs.joinpath(cwd, '.xdg', 'state')
vim.env.XDG_CACHE_HOME = vim.fs.joinpath(cwd, '.xdg', 'cache')
vim.fn.mkdir(vim.fs.joinpath(vim.env.XDG_CACHE_HOME, appname), 'p')
local stdPathConfig = vim.fn.stdpath('config')

vim.opt.runtimepath:prepend(stdPathConfig)
vim.opt.packpath:prepend(stdPathConfig)

local function gitClone(url, installPath, branch)
  if vim.fn.isdirectory(installPath) ~= 0 then
    return
  end

  local command = {'git', 'clone', '--', url, installPath}
  if branch then
    table.insert(command, 3, '--branch')
    table.insert(command, 4, branch)
  end
  local sysObj = vim.system(command, {}):wait()
  if sysObj.code ~= 0 then
    error(sysObj.stderr)
  end
  vim.notify(sysObj.stdout)
  vim.notify(sysObj.stderr, vim.log.levels.WARN)
end

local pluginsPath = 'nvim/pack/plugins/opt'
vim.fn.mkdir(pluginsPath, 'p')
pluginsPath = vim.uv.fs_realpath(pluginsPath)

--- @type table<string, {url:string, branch: string?}>
local plugins = {
  ['nvim-treesitter'] = {url = 'https://github.com/nvim-treesitter/nvim-treesitter', branch = 'main'},
  ['nvim-treesitter-textobjects'] = {url = 'https://github.com/nvim-treesitter/nvim-treesitter-textobjects', branch = 'main'}
}

for name, repo in pairs(plugins) do
  local installPath = vim.fs.joinpath(pluginsPath, name)
  gitClone(repo.url, installPath, repo.branch)
[...]

PR #659, opened 2024-08-03
fix: avoid pass nonexistent field

ofseed:
This line tries to access `range` on `TSTextObjects.Range`, which does not have this field, so the passed value will be `nil`. This bug makes the `move` feature usable.


ofseed:
Close https://github.com/nvim-treesitter/nvim-treesitter-textobjects/issues/649

PR #662, opened 2024-08-04
ci(tests): make tests actually run on `main`

clason:
Also prevent running duplicate tests and fix typos.

### DIFF
--- a/.github/workflows/check_query_files.yml
+++ b/.github/workflows/check_query_files.yml
@@ -1,21 +1,18 @@
 name: Check loading of syntax files
 
 on:
-  push:
   pull_request:
-  schedule:
-    - cron: "0 0 * * *" # every day at midnight
+    branches:
+      - "main"
+  workflow_dispatch:
 
 jobs:
   luacheck:
     name: Check Query Files
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
-
-      - name: Install tree-sitter CLI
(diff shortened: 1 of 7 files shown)


clason:
Not sure why the Python test is failing here. @kiyoon @ribru17 @theleop any idea?

kiyoon:
It seems like it's unable to locate the textobjects queries

clason:
Yes, but why?

kiyoon:
Honestly I don't know unless I try to set it up on my own again.. What were you trying to do in the PR exactly? I think the tests were alright it had no problem for a while

clason:
> What were you trying to do in the PR exactly? I think the tests were alright it had no problem for a while

No, the tests on `main` were completely broken. I'm trying to fix them.

clason:
OK, something is wrong with the `run_tests.sh` script; there's an unspecified error and then it just hangs. That's beyond my expertise, though.

clason:
OK, things are working as far as I can see. I suspect the test failures (which are some, not all!)  are actually real regressions.

dundargoc (approved):

clason:
Regressions will have to be fixed in a follow-up PR.

clason:
First test failure is due to selection mode not being able to be specified anymore (see https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/ca93cb2c34b67ab22d01976fc90bc95627a3317f/scripts/minimal_init.lua#L69-L71)

PR #663, opened 2024-08-04
fix: avoid unnecessary closures

ofseed:
The actual way to use it is `require("nvim-treesitter-textobjects.swap").swap_next("@parameter.inner")()`, not as the [what the readme said](https://github.com/nvim-treesitter/nvim-treesitter-textobjects/tree/main#text-objects-swap). The closure is unnecessary.


clason:
I suspect the closure was needed for the way mappings were automatically set up by the module system -- which we're removing anyway. If you've tested this and it works, this is fine with me.

PR #667, opened 2024-08-06
refactor(shared): move out Range class, memoize, code shuffle and type fixes

clason:
- **refactor: move main module to init.lua**
- **refactor(shared): move out Range class**
- **refactor(shared): inline and memoize match preparation**
- **refactor(shared): more memoize, type fixes**
- **refactor: consistent local api**
- **refactor(shared)!: drop support for specifying textobjects via regex**
- **refactor(shared): move {next,previous}_textobject to swap**

Includes smaller code reshuffles and todos.


clason:
@ribru17 sorry, this again conflicts (cosmetically) with your PR. I know this is asking much, but if you could looking into caching/memoizing while do it, that would be appreciated.

clason:
Cursory tests seem to work fine, but I'd appreciate someone who actually uses this plugin to check for regressions.

clason:
(The idea behind the refactors here and mentioned in the todos is to structure the modules so people will only have to pay for what they use, when they use it. It will also make it easier to work on each module without worrying about breaking other modules.)

clason:
Scope textobject already fails on `main`; looks like the range class (or its call sites) needs a lot more `nil` guards.

kiyoon (approved):

kiyoon:
@clason maybe the "lua match" matching of query names for the move feature should be removed as well. Maybe not many people use it and it just confuse them? Or make it as an option so that it is more clear that we opt-in.

But I would like to keep the "gradual movement" feature where you can set multiple query names. It is sometimes useful.

clason:
Yes, I agree, globs are overkill (but lists are useful). Not sure I want to remove them in this PR, though.

clason:
And thoughts on further splitting up shared?

* `shared.get_adjacent` and `{next,previous}_textobject` are only used in `swap`
* `shared.check_support` and `shared.available_textobjects` can be moved to config to more cheaply check when setting up keymaps

(Alternative: the motion mappings themselves check and return false if not supported; the wrapper can then check the return value and call the original mapping instead: `move(...) or `]m`.)

clason:
I can make those as a separate PR, though. Don't want this one to become unreviewable. (I'll see about fixing the locals failure in a separate PR, too.)

clason:
OK, I'm done with my ~~murder~~ refactoring spree for now. The next step is to rework the config (less metatable magic, more explicit) and the helpers for (buffer local) mapping setup, but that requires thought.

But before that, we should fix the many, _many_, regressions from master... (including getting the tests to pass again).

clason:
Two test failures remain:
* first failure is repeatable `f` move -- apparently doesn't work properly with count? -- which I can't reproduce locally;
* second failure is a feature regression: you can no longer specify selection type for a textobject (or at least I don't see how).

clason:
Tests pass now; will merge if nobody complains @kiyoon @ribru17

ribru17:
Thanks a lot for this, can review now

ribru17 (approved):

ribru17:
> * `shared.check_support` and `shared.available_textobjects` can be moved to config to more cheaply check when setting up keymaps (this would require the mapped operations to silently fail on unsupported textobjects, which I think is the better UX anyway and is how every other textobject behaves).
> 
> (Alternative: the motion mappings themselves check and return false if not supported; the wrapper can then check the return value and call the original mapping instead: `move(...) or ]m`.)

Agree with all this

> @ribru17 sorry, this again conflicts (cosmetically) with your PR. Mostly it's just shuffling things around, though, so your logic changes should still apply pretty cleanly.

No worries, I have accepted that that PR will come with lots of rebasing haha

clason:
all yours, @ribru17 !

ribru17:
lord have mercy on my soul...

PR #671, opened 2024-08-19
refactor(repeatable_move): extract the `MoveFunction` interface

ofseed:
`repeatable_move` is out of the plugin's scope, and I agree with simplifying it. But even if the `make_repeatable_move_pair` was removed, the current API is good enough for users to define their repeatable move, it only needs some refactoring and documentation. Also contains some tiny improvements for type annotations.

`nvim-next` does not support the `main` branch, so there's no replacement for users now, but since its author knows this so it might be implemented in the future, so I kept the reference to `nvim-next`. see https://github.com/ghostbuster91/nvim-next/issues/22.

Edit:
I changed this PR to `feat` because the `MoveFunction` interface should be guaranteed.

### DIFF
--- a/lua/nvim-treesitter-textobjects/move.lua
+++ b/lua/nvim-treesitter-textobjects/move.lua
@@ -36,11 +36,13 @@ end
 local M = {}
 
 ---@param opts TSTextObjects.MoveOpts
-local function move(opts)
-  local query_group = opts.query_group or "textobjects"
-  local query_strings = type(opts.query_strings) == "string" and { opts.query_strings } or opts.query_strings
+---@param query_strings string[]|string
+---@param query_group? string
+local function move(opts, query_strings, query_group)
+  query_group = query_group or "textobjects"
+  query_strings = type(query_strings) == "string" and { query_strings } or query_strings
 
-  local winid = opts.winid or api.nvim_get_current_win()
+  local winid = api.nvim_get_current_win()
   local bufnr = api.nvim_win_get_buf(winid)
 
   local forward = opts.forward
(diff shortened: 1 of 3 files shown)


kiyoon:
This is intended to be dropped out from the README to limit the support and make it easier to maintain. Now it's not intended to be used with other plugins.

ofseed:
Currently, no plugin can replace this functionality. `nvim-next` is based on the previous module system, and more importantly, I think the current implementation is already simple enough to be easily maintained. 

The purpose of this PR is merely to show users how to use an already exposed API.

I think the better approach would be either not to provide this functionality at all (which would be the easiest to maintain) or to expose the internally used method in the most maintainable way (which is the current situation). After all, we have already implemented this feature. It seems odd to keep a feature that we originally intended users to implement with other plugins.

Or do you mean that guaranteeing the `make_repeatable_pair` API would also harm maintenance?

kiyoon:
I honestly agree with you that is has been stable and simple enough to maintain. However, there was another voice that any dependency with different plugins should be completely removed so we can focus more on the treesitter features. See the discussion at #666.

I think I would personally implement my own version of repeatable movement plugin (which is basically a copy of the original implementation) once I transit to the main branch.

ofseed:
Except for adding an explanation on README, this PR also did some refactoring, which made the internal API more flexible and avoided some diagnostic errors. Could still consider merging it even if additions to the README are not needed (I can or you can delete it).

lua/nvim-treesitter-textobjects/move.lua
```diff
@@ -36,11 +36,15 @@ end
...
+---@param query_strings string[]|string
+---@param query_group? string
+local function move(opts, query_strings, query_group)
+  query_group = query_group or "textobjects"
+  if type(query_strings) == "string" then
+    query_strings = { query_strings }
+  end
 
-  local winid = opts.winid or api.nvim_get_current_win()
+  local winid = api.nvim_get_current_win()
```
kiyoon, line 47: why was `opts.winid` removed?

kiyoon:
Any reasoning as to why `query_strings` and `query_group` are removed from `MoveOpts` and instead placed as the `move()` function parameters?

lua/nvim-treesitter-textobjects/move.lua
ofseed: The exposed API never passes `winid`, so it will always be nil and `get_current_win()`.

ofseed:
> Any reasoning as to why `query_strings` and `query_group` are removed from `MoveOpts` and instead placed as the `move()` function parameters?

To keep the `repetable_move` module independent. The original motive is to make it easier to expose API, but I think it should also be reasonable even if we do not expose it.

lua/nvim-treesitter-textobjects/move.lua
kiyoon: I thought it was used somewhere, but honestly everyone uses textobjects and movements within the current window. So I guess it's not a problem to remove them or we should add winid everywhere to make it consistent.

ofseed:
If we decide to remove the `repeatable_move` module one day, we can only remove the `opts` parameter. It should be a good refactor from both perspectives.

kiyoon:
I think we won't remove the `repeatable_move` itself, but we're not exposing the API. So, users may use it with this plugin but they're not supposed to use it with other plugins. Of course, they can hack it, but that won't be guaranteed to be stable and is out of scope for maintenance. 

Given that in mind, you can remove the README and further refactor if you wish. I think what you did is great and I appreciate it.

lua/nvim-treesitter-textobjects/move.lua
ofseed: In our current exposed API, the second parameter is already optional. If we add a third parameter as an optional one, when we want to pass the third parameter, we would at least have to pass `nil` as the second parameter. So, in my opinion, unless it's particularly useful, it's better not to add it.

kiyoon: I think that's why we had it in the `opts`, no?

ofseed:
Okay, I deleted explanations on README and squashed these commits into three to avoid being overly finegrained.

lua/nvim-treesitter-textobjects/move.lua
ofseed: In most APIs with optional parameters in neovim, yes. However the API in this plugin does not have `opts`, it has two fixed parameters. The function with `opts` parameter is internally used.

PR #672, opened 2024-08-19
refactor(range): remove `range.lua`

ofseed:
Got some time to refactor this plugin to remove unneeded abstracts.


kiyoon:
If I understood correctly, it's using the neovim's Range4 and Range6 types instead of defining a new type ourselves?

ofseed:
Yes, this PR completely removes the custom `Range` in favor of using the types from upstream. I don't see this as the final form of this plugin but rather as an intermediate step in a series of refactorings that gradually take advantage of the new upstream APIs. Explicitly using `Range4" and "Range6` should be temporary, but it should make it easier to transition internal functions to use the upstream implementation.

ofseed:
I see @clason added this as a TODO, and I hope I understood it correctly.
https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/5fa1b337af6f3ff3450794da61e6623bb92a3b49/lua/nvim-treesitter-textobjects/shared.lua#L2

kiyoon:
I think this is a good move, but I need @clason to confirm this.

lua/nvim-treesitter-textobjects/shared.lua
```diff
@@ -82,35 +80,31 @@ local get_query_matches = memoize(function(bufnr, query_group, root, root_lang)
...
-          Range:new(
-            metadata.range[1],
-            metadata.range[2],
-            metadata.range[3],
-            metadata.range[4],
-            metadata.range[5],
-            metadata.range[6],
-            "-1",
-            "-1",
-            bufnr
```
clason, line 111: This undoes a change I made deliberately to fix failing tests. Can you comment on why you did that?

clason:
Yes, that was exactly what I had in mind. As long as tests (and basic functionality in manual testing) still work, I'm fine with it.

As you say there's a bit more work that can be done; I'm not yet happy with
* the lsp conversion (but then I'm not happy with tying this to LSP, period);
* the still exported range4--range6 conversion function, but that is probably necessary until upstream exports more functionality.

Have you looked at `vim.treesitter._range`? Would that help abstracting from the range4/range6 difference? Ideally, we could stick to one throughout _our_ code.

ofseed:
> * the lsp conversion (but then I'm not happy with tying this to LSP, period);

Its purpose is merely to allow us to perform more intuitive text operations using `vim.lsp.util.apply_text_edits`. This function should only be used within `swap.lua`. Could be easy to delete if we implement ours or drop `swap.lua`.

> * the still exported range4--range6 conversion function, but that is probably necessary until upstream exports more functionality.
>
> Have you looked at `vim.treesitter._range`? Would that help abstracting from the range4/range6 difference? Ideally, we could stick to one throughout _our_ code.

There's a `unpack4` function that could convert any `Range`(an alias of `Range2|Range4|Range6`) to `Range4`, so it will not be a big deal if we copy/import it.

Furthermore, by using conversion functions nearby, we could make all the functions receive any `Range`, because they could use `_range.unpack4` to get a `Range4` or use `_range.add_bytes` to get `Range6`, and as a result, every function's parameter could be any subtype of `Range`. Indeed treating vim.treesitter._range as a public API could also be a good approach because otherwise, we will copy a lot. There are some other useful functions such as `vim. treesitter._range.contains` design well, could replace two functions in our current codebase.

After all, I am not the decision-maker. Our final goal should be a plugin that only depends on `vim.treesitter`'s public API, so I'm not sure if introducing these functions in `vim.treesitter._range` will lead us down the wrong path, I just took the necessary steps.

lua/nvim-treesitter-textobjects/shared.lua
ofseed: The last three parameters are for `id`, `parent_id`, and `bufnr`, all of these I removed because these fields only exist on the former `TSTextObject.Range`(In other words, we must remove them). IIRC except for checking whether the two ranges are equivalent, The only usages of these fields in `swap.lua`, interestingly these usages are in a true condition, e.g. `current_range.id == current_range.id`, so I delete them directly. There might be a regression here, but the regression has already happened before this PR.

lua/nvim-treesitter-textobjects/shared.lua
```diff
@@ -220,47 +214,80 @@ function M.find_best_range(bufnr, capture_string, query_group, filter_predicate,
   return best
 end
 
+---@param range Range4
+---@param row integer
+---@param col integer
+---@return boolean
+local function is_in_range(range, row, col)
```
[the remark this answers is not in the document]
ofseed: `vim.treesitter._range.contains` should be a better choice because it could replace `contains` beneath this function. I kept them only because they are originally methods of `TSTextObject.Range`.

clason:
As I said, those were general remarks about the inherited tech debt, not complaints about the changes in this PR (which I am fine with).

> Its purpose is merely to allow us to perform more intuitive text operations using vim.lsp.util.apply_text_edits. This function should only be used within swap.lua. Could be easy to delete if we implement ours or drop swap.lua.

Yes, I know. I am not convinced using `apply_text_edits` for this is a good design. (I understand why it was done back when these modules were first written, but that doesn't mean we should keep what was clearly an early kludge.) Again, out of scope of this PR; isolating the functionality to this module is a good step in the right direction.

> Indeed treating vim.treesitter._range as a public API could also be a good approach because otherwise, we will copy a lot.

Thank you for checking; that is valuable information. I'll see about the possibility of stabilizing that. 

In general, we should always prefer the higher-level `vim.treesitter` API (e.g., see comment above; I suspect there are more cases where custom code could be replaced by API functions added since the code was written), but if there's a significant chunk of code left that can be replaced by `vim.treesitter._range`, I would be fine to use make use of it despite its "pre-release status".

(Rationale: This plugin is unreleased, too, and maintenance is tightly coupled to Neovim itself so breaking changes can be adapted to quickly.)

Again, that can -- and should -- be a follow-up PR.

lua/nvim-treesitter-textobjects/shared.lua
clason: Well, there was a regression that I fixed by re-adding them. Can you confirm that swapping, e.g., function arguments in Lua still works after this change?

lua/nvim-treesitter-textobjects/shared.lua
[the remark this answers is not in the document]
clason: Right, it's not related to any node, sorry about the noise. As far as I can see, `contains` was only used in one place (`shared.textobject_at_points`), so maybe we can simply do without that abstraction?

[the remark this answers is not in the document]
ofseed: Okay...I found it could not work, but only because of a typo not related to this, I've fixed it and force-pushed.

https://github.com/nvim-treesitter/nvim-treesitter-textobjects/compare/734f9acc449ea3819a80c1ca32140c9a05bab53b..af0aad9c5498a1a7cbf7b2b5644ab3fad9b5c44e

lua/nvim-treesitter-textobjects/shared.lua
ofseed: Okay...I found it could not work, but only because of a typo not related to this, I've fixed it and force-pushed.

https://github.com/nvim-treesitter/nvim-treesitter-textobjects/compare/734f9acc449ea3819a80c1ca32140c9a05bab53b..af0aad9c5498a1a7cbf7b2b5644ab3fad9b5c44e

clason: Yeah, I'm afraid the test suite is not comprehensive enough; I had to do a lot of manual testing to find and fix regressions in my last big refactor PR. (Another thing that is not sufficiently covered and needs to be tested manually is using locals for selection; I use the `minimal_init` config and just go through the list of blocks, testing one command from each.)

lua/nvim-treesitter-textobjects/shared.lua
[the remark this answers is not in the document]
ofseed: Another reason  I did not inline them is that they are for `Range4`, but the variables we have is `Range6` in the scope these functions be called. Also another reason for using `contains`, it handles all subtypes of `Range`. I think we could do it after making the decision of whether to use functions from `vim.treesitter._range` or finally choose a `Range{n}` for all of our functions.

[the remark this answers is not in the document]
clason: How feasible is fixing an `n` for all function signatures?

ofseed:
In fact, I have already made such an attempt, though I haven't pushed it to this branch. What I found is that if we decide to use the functions from `_range`, almost every file will end up requiring it.

lua/nvim-treesitter-textobjects/shared.lua
```diff
@@ -447,4 +472,28 @@ function M.check_support(bufnr, query_group, queries)
   return true
 end
 
+M.range6_range4 = range6_range4
```
clason, line 475: Avoid unnecessary wrappers (makes it harder to use LSP for checking references); just expose the original function.

clason:
> In fact, I have already made such an attempt, though I haven't pushed it to this branch. What I found is that if we decide to use the functions from `_range`, almost every file will end up requiring it.

Well, that's fine (every file uses `vim.treesitter`, too) if it improves (reduces the amount of) code in those files.

But I would like to keep that move to a separate PR so we can evaluate it in isolation. (There's too many other -- welcome -- refactors in this PR.)

lua/nvim-treesitter-textobjects/shared.lua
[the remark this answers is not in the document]
ofseed: The information we get directly from the query is `Range6`. It contains a byte-level location for files and some functions need it, so if we chose one single type for all functions, it must be `Range6`. But some functions do not need it, doing so would make it more cumbersome.

We cannot convert `Rnage4` to `Range6` without knowing which source the range refers to, the bytes locations can not be calculated by only a rows and columns tuple.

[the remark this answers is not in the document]
ofseed: On the other hand, I'm not very familiar with the latest `vim.treesitter` API. I'm not sure if rewriting these functions to utilize the new API would change this situation. So, for now, I haven't done anything.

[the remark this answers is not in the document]
clason: Well, the question is whether using `Range6` consistently (and ignoring bytes in some places) is more or less cumbersome than having to explicitly convert them all over the place ;)

lua/nvim-treesitter-textobjects/select.lua
```diff
@@ -78,7 +78,8 @@ end
 ---@return integer? row
 ---@return integer? col
 local function next_position(bufnr, row, col, forward)
-  local max_col = #get_line(bufnr, row)
+  local result = get_line(bufnr, row)
+  local max_col = #result
```
clason, line 82: Nit: If the `result` is not used by itself, I prefer the previous version.

lua/nvim-treesitter-textobjects/swap.lua
```diff
@@ -6,20 +6,59 @@ local shared = require "nvim-treesitter-textobjects.shared"
...
+    ["end"] = {
+      line = end_row,
+      character = end_col,
+    },
+  }
+end
+
+---@param range Range4
+---@return string[]
+local function get_text(bufnr, range)
```
clason, line 28: Can't this be replaced by `nvim_buf_get_text`?

ofseed: Done, and this might be another reason for keeping `Range4`, it is more intuitive for some APIs like this.

clason: Agreed.

clason:
So I think I have a good idea of the problem landscape now; here's what I propose:

* keep the range4/range6 split
* try to replace as many custom functions with current Neovim API (vim.treesitter or vim.api, target Nvim 0.11/HEAD for that) as possible (in this PR, as long as you're not running out of steam)
* after that, look what is left and how much `vim.treesitter._range` would save us (follow-up PR).

ofseed:
I removed `to_vim_range` because we will have two different meanings of `Range4` at the same time, it brings much chaos. And I think this is all I can do for now.

clason:
Did you do some manual sanity check (not trusting these tests...)? 
* select inner/outer Lua function
* select local scope
* swap Lua function arguments
* (repeated) movement around Lua functions

Also, do you want to squash (some of) your commits? We don't need very finegrained commit history here, I believe.

lua/nvim-treesitter-textobjects/shared.lua
```diff
@@ -220,47 +214,80 @@ function M.find_best_range(bufnr, capture_string, query_group, filter_predicate,
...
+---@param range1 Range4
+---@param range2 Range4
+---@return boolean
+local function contains(range1, range2)
+  return is_in_range(range1, range2[1], range2[2]) and is_in_range(range1, range2[3], range2[4])
+end
+
+---@param range Range6
+---@return Range4
+function M.range6_range4(range)
```
clason, line 246: Not 100% happy with the name here (I know it's a temporary utility function, but old habits...)

Better (explicit) would be `range6_to_range4`, though that's a bit long maybe. Just `torange4` or `make_range4`?

ofseed:
> Did you do some manual sanity check (not trusting these tests...)?
> 
> * select inner/outer Lua function
> * select local scope
> * swap Lua function arguments
> * (repeated) movement around Lua functions

Except for swap, I use all of them when I write, and the deprecation of `to_vim_range` certainly takes a lot of regression, so I think I've tested a lot.

clason:
Can you briefly test `swap`, too, then? Just as a quick check that there's no regression there, as it's a bit fickle. If that works, I'll merge.

clason:
If you want to continue with the refactor using vim.treesitter._range, that would be welcome!

ofseed:
@clason Another reason I didn't use the upstream `_range` yet: I was wondering if it's possible to store the intermediate results as `TSNode` instead. In the current implementation, we immediately convert all `TSNode` instances into `Range` after the query is completed, which causes us to lose a lot of information that `TSNode` can contain. This might be unnecessary since we only need to use `TSNode:range()` to get the `Range` when we actually need it. Most importantly, many public APIs support `TSNode`, which should significantly reduce the amount of code in this plugin. Also, aligning with the upstream API could lower maintenance costs.

However, I'm not that familiar with `vim.treesitter`, and since `TSNode` is `userdata` and not a `table`, I noticed that this plugin uses some methods to store query results, and I'm not sure if those methods would work on `userdata`.

clason:
The idea was to precompute and cache as much as possible on buffer changes so that using the functions on an unchanged buffer is cheap. But nobody complained about speed even back when nothing was cached, so this might be premature optimization and not worth the complexity. (Caching is mostly needed for code that runs in a hot loop like highlighting and folding.)

I'm very open to any refactor that simplifies the code and relies more on upstream highlevel APIs. Do you want to open a PR for your suggestion and we can discuss details there?

ofseed:
I'm happy to continue with the refactoring, but I won't have any more time over the next one or two weeks. Next time I attempt it, I can open a PR (possibly a draft PR) for discussion. Feel free to implement your ideas, when I have time, I'll check the commit log to see if I can still contribute any help.

clason:
No hurry; thanks for your help, and I'm looking forward to the next PR ;)

PR #675, opened 2024-08-30
feat(zig): update queries to parser update

ribru17:
See https://github.com/nvim-treesitter/nvim-treesitter/pull/7097


amaanq:
nice

PR #676, opened 2024-08-30
fix(shared): resolve metadata from directives

ofseed:
A refactoring regression.


clason:
Going to merge this now as this is a self-contained fix (and the readme fix is valuable, too); if there's a better (more involved) fix, that can be done in the refactor PR. @ribru17

PR #678, opened 2024-09-04
feat(systemverilog): Add SV textobjects

henrykvdb:
Added very basic textobjects for SystemVerilog. Tested them in nvim also and they seemed to work fine :)

This is the first time I am writing treesitter querries so any critique is welcome!
(I ran the querry formatter so that should be okay)


henrykvdb:
After some further testing I realized my changeset is not fully complete.

I want to match code blocks like so:

```
(seq_block
    (simple_identifier)
    .
    (_) @_start @_end
    (_)* @_end
    .
    (#make-range! "block.inner" @_start @_end)) @block.outer
```

However the (simple_identifier) is optional and not always present, I tried making it optional by adding `?` to it

```
(seq_block
    (simple_identifier)?
    .
    (_) @_start @_end
    (_)* @_end
    .
    (#make-range! "block.inner" @_start @_end)) @block.outer
```

This however does not work because now the (simple_identifier) will be matched by the `(_) @_start @_end`.

So I would need either
- `?` but greedy, so it always matches (simple_identifier) if present OR
- I need to somehow replace it so (_) matches anything except `(simple_identifier)`.

I am sure there is some easy/clean syntax to accomplish this, but I could not manage to find it yet from reading through the documentation. Any suggestions?

henrykvdb:
I fixed some small issue and now everything behaves as expected.

This PR is linked to https://github.com/nvim-treesitter/nvim-treesitter/pull/7170 and should be accepted at the same time ideally.

ribru17 (approved):
LGTM! One nit: would you mind changing the `@_start @_end` double captures to just `@_start`? `make-range` knows to ignore the range if `@_end` is `nil`

henrykvdb:
Hi @ribru17 thank you for the review.

I don't have a setup ready to quickly test the proposed changes and I am not comfortable doing it without testing, however feel free to modify it as you like

clason:
(`#make-range!` is slated for removal anyway)

PR #682, opened 2024-09-06
feat: modify behaviour of `@assignment.outer` so it includes keywords like `const`, `local` etc. and semicolon

### LINKED ISSUE #681: Change behaviour of @assignment.outer to include `local` and similar keywords
Deleting an entire assignment is a common task and it would be nice to have a text object for that

New text object `@assignment.full` will capture the entire assignment. 

`[` and `]` denote what area the textobject will capture, if our cursor is at `*`

E.g. in javascript

```
[con*st foo = "foo";]
```
```js
[const bar = (bar) => "b*ar";]
```
```js
[const baz = function (str, num) {
  console.log(str)
  const obj = {
    str,
    num,
  }
retu*rn obj
};]
```

```js
const baz = function (str, num) {
  console.log(str)
  [const obj = {
    str,*
    num,
  };]
return obj
};
```

nikitarevenco:
Goal of PR (For all languages which support either of the first 1-4 text objects): 

- Change `assignment.outer` so that it captures extra tokens like `const` and semicolon

### ecma (js, jsx, ts, tsx):

```ts
const x = 5;
```
- `@assignment.outer`: `const x = 5;`

### c and c++

```c
int x = 5;
```
- `@assignment.outer`: `int x = 5;`

### bash

```bash
x=5
```

- `@assignment.outer`: `x=5`

### python

```python
x = 5
```

- `@assignment.outer`: `x=5`

### css and scss:
```css
body {
  margin: 1px 2px 3px 4px;
}
```

- `@assignment.outer`: `margin: 1px 2px 3px 4px;`

### Languages changed:

a -> b, c means b and c inherit from a, so no changes were made to b or c

- [x] apex 
- [x] go
- [x] ecma -> js, ts, jsx, tsx, 
- [x] lua
- [x] julia
 
### Languages not changed since they don't have keywords like `const` or similar, OR they were already working well:

- [x] nim
- [x] python
- [x] bash
- [x] yaml
- [x] rust 
- [x] cuda 
- [x] fish
- [x] css -> sass
- [x] hcl -> terraform
- [x] c -> fennel, ( cpp -> glsl, ( hlsl -> slang ) )
- [x] inko
- [x] matlab
- [x] odin
- [x] r
- [x] tact
- [x] v
- [x] vim

Closes #681

### DIFF
--- a/queries/apex/textobjects.scm
+++ b/queries/apex/textobjects.scm
@@ -99,6 +99,12 @@
 ] @comment.outer
 
 ; assignment
-(variable_declarator
-  name: (identifier) @assignment.lhs
-  value: (_) @assignment.rhs) @assignment.inner @assignment.outer
+(field_declaration
+  (variable_declarator
+    name: (identifier) @assignment.lhs
+    value: (_) @assignment.rhs) @assignment.inner) @assignment.outer
+
+(local_variable_declaration
+  (variable_declarator
+    name: (identifier) @assignment.lhs
+    value: (_) @assignment.rhs) @assignment.inner) @assignment.outer
--- a/queries/ecma/textobjects.scm
+++ b/queries/ecma/textobjects.scm
@@ -239,9 +239,10 @@
 ; number
 (number) @number.inner
 
-(variable_declarator
-  name: (_) @assignment.lhs
-  value: (_) @assignment.inner @assignment.rhs) @assignment.outer
+(lexical_declaration
+  (variable_declarator
+    name: (_) @assignment.lhs
+    value: (_) @assignment.inner @assignment.rhs)) @assignment.outer
 
 (variable_declarator
   name: (_) @assignment.inner)
--- a/queries/go/textobjects.scm
+++ b/queries/go/textobjects.scm
@@ -157,18 +157,22 @@
   left: (_) @assignment.lhs
   right: (_) @assignment.rhs @assignment.inner) @assignment.outer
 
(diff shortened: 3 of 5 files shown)


kiyoon:
I don't really agree that assignment.inner should be gone.

First, I considered assignment.lhs and rhs to be more of an addition for people who'd like granular control but I expected that with assignment.inner you can still do most things.

I just think that most of the time people expect assignment.outer to include the whole statement. The problem was that I think with languages with more complex syntax it was hard to design such thing (like it can consist of so many different mixture of const, final, variable types etc. in different order) and it ended up having a simple version of the implementation.

nikitarevenco:
> I don't really agree that assignment.inner should be gone.
> 
> First, I considered assignment.lhs and rhs to be more of an addition for people who'd like granular control but I expected that with assignment.inner you can still do most things.
> 
> I just think that most of the time people expect assignment.outer to include the whole statement. The problem was that I think with languages with more complex syntax it was hard to design such thing (like it can consist of so many different mixture of const, final, variable types etc. in different order) and it ended up having a simple version of the implementation.

Fair enough, I'll update the PR to restore behaviour of assignment.inner

nikitarevenco:
Update: I have looked over every single language, and modified all the ones that need modifying. Some languages like `Rust` and `C` already have expected behaviour: `@assignment.outer` selecting the *entire* assignment.

PR is ready for review

PR #683, opened 2024-09-07
feat(ecma): add support for @return.inner, @return.outer and @statement.outer

### LINKED ISSUE #686: javascript @return.inner, @return.outer, and @statement.outer support
It would be amazing to be able to use these text objects in the best language ever created

(and for typescript + jsx, which inherit from @ecma )

nikitarevenco:
This includes js, ts, jsx, and tsx

Closes #686


kiyoon:
Note that you shouldn't change the README manually because there is an automatic CI that does it

nikitarevenco:
> Note that you shouldn't change the README manually because there is an automatic CI that does it

I have reverted changes to the Md file

ribru17:
Would you mind running the query formatter? Then we can merge

nikitarevenco:
> Would you mind running the query formatter? Then we can merge

Ok!

ribru17:
Thank you! Also one last thing, I apologize for not mentioning earlier- would you mind putting this in an alteration? It would probably be easier to read. Something like:
```
[
  (for_statement)
  (while_statement)
  ...etc
] @statement.outer
```

nikitarevenco:
> Thank you! Also one last thing, I apologize for not mentioning earlier- would you mind putting this in an alteration? It would probably be easier to read. Something like:
> 
> ```
> [
>   (for_statement)
>   (while_statement)
>   ...etc
> ] @statement.outer
> ```
> 
> the return statement can stay on its own

Sure!, I agree this makes it more reasonable. I wasn't even aware of this syntax till now, that's cool!

ribru17:
Thanks!

PR #685, opened 2024-09-08
feat(typescript): add @annotation.inner and @annotation.outer

### LINKED ISSUE #684: @annotation.inner and @annotation.outer
I'd like to `@annotation.inner` and `@annotation.outer`

if this gets accepted, i'll do a PR for typescript

@annotation.inner and @annotation.outer
```scm
(type_annotation
  (_) @annotation.inner) @annotation.outer
```

If we have

```ts
let isStringVar: IsString<string> = true;
```
- `@annotation.inner`: `IsString<string>`
- `@annotation.outer`: `: IsString<string>`

We can use a different name, it's not a big deal to me

NikitaRevenco:
This PR adds support for two new text objcects

I'm going to add note to the README about the new text objects if this PR gets accepted

- Closes #684

### DIFF
--- a/queries/typescript/textobjects.scm
+++ b/queries/typescript/textobjects.scm
@@ -38,3 +38,6 @@
     .
     "}"
     (#make-range! "class.inner" @_start @_end)))
+
+(type_annotation
+  (_) @annotation.inner) @annotation.outer


clason:
Adding new textobjects is something we do not do lightly. You would need to add them (with explanation!) to `CONTRIBUTING.md` (which is used in CI to enforce valid captures) and the README table.

nikitarevenco:
> Adding new textobjects is something we do not do lightly. You would need to add them (with explanation!) to `CONTRIBUTING.md` (which is used in CI to enforce valid captures) and the README table.

I'm happy to do those steps and complete this PR for as many languages as I can if #684 gets accepted

NikitaRevenco:
i dont use neovim anymore so, someone else can rebase this PR if they want

Issue #627, opened 2024-05-29
bug: slow textobject select queries for Rust


aclementev:
**Describe the bug**
Opening a buffer for any Rust file with `textobject.select.enable = true` option is very slow.
The time difference when opening an empty `foo.rs` file is very noticeable (on my setup it's around ~700ms).

**To Reproduce**
1. Enable `select` plugin
2. Open a new rust file: `nvim foo.rs`
3. Notice the long time to open the buffer

**Expected behavior**
I expect the buffer to open quickly, not slow enough to be noticeable.

**Output of `:checkhealth nvim-treesitter`**
```
nvim-treesitter: require("nvim-treesitter.health").check()

Installation ~
- WARNING `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
- OK `node` found v20.12.1 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
- OK `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
  Version: cc (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0
- OK Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

OS Info:
{
  machine = "x86_64",
  release = "5.15.133.1-microsoft-standard-WSL2",
  sysname = "Linux",
  version = "#1 SMP Thu Oct 5 21:02:42 UTC 2023"
} ~

Parser/Features         H L F I J
  - bash                ✓ ✓ ✓ . ✓
  - c                   ✓ ✓ ✓ ✓ ✓
  - clojure             ✓ ✓ ✓ . ✓
  - comment             ✓ . . . .
  - css                 ✓ . ✓ ✓ ✓
  - diff                ✓ . . . .
  - dockerfile          ✓ . . . ✓
  - elixir              ✓ ✓ ✓ ✓ ✓
  - git_config          ✓ . . . .
  - gitcommit           ✓ . . . ✓
  - gitignore           ✓ . . . .
  - go                  ✓ ✓ ✓ ✓ ✓
  - gomod               ✓ . . . ✓
  - html                ✓ ✓ ✓ ✓ ✓
  - htmldjango          ✓ . ✓ ✓ ✓
  - javascript          ✓ ✓ ✓ ✓ ✓
  - json                ✓ ✓ ✓ ✓ .
[...]

Issue #461, opened 2023-06-05
Zig file takes a long time to open


lnc3l0t:
**Describe the bug**
When I open a `zig` file with this extension enabled it takes 3 seconds on my machine to start it

**To Reproduce**
I made a minimal config to reproduce:
`git clone https://github.com/lnc3l0t/treesitter-textobjects-issue /tmp/nvim; cd /tmp/nvim; . launch.sh`
It contains:
- `launch.sh` script which clone `nvim-treesitter` and `nvim-treesitter-textobjects`, installs the zig parser, launches neovim and benchmarks it.
- `init.lua` minimal neovim config that sets up `package.path` and `nvim-treesitter.config`
- `main.zig` a minimal zig file to open

**Expected behavior**
It should open like any other filetypes but on my machine the benchmarks take around 3 seconds:
<details>
<code>

Time

real    0m3.181s
user    0m3.159s
sys     0m0.024s

Hyperfine
Benchmark 1: nvim --clean -u init.lua main.zig -c "q"
  Time (mean ± σ):      3.202 s ±  0.007 s    [User: 3.181 s, System: 0.019 s]
  Range (min … max):    3.189 s …  3.209 s    10 runs

</code>
</details>

**Output of `:checkhealth nvim-treesitter`**

<details>
<code>

==============================================================================
nvim-treesitter: require("nvim-treesitter.health").check()
Installation ~
- OK `tree-sitter` found 0.20.8 (parser generator, only needed for :TSInstallFromGrammar)
- OK `node` found v20.2.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
- OK `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
 Version: cc (GCC) 13.1.1 20230429
- OK Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.
OS Info:
{
 machine = "x86_64",
 release = "6.3.5-arch1-1",
 sysname = "Linux",
 version = "#1 SMP PREEMPT_DYNAMIC Tue, 30 May 2023 13:44:01 +0000"
} ~
Parser/Features         H L F I J
 - bash                ✓ ✓ ✓ . ✓
 - c                   ✓ ✓ ✓ ✓ ✓
 - cmake               ✓ . ✓ . .
 - cpp                 ✓ ✓ ✓ ✓ ✓
[...]

PR #692, opened 2024-09-23
feat: cache textobjects created from a ts query

gepbird:
### Problem

When the user opens a file with this plugin installed (most notably latex, dart, zig and rust files), a slow [ts.get_query](https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/bf8d2ad35d1d1a687eae6c065c3d524f7ab61b23/lua/nvim-treesitter/textobjects/shared.lua#L99) will get called, that can take 3500ms for a latex file on an older machine. The reason for this is tree-sitter is trying to do some [aggressive initial analysis](https://github.com/tree-sitter/tree-sitter/issues/973#issuecomment-2182951245) of the query (eg. [latex/textobjects.scm](https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/bf8d2ad35d1d1a687eae6c065c3d524f7ab61b23/queries/latex/textobjects.scm)).

Neovim acknowledged that this function call is slow and [memoized](https://github.com/neovim/neovim/blob/78b85109338592c2bc89154278f2e961a14eee96/runtime/lua/vim/treesitter/query.lua#L216) it, however it doesn't get cached for too long, as the cache gets [garbage collected](https://github.com/neovim/neovim/blob/78b85109338592c2bc89154278f2e961a14eee96/runtime/lua/vim/func.lua#L17-L18). For me this cache gets cleared after opening 2-3 files, so it's not that useful.

### This PR's solution

This PR puts computed queries in a cache which lives as long as neovim is open.

### Further work

A proper upstream fix would be the best, but this performance issue is present for months and I don't see any progress on it.
In order to persist the cache between neovim sessions, we could store the cache somewhere in probably ~/.local/state/nvim. And for cache invalidation, we could also store the hash of the .scm files and invalidate the cache when the hash changes.

### Other issues and PRs

Helps a lot with #461, #627.
Upstream issue this PR is working around: https://github.com/tree-sitter/tree-sitter/issues/973.
In neovim a similar caching system was implemented in https://github.com/neovim/neovim/pull/24222, but got replaced with the garbage collected memoization in https://github.com/neovim/neovim/pull/25201.
Other plugins are resorting to this kind of workaround, for example [noice](https://github.com/folke/noice.nvim/blob/6c87c1d11c38180fb72bf8f45518a0a3e081afc1/lua/noice/text/treesitter.lua#L25).

### DIFF
--- a/lua/nvim-treesitter/textobjects/shared.lua
+++ b/lua/nvim-treesitter/textobjects/shared.lua
@@ -93,9 +93,14 @@ function M.get_query_strings_from_regex(query_strings_regex, query_group, lang)
   return query_strings
 end
 
+local textobjects_cache = {}
 function M.available_textobjects(lang, query_group)
   lang = lang or parsers.get_buf_lang()
   query_group = query_group or "textobjects"
+  if textobjects_cache[lang] and textobjects_cache[lang][query_group] then
+    return textobjects_cache[lang][query_group]
+  end
+  textobjects_cache[lang] = textobjects_cache[lang] or {}
   local parsed_queries = ts.get_query(lang, query_group)
   if not parsed_queries then
     return {}
@@ -109,6 +114,7 @@ function M.available_textobjects(lang, query_group)
       end
     end
   end
+  textobjects_cache[lang][query_group] = found_textobjects
   return found_textobjects
   --patterns = {
   --[2] = { { "make-range!", "function.inner", 2, 3 } },


clason:
Thank you, but the master branch is basically frozen; any development now happens on the `main` branch which will become the default (and which already contains caching).

clason:
(And if you add a cache, you need to make sure it gets invalidated on changes!)

gepbird:
@clason Thanks for the info. I just tried loading a dart file on `main` and it's instant, even for the first try, very impressive work! Latex may be a little slower if not the same speed, or at least I can feel the sometimes fast, sometimes slower loading that I noticed with neovim's garbage collected `memoize`.

How stable is the `main` branch? Since it's not the default I assume most users and distro package maintainers shouldn't switch yet, is there some tracking issue?

(Since AFAIK the query won't change during runtime, I deliberately didn't invalidate it.)

clason:
It's not stable yet; there's some regressions and some more work to be done that could be breaking (see open PRs labeled NEXT). No ETA but we hope those will get resolved in a timely manner. It's volunteer work, though, so impossible to plan.

ribru17:
Note that this will likely be helped greatly by https://github.com/neovim/neovim/pull/30227, as it will allow the `query.get` memoization to use a strong table and be manually invalidated only when necessary (e.g. after `query.set`)

gepbird:
Thanks for your work in upstream and linking it here, that will also help other plugins :)

PR #697, opened 2024-09-29
feat(tact): update queries

novusnota:


novusnota:
Updated queries to Tact v1.6.0 :)

PR #708, opened 2024-10-18
fix(julia): query syntax error

phanen:
This is introduced by [aad2c8e6b029](https://github.com/nvim-treesitter/nvim-treesitter-textobjects/commit/aad2c8e6b029ada30621e67780801420ee9debe9).

### DIFF
--- a/queries/julia/textobjects.scm
+++ b/queries/julia/textobjects.scm
@@ -12,14 +12,14 @@
   .
   (_) @block.inner
   (_)? @block.inner .)
-)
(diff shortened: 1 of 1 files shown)


kiyoon:
I thought query PRs should go to master and should be rebased to main. Am I right? @clason

ribru17:
In this case I think no because I changed (and broke, here) the queries on main specifically (oops)

ribru17 (approved):

PR #717, opened 2024-11-18
Update @statement.outer query for Ecma to also include lexical declarations

zenshixd:
Added lexical declarations to @statement.outer capture group, e.g:
```
const a = 1;
```

should be selectable using @statement.outer capture group.


kiyoon:
Thank you!

PR #718, opened 2024-11-21
feat(c_sharp) add @assignment support for all possible assignable statements

prashanthsp6498:
Supports for all assignment statements.

Lambda Method
```cs
    public static string Method() => $"return result";
```

class/struct/record fields
```cs
    const int CONST_FIELD = 10;
    public static Node head;
    public static Node tempHead = new Node();

```

local variables declarations
```cs
    int count = 10;
    int[] arr = new int[] {1, 2, 3, 4, 5, 6};
    head = new Node{data = arr[0]};
    count = 100;
```
Above local fields also support in field scope

### DIFF
--- a/queries/c_sharp/textobjects.scm
+++ b/queries/c_sharp/textobjects.scm
@@ -196,3 +196,64 @@
 
 ; comments
 (comment) @comment.outer
+
+; assignments
+(field_declaration
+  (_)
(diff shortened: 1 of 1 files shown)


prashanthsp6498:
Can anyone help me on this
Why above test failing?

PR #720, opened 2024-11-22
feat(c_sharp) add @assignment support for all possible assignable statements

prashanthsp6498:
Supports for all assignment statements.

Lambda Method
```cs
    public static string Method() => $"return result";
```

class/struct/record fields
```cs
    const int CONST_FIELD = 10;
    public static Node head;
    public static Node tempHead = new Node();

```

local variables declarations
```cs
    int count = 10;
    int[] arr = new int[] {1, 2, 3, 4, 5, 6};
    head = new Node{data = arr[0]};
    count = 100;
```
Above local fields also support in field scope

### DIFF
--- a/queries/c_sharp/textobjects.scm
+++ b/queries/c_sharp/textobjects.scm
@@ -196,3 +196,64 @@
 
 ; comments
 (comment) @comment.outer
+
+; assignments
+(field_declaration
+  (_)
+  (variable_declaration
+    type: (_)
+    (variable_declarator
+      name: (_) @assignment.lhs
+      (_) @assignment.rhs) @assignment.inner)) @assignment.outer
+
+(field_declaration
+  (variable_declaration
+    type: (_)
+    (variable_declarator
+      name: (_) @assignment.lhs
+      (_) @assignment.rhs) @assignment.inner)) @assignment.outer
+
+(field_declaration
+  (_)
+  (variable_declaration
+    type: (_)
+    (variable_declarator
+      name: (_) @assignment.lhs @assignment.inner))) @assignment.outer
+
+(local_declaration_statement
+  (_)
+  (variable_declaration
+    type: (_)
+    (variable_declarator
+      name: (_) @assignment.lhs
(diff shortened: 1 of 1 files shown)


prashanthsp6498:
Any update on this PR...

kiyoon:
@prashanthsp6498 Hi, sorry for the delay. C# is not a language I personally use and I find it time-consuming to test all query PRs.  
Could you kindly add some test cases?

1. Make an example C# file in `tests/select/c_sharp/*` (the same as the description is fine)
2. The assignment textobjects are not defined in the `scripts/minimal_init.lua`. Add bindings for `vig` (`@assignment.inner`), `vag` (`@assignment.outer`), `vik` (`@assignment.lhs`), `viv` (`@assignment.rhs`). The k and v refers to key and value.
3. Write `tests/select/c_sharp_spec.lua` similar to the one in python. Make sure you write clear comments to indicate what the test is doing.
    a. For example, I would equality-compare commands like `vt=bd` (delete everything before = sign, except for the spaces) with `vig` and `vik`.

@clason I think this should be mandatory in all query PRs. It's hard to really keep track of all queries manually.

queries/c_sharp/textobjects.scm
```diff
@@ -196,3 +196,64 @@
...
+  (_)
+  (variable_declaration
+    type: (_)
+    (variable_declarator
+      name: (_) @assignment.lhs @assignment.inner))) @assignment.outer
+
+(expression_statement
+  (assignment_expression
+    left: (_) @assignment.lhs
+    right: (_) @assignment.rhs)) @assignment.inner @assignment.outer
```
kiyoon, line 248: Inner and outer should not mean the same thing. Inner refers to "either" LHS or RHS, outer means select both. Sorry, this isn't documented very clearly as of now.

prashanthsp6498: Ok, I will fix these queries and add test cases.

clason:
Closing as stale; new PR (against `main`) welcome.

PR #721, opened 2024-11-23
fix: offset_encoding is required now in `make_position_params`

rockyzhang24:
Adapt to the upstream change https://github.com/neovim/neovim/commit/629483e24eed3f2c07e55e0540c553361e0345a2 to get rid of the warning message "warning: offset_encoding is required, using the offset_encoding from the first client".

Since the commit https://github.com/neovim/neovim/commit/2dcbfe78fcec5f73ce061bb24b718187b9c6b134, `params` in `buf_request` can accept a function that takes the client as a parameter. So instead of sending the same params to all servers, the params with correct offset_encoding will send to each single server.

### DIFF
--- a/lua/nvim-treesitter/textobjects/lsp_interop.lua
+++ b/lua/nvim-treesitter/textobjects/lsp_interop.lua
@@ -111,7 +111,11 @@ function M.peek_definition_code(query_string, query_group, lsp_request, context)
   if vim.tbl_contains(vim.api.nvim_list_wins(), floating_win) then
     vim.api.nvim_set_current_win(floating_win)
   else
-    local params = vim.lsp.util.make_position_params()
+    local win_id = vim.api.nvim_get_current_win()
+    local params = vim.fn.has "nvim-0.11" == 0 and vim.lsp.util.make_position_params()
+      or function(client)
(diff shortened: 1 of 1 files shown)


clason:
That is an LSP change; why should this affect a treesitter plugin?

rockyzhang24:
@clason Sorry to ping you but I got an email saying I have a comment from you "That is an LSP change; why should this affect a treesitter plugin?" But, I didn't see this comment here. 

To answer it: Now `offset_encoding` is now required for `make_position_params` after that commit; otherwise a warning is shown. Thank you.

rockyzhang24:
Btw, this change needs Noevim nightly, do I need to handle the backward compatibility?

rockyzhang24:
I formatted the code conforming to stylua. Now I think it's ready to be previewed and merged. Thank you very much.

rockyzhang24:
I just realized that the `has` function for checking version numbers returns `0` or `1` instead of a boolean. I’ve fixed it, and everything should be fine now.

JeanMertz:
Any chance of this getting merged? 🙏

kiyoon:
Thank you!

PR #723, opened 2024-12-08
feat(ocaml): Add OCaml queries

beajeanm:
I'm a treesitter queries newbie, so I'd appreciate the feedback on these queries.

This is an example of syntactically correct ocaml I used to test them:

```ocaml
let example_function param1 param2 = param1 + param2

class example_object =
  object (self)
    method do_stuff x = raise Not_found
  end

(* This is a comment. *)

let some_loops () =
  for variable = start_value to end_value do
    expression
  done;
  while boolean - condition do
    expression
  done

let conditionals () =
  let if_condition = if a then b else c in
  let match_cases = match a with b -> 1 | c -> 2 in
  let function_expression = function b -> 1 | c -> 2 in
  ()

let test_call =
  (* A function call, followed by 3 arguments. *)
  String.fold (fun a b -> a) 0 xs

let some_blocks =
  let begin_block = begin
    let inner_block = a in
    b
    end
  in
  let paren_block =
    (
      let inner_block = a in
      b
    )
  in c

```

### DIFF
--- a/queries/ocaml/textobjects.scm
+++ b/queries/ocaml/textobjects.scm
@@ -1,0 +1,57 @@
+(value_definition
+  (let_binding
+    body: (_) @function.inner)) @function.outer
+
+(method_definition
+  body: (_) @function.inner) @function.outer
+
+(class_definition
+  (class_binding
+    body: (_) @class.inner)) @class.outer
+
+(for_expression
+  (do_clause
+    (_) @loop.inner)) @loop.outer
+
+(while_expression
+  (do_clause
+    (_) @loop.inner)) @loop.outer
+
+(if_expression
+  condition: (_)
+  (then_clause
+    (_) @conditional.inner)
+  (else_clause
+    (_) @conditional.inner)) @conditional.outer
+
+(if_expression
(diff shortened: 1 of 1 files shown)


rbjorklin:
Hey @beajeanm! I just stumbled over this plugin and now I'm curious if you have been using this with OCaml? If you are would you mind updating the README in the PR to show which built-in textobjects you have implemented?

beajeanm:
> Hey @beajeanm! I just stumbled over this plugin and now I'm curious if you have been using this with OCaml? If you are would you mind updating the README in the PR to show which built-in textobjects you have implemented?

Hey, yes, I'm using this in my fork of the plugin. It work fine for the way I write my OCaml 😄 .
I'll try to update the README in the next couple of days.

kiyoon:
The README is generated automatically, no need to change it!

And I had no time to test it but if you say it's been working for months, I'd accept it.

rbjorklin:
For what it's worth I ended up trying your fork after posting and it has worked well with my rudimentary usage as well.

PR #741, opened 2025-03-12
feat(jsx): add jsx/tsx element captures

mawkler:
This PR adds `@jsx_element.inner`/`@jsx_element.outer` for jsx and tsx. Example usage:

- `dit`: turns `<Foo bar="bar" />` into `<Foo />`
- `dat`: turns `<Foo bar="bar" />` into ` ` (deletes the entire element)

With the following config:

```lua
vim.keymap.set({ "o", "x" }, "at", "<cmd>TSTextobjectSelect @jsx_element.outer<CR>") -- `t` for "tag"
vim.keymap.set({ "o", "x" }, "it", "<cmd>TSTextobjectSelect @jsx_element.inner<CR>")
```

This also works with jsx/tsx element pairs (i.e. non-self-closing tags), just like how Neovim's default `at`/`it` work.

### DIFF
--- a/queries/jsx/textobjects.scm
+++ b/queries/jsx/textobjects.scm
@@ -3,3 +3,22 @@
 (jsx_attribute
   (property_identifier)
   (_) @parameter.inner) @parameter.outer
+
+; Self closing tags
+(jsx_self_closing_element
+  name: (identifier)
+  .
+  (_) @_start
+  (_)* @_end
+  .
+  (#make-range! "jsx_element.inner" @_start @_end)) @jsx_element.outer
+
+; Paired tags
+(jsx_element
+  open_tag: (_)
+  .
+  (_) @_start
+  (_)* @_end
+  .
+  close_tag: (_)
+  (#make-range! "jsx_element.inner" @_start @_end)) @jsx_element.outer


clason:
I think that defeats the point of this plugin (consistent _language independent_ text objects).

Such custom queries belong in user config or a filetype pluign.

mawkler:
@clason What's the downside of including language-specific queries? It can't have any performance impact, right?

For what it's worth, this repo does contain [LaTeX specific queries](https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/master/CONTRIBUTING.md#textobjects) already.

clason:
Maintenance and documentation burden. LaTeX is a snowflake, and that decision needs to be revisited.

mawkler:
Ok

kiyoon:
@mawkler Hi, not sure if you understood, but it doesn't have to be merged for you to use it. Just put the queries in your config and you can use this plugin with the query you want!

mawkler:
@kiyoon Yes I'm aware of that, but I wanted to share this with the rest of the community since it feels like a very basic but useful feature for anyone working with jsx/tsx.

Are you aware of any plugin like nvim-treesitter-textobjects, that includes language-specific queries? The more queries the merrier, in my opinion.

kiyoon:
@mawkler You don't have to implement another nvim-treesitter-textobjects. Just put the query in the plugin and that's it.

There are many plugins that include treesitter queries to utilise nvim-treesitter, nvim-treesitter-textobjects, etc. And of course there are many language-specific plugins that include some specific queries for that language.

mawkler:
To anyone reading this, I put this functionality in a plugin: [jsx-element.nvim](https://github.com/mawkler/jsx-element.nvim).

clason:
This Is The Way.

kiyoon:
But the plugin doesn't really need lua at all. Just the queries is fine.

```lua
  textobjects = {
    select = {
      keymaps = {
        ["at"] = "@jsx_element.outer",
      }
    }
  }
}
```

And the dependencies would be like this

```lua
{
  'nvim-treesitter/nvim-treesitter',
  dependencies = {
    'nvim-treesitter/nvim-treesitter-textobjects',
    'mawkler/jsx-element.nvim',
  },
}

kiyoon:
There is no API for that unfortunately. You need to map the function or vim command to your liking. And the fact that it's too specific is exactly the reason it didn't belong in the plugin, so, you have a point. Nice work!

srcrip:
For what it's worth, as a random Neovim user, I think this is the wrong call. I understand you don't want language specific stuff in this repo, but then I think you guys should make a `nvim-treesitter-textobjects-extra` repo and invite contributions from the community.

PR #748, opened 2025-04-03
fix (kotlin): capture `@class.outer` on classes without a body

alphashuro:
In Kotlin classes, the body is optional, but the `@class` text objects were failing to capture classes without a body.
This change makes `@class.outer` work for those cases, but `@class.inner` is ignored since these classes does not have an "inner"

### DIFF
--- a/queries/kotlin/textobjects.scm
+++ b/queries/kotlin/textobjects.scm
@@ -2,7 +2,7 @@
   [
     (class_body)
     (enum_class_body)
-  ] @class.inner) @class.outer
+  ]? @class.inner) @class.outer
 
 [
   (function_declaration


alphashuro:
I see flakey tests but they seem unrelated to my change, are these flakey tests?

kiyoon:
The tests are broken since nvim 0.11, so ignore those.

Can you explain what bodyless class is? I'm new to Kotlin.

kiyoon:
For example,

1. If the class is bodyless, how does @class.inner work?
2. Does the change not affect the classes with a body?

alphashuro:
According to [the Kotlin docs](https://kotlinlang.org/docs/classes.html), the body of a class is optional.

It looks like this
```
class Person
```

> If the class is bodyless, how does @class.inner work?

`@class.inner` is just ignored, because the class has no "inside". Inspecting the treesitter tree at on this class shows

```
(class_declaration
  (type_indentifier))

> Does the change not affect the classes with a body?

No, it only makes the body _optional_, but when the body exists it is captured the same way it was before this PR

alphashuro:
just to be clear, the word "bodyless" is not actually a technical term 😅 I updated the PR title to make it clearer

kiyoon:
Thank you!

PR #755, opened 2025-04-26
feat(toml): add `@comment.outer`

chrisgrieser:
The failing tests are about something about python which I didn't touch; also they fail at other PRs as well, so I assume those can are not related to this PR?


kiyoon:
Yes, don't worry about the tests. Thank you!

PR #757, opened 2025-05-06
feat(css): add `@comment.outer`

chrisgrieser:
edit: the failing tests seem to be some python issue, so I assume they are irrelevant to this PR and can thus be ignored?


kiyoon:
Thank you!

PR #782, opened 2025-07-05
fix(python):fix class/function with decorator

SavingFrame:
If in the same file you have both class/function with decorator and without it then move/select won't work for class/function without decorator. I found this bug in the `main` branch, so maybe main problem not in the query, but this fix works.

### DIFF
--- a/queries/python/textobjects.scm
+++ b/queries/python/textobjects.scm
@@ -1,10 +1,14 @@
-((decorated_definition)?
-  (function_definition
-    body: (block)? @function.inner)) @function.outer
+(decorated_definition
+  (function_definition)) @function.outer
 
-((decorated_definition)?
-  (class_definition
-    body: (block)? @class.inner)) @class.outer
+(function_definition
+  body: (block)? @function.inner) @function.outer
+
+(decorated_definition
+  (class_definition)) @class.outer
+
+(class_definition
+  body: (block)? @class.inner) @class.outer
 
 (while_statement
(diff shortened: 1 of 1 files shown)


kiyoon:
Thank you for the PR.

Firstly, you can't directly change queries on the main branch. Instead, they have to be changed on master and then later rebase them to the main.

Second, I don't have the problem on the master branch. However, I have a problem where the decorator aren't really selected.

SavingFrame:
> Thank you for the PR.
> 
> Firstly, you can't directly change queries on the main branch. Instead, they have to be changed on master and then later rebase them to the main.
> 
> Second, I don't have the problem on the master branch. However, I have a problem where the decorator aren't really selected.

Ye, i have that problem with decorator as well, but if set cursor position on the decorator and use "class.outer" it would work.
But in my PR i tried to fix problem with selection in the main. We have this problem only in the main branch. I recorded a few videos with comparing master to main behavior.
1. treesitter + textobjects from the main

https://github.com/user-attachments/assets/0a92febf-082b-44c1-8e0f-5710b77fa68c


2. treesitter+textobject from the master

https://github.com/user-attachments/assets/9e211ea4-56fc-471e-b9f3-031f17697224


I used select inner/outer class

kiyoon:
Regardless of the initial claim, I found this change helpful because if

```python
@decorator
async def foo():
    pass

def bar():
    pas
```

Selecting from the decorator selects all, including the below function. Thus merging this.

PR #784, opened 2025-07-17
feat(enforce): add enforce queries

simonvic:
This PR adds back again queries for the Enforce language (currently only available on the master branch)

Compared to the version on master, the queries don't use `make-range` anymore


kiyoon:
Sorry, I missed this. I added on #820 so if there isn't anything missing, I'll close this.

simonvic:
Looks good, thanks :)

Issue #775, opened 2025-05-30
Repeating some moves gives "Vim(normal):E523: Not allowed here" error on main


michelesr:
**Describe the bug**
In the `main` branch, now you need to set `{expr = true}` when mapping the `;`, `,` repeat functions, so that you can also configure repeat for builtins `f`,`F`,`t`,`T`.

```lua
-- Repeat movement with ; and ,
-- ensure ; goes forward and , goes backward regardless of the last direction
vim.keymap.set({ "n", "x", "o" }, ";", ts_repeat_move.repeat_last_move_next, { expr = true })
vim.keymap.set({ "n", "x", "o" }, ",", ts_repeat_move.repeat_last_move_previous, { expr = true })
```

By doing this, moves that make use of `:normal` commands return error.

**To Reproduce**
1. configure mapping for `repeat_last_move_next` and `repeat_last_move_previous`
2. try a repeatable action, for example "go to next start of `@function.outer`" (e.g. `]m`)
3. try to repeat with `;`, `,`

**Expected behavior**
The action should be repeated without errors.

** What happens **
```
E5108: Error executing lua: vim/_editor.lua:0: nvim_exec2(), line 1: Vim(normal):E523: Not allowed here
stack traceback:
        [C]: in function 'nvim_exec2'
        vim/_editor.lua: in function 'cmd'
        ...ter-textobjects/lua/nvim-treesitter-textobjects/move.lua:16: in function 'goto_node'
        ...ter-textobjects/lua/nvim-treesitter-textobjects/move.lua:135: in function <...ter-textobjects/lua/nvim-treesitter-textobjects/move.lua:41>
```

**Output of `nvim --version`**
```
NVIM v0.11.1
Build type: Release
LuaJIT 2.1.1744318430
```

** Additional info **

When I was trying to make my own repetable function work, I noticed that wrapping with `vim.schedule()` bypasses the issue:

```lua
  local diag_move = require('nvim-treesitter-textobjects.repeatable_move').make_repeatable_move(function(opts)
    if opts.forward then
      -- schedule() prevents Vim(normal):E523: Not allowed here
      -- that would arise because ; , are mapped with { expr = true }
      vim.schedule(function() vim.diagnostic.jump({count = 1}) end)
    else
      vim.schedule(function() vim.diagnostic.jump({count = -1}) end)
    end
  end)
```

michelesr:
[...]

PR #795, opened 2025-08-27
fix(repeatable_move): bug when nil was returned or normal command used; no longer requires { expr = true }

mezdelex:
Should fix https://github.com/nvim-treesitter/nvim-treesitter-textobjects/issues/775

https://github.com/user-attachments/assets/afecce63-6db3-42c8-977b-e1a209ae9c68

Not sure if this fix is already being addressed in another way or refactor process, but just in case, the move_repeatable function was failing when no string was being returned (nil in this case) due to { expr = true } keymap options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,8 +170,8 @@ local ts_repeat_move = require "nvim-treesitter-textobjects.repeatable_move"
 
 -- Repeat movement with ; and ,
 -- ensure ; goes forward and , goes backward regardless of the last direction
-vim.keymap.set({ "n", "x", "o" }, ";", ts_repeat_move.repeat_last_move_next, { expr = true })
-vim.keymap.set({ "n", "x", "o" }, ",", ts_repeat_move.repeat_last_move_previous, { expr = true })
+vim.keymap.set({ "n", "x", "o" }, ";", ts_repeat_move.repeat_last_move_next)
+vim.keymap.set({ "n", "x", "o" }, ",", ts_repeat_move.repeat_last_move_previous)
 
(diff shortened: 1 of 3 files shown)


mezdelex:
I don't have a clue about Fennel, but it looks like there's a bug in the Fennel textobject parser 
image
as I'm getting this error indicator (probably didn't install Fennel right...) but it's exactly at line 105 as the pipeline indicates 
image
it looks like set_form text object doesn't have binding_pair

Anybody with more experience with Fennel I could ping? this looks like an existing bug

kiyoon:
Can you keep the change minimal, like get rid of the fennel textobject changes?

Also, you can just paste the snippet from the `master` branch because it's been working without issues. Introducing a new logic can introduce another bug.

mezdelex:
@kiyoon oh yeah my apologies.

Discarded the mentioned textobject change and kept one of the both suggested keymaps options for repeatable_move commented as you suggested.

kiyoon:
For example, the `master` has this snippet.

https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/71385f191ec06ffc60e80e6b0c9a9d5daed4824c/lua/nvim-treesitter/textobjects/repeatable_move.lua#L107-L121

Is there any reason why you would change it to `feedkeys`? If not, I'd prefer to stick to the same logic.

mezdelex:
No particular reason really, I just searched for a way to feed keys to the input, but the master branch snippet should in fact be more performant. Changed as suggested.

PR #802, opened 2025-09-18
feat(prisma): add prisma queries

purarue:
- **refactor!: make this an standalone plugin**
- **fix: regressions after refactor (#607)**
- **fix: handle nil start_nodes for `query_predicates`. (#655)**
- **fix: modify range instance instead of creating a new table**
- **fix: correct order and key relationship**
- **docs: fix the position of the `include_surrounding_whitespace`**
- **fix: avoid pass nonexistent field**
- **docs: parameter should be optional**
- **docs: `RepeatableMove.func` could be a string**
- **refactor: make `to_vim_range` return a `Range` instead of four numbers**
- **fix: avoid unnecessary closures**
- **feat!: remove lsp interop**
- **fix(repeatable_move): remove deprecated functions**
- **refactor(repeatable_move)!: only expose necessary API**
- **refactor(repeatable_move): inline `set_last_move`**
- **ci(tests): fix parser installation**
- **fix(repeatable_move): correct vim.cmd for f/F/t/T**
- **fix(shared): correct language detection**
- **fix(minimal_init): correct signature, use locals**
- **refactor: move main module to init.lua**
- **refactor(shared): move out Range class**
- **refactor(shared): inline and memoize match preparation**
- **refactor(shared): more memoize, type fixes**
- **refactor: consistent local api**
- **refactor(shared)!: drop support for specifying textobjects via regex**
- **refactor(shared): move {next,previous}_textobject to swap**
- **refactor!: fail silently if textobjects not available**
- **refactor(range): remove `range.lua`**
- **refactor(shared)!: remove `to_vim_range`**
- **refactor(swap): remove custom `get_text`**
- **refactor(shared): better convertion function name**
- **refactor(repeatable_move): extract the `MoveFunction` interface (#671)**
- **fix(shared): resolve metadata from directives**
- **docs(readme): fix swap example**
- **feat: support quantified captures (#612)**
- **fix(julia): query syntax error**
- **fix(repeatable_move)!: make `d;` work like the builtin behavior (#707)**
- **docs: correct typo (#727)**
- **fix: Adjust selected range if `set selection?` is `exclusive` (#687)**
- **fix(julia): adapt queries to parser changes**
- **fix(shared): ensure tree is parsed**
- **feat(tests): port test infrastructure**
- **chore: reformat with new stylua settings**
- **chore: remove obsolete scripts**
- **docs(readme): regenerate**
- **fix(lua): fix luals warnings**
- **fix(queries): fix lint**
- **ci: migrate to nvim-treesitter**
- **fix(repeatable_move): bug when nil was returned or normal command used; no longer requires { expr = true } (#795)**
- **feat(prisma): add prisma queries**


purarue:
ahh - so that is why I saw these kinda PRs on the repo, opened against the wrong target, my bad

PR #803, opened 2025-09-18
feat(prisma): add prisma queries

purarue:
```
model Product {
  id     String  @id @default(auto()) @map("_id") @db.ObjectId
  name   String
  photos Photo[]
}
```

I did not add support for the attributes (like `@id @default(auto())`) and such above, was not sure if those would be `@parameter` as well or `@attribute`? I don't see that many uses of attribute in the codebase, wasnt sure if that would make sense to use that.

So, just sticked to basic comment/field support for most commonly used things (model/enums/types ([those look like structs](https://www.prisma.io/docs/orm/reference/prisma-schema-reference#type), hence `@class.outer`)


purarue:
updated to make this match the new documentation format

PR #812, opened 2025-10-03
fix(select): fix adjusting end column of selection

### LINKED ISSUE #809: Column value outside range when selecting @function.inner or @block.outer
## Description

There seems to be an issue when using `@inner` or `@outer` textobjects with certain parsers (e.g., Go, Markdown) with objects like **functions** or **blocks**.

In some cases, the selection fails with an error because the calculated column is out of range.

## Steps to reproduce

* **Neovim version:** `NVIM v0.12.0-dev-1305+ga0c60e819d, Build type: Debug, LuaJIT 2.1.1753364724`
* **Treesitter version:** `0.25.8 (f2f197b6b27ce75c280c20f131d4f71e906b86f7)`
* **Plugin version:** `commit @1b2d85d3 (main branch)`

1. In a Go file, create a simple function such as:

   ```go
   func main() {
       // some irrelevant code here
   }
   ```

2. Call:

   ```lua
   require("nvim-treesitter-textobjects.select").select_textobject("@function.inner", "textobjects")
   ```

3. Observe the following error:

   <details>
   <summary>Output</summary>

   ```
   E5108: Lua: ...r-textobjects/lua/nvim-treesitter-textobjects/select.lua:32: Column value outside range
   stack traceback:
           [C]: in function 'nvim_win_set_cursor'
           ...r-textobjects/lua/nvim-treesitter-textobjects/select.lua:32: in function 'update_selection'
           ...r-textobjects/lua/nvim-treesitter-textobjects/select.lua:174: in function 'select'
           .../elias/.config/nvim/lua/elias/plugins/lsp/treeSitter.lua:101: in function <.../treeSitter.lua:100>
   ```

   </details>

This also occurs with `@block.outer` in Markdown files.

## Root cause (tentative)

From some debugging, the problem seems to be in the `update_selection` function inside
`nvim-treesitter-textobjects/lua/nvim-treesitter-textobjects/select.lua`:

```lua
api.nvim_win_set_cursor(0, { end_row + 1, end_col - end_col_offset })
```

The issue appears when `end_col - end_col_offset` is **less than 0** or **greater than the actual number of columns in the line**, causing `nvim_win_set_cursor` to fail.

## Hotfix

As a quick workaround in my local environment, I added a bounds check before setting the cursor:

```lua
-- Check bounds
local line_length = vim.fn.col({ end_row + 1, '$' }) - 1
end_col = end_col - end_col_offset
if end_col < 0 then
  end_col = 0
end
if end_col > line_length then
  end_col = line_length
end

-- Position is 1, 0 indexed.
api.nvim_win_set_cursor(0, { start_row + 1, start_col })
vim.cmd('normal! o')
api.nvim_win_set_cursor(0, { end_row + 1, end_col })
```

This prevents the error and makes selections work again, at least in the cases I tested.

## Notes

* I am not deeply familiar with the internals of this plugin, so I cannot say for certain if this is the **proper place** to fix the bug.
* This hotfix simply allowed me to continue using my local environment without errors.
* If this is indeed the right place for the fix, I’d be happy to open a PR with the change.

w1ck3dg0ph3r:
This fixes transformation from TSNode range to nvim_win_set_cursor (1,0)-based coordinates in `update_selection`.

fixes #809


kiyoon:
Thank you.

In nvim-treesitter master branch, there is this equivalent function.

https://github.com/nvim-treesitter/nvim-treesitter/blob/42fc28ba918343ebfd5565147a42a26580579482/lua/nvim-treesitter/ts_utils.lua#L253-L280

I think you can copy-paste this snippet in this plugin, because it has been tested and used for many years thus it is very stable.

w1ck3dg0ph3r:
> I think you can copy-paste this snippet in this plugin, because it has been tested and used for many years thus it is very stable.

My reasoning for not lifting get_vim_range from ts_utils verbatim was to better align the signature to this module, otherwise they are indeed equivalent (apart from working only on current buffer, which doesn't seem to be a factor for selection).

kiyoon:
Sorry, but this is superseded by #828

Issue #813, opened 2025-10-04
Error when opening Zig file


ocglossop:
**Describe the bug**
When opening a zig file, an error appears. This only occurs when the parser for Zig is installed. 

**To Reproduce**
1. Launch LazyVim with nvim-treesitter and nvim-treesitter-textobjects installed.
2. Run `TSInstall zig`.
3. Open a file with the `.zig` extension. You may need to repeat this multiple times before the error occurs.
4. A window should open below with the following message:
<details>
<code>
   Error  11:42:47 msg_show.lua_error Error executing vim.schedule lua callback: vim/_editor.lua:0: nvim_exec2()[1]..BufReadPost Autocommands for "*": Vim(append):Error executing lua callback: /opt/nvim-linux-x86_64/share/nvim/runtime/filetype.lua:36: nvim_exec2()[1]..BufReadPost Autocommands for "*"..FileType Autocommands for "*": Vim(append):Error executing lua callback: ...x-x86_64/share/nvim/runtime/lua/vim/treesitter/query.lua:373: Query error at 59:3. Impossible pattern:
  "("
  ^

stack traceback:
	[C]: in function '_ts_parse_query'
	...x-x86_64/share/nvim/runtime/lua/vim/treesitter/query.lua:373: in function 'fn'
	...inux-x86_64/share/nvim/runtime/lua/vim/func/_memoize.lua:78: in function 'fn'
	...inux-x86_64/share/nvim/runtime/lua/vim/func/_memoize.lua:78: in function 'get'
	.../share/nvim/lazy/LazyVim/lua/lazyvim/util/treesitter.lua:23: in function 'have_query'
	.../share/nvim/lazy/LazyVim/lua/lazyvim/util/treesitter.lua:40: in function 'have'
	...are/nvim/lazy/LazyVim/lua/lazyvim/plugins/treesitter.lua:155: in function <...are/nvim/lazy/LazyVim/lua/lazyvim/plugins/treesitter.lua:154>
	[C]: in function 'nvim_cmd'
	/opt/nvim-linux-x86_64/share/nvim/runtime/filetype.lua:36: in function </opt/nvim-linux-x86_64/share/nvim/runtime/filetype.lua:35>
	[C]: in function 'pcall'
	vim/shared.lua: in function <vim/shared.lua:0>
	[C]: in function '_with'
[...]

PR #817, opened 2025-10-28
fix(zig): wrap parentheses patterns in arguments node for call_expression

FilipNikolovski:
This update adapts the queries to the parser changes.

Fixes: #813

### DIFF
--- a/queries/zig/textobjects.scm
+++ b/queries/zig/textobjects.scm
@@ -70,21 +70,23 @@
 ; arguments
 (call_expression
   function: (_)
-  "("
-  "," @_start
-  .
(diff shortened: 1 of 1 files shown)


kiyoon:
It's because the master branch of nvim-treesitter is frozen, so the zig parser isn't updated. Thus this repo also uses the old syntax on the master branch, and it makes more sense to freeze the master branch and update queries on the main branch from now.

Can you open the PR on the main branch? Thank you

FilipNikolovski:
Here's the new PR: https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/819

Issue #811, opened 2025-10-02
Error with fennel "Query error at 105:3. Impossible pattern: (binding_pair"


meinside:
**Describe the bug**
1. I see error messages with plugins that depend on nvim-treesitter-textobjects like this:

```
E5108: Error executing lua: Vim:Error executing Lua callback: BufNewFile Autocommands for "*": Vim(append):Error executing lua callback: ...rew/Cellar/neovim/0.11.4/share/nvim/runtime/filetype.lua:36: BufNewFile Autocommands for "*"..FileType Autocommands for "*": Vim(append):Error executing lua callback: ...m/0.11.4/share/nvim/runtime/lua/vim/treesitter/query.lua:373: Query error at 105:3. Impossible pattern:
  (binding_pair
  ^

stack traceback:
	[C]: in function '_ts_parse_query'
	...m/0.11.4/share/nvim/runtime/lua/vim/treesitter/query.lua:373: in function 'fn'
	...ovim/0.11.4/share/nvim/runtime/lua/vim/func/_memoize.lua:78: in function 'fn'
	...ovim/0.11.4/share/nvim/runtime/lua/vim/func/_memoize.lua:78: in function 'get'
	...o/data/nvim/lazy/LazyVim/lua/lazyvim/util/treesitter.lua:23: in function 'have_query'
	...o/data/nvim/lazy/LazyVim/lua/lazyvim/util/treesitter.lua:40: in function 'have'
	...ata/nvim/lazy/LazyVim/lua/lazyvim/plugins/treesitter.lua:155: in function <...ata/nvim/lazy/LazyVim/lua/lazyvim/plugins/treesitter.lua:154>
	[C]: in function 'nvim_cmd'
	...rew/Cellar/neovim/0.11.4/share/nvim/runtime/filetype.lua:36: in function <...rew/Cellar/neovim/0.11.4/share/nvim/runtime/filetype.lua:35>
	[C]: in function 'pcall'
	vim/shared.lua: in function <vim/shared.lua:0>
	[C]: in function '_with'
	...rew/Cellar/neovim/0.11.4/share/nvim/runtime/filetype.lua:35: in function <...rew/Cellar/neovim/0.11.4/share/nvim/runtime/filetype.lua:10>
	[C]: in function 'bufload'
	...nel/.repro/data/nvim/lazy/conjure/lua/conjure/buffer.lua:27: in function 'upsert_buf'
[...]

PR #821, opened 2025-10-31
Update fennel set_form grammar structure

micampe:
The fennel grammar was updated to [remove binding_pair](https://github.com/alexmozaidze/tree-sitter-fennel/commit/49293d2) from set.

Without this trying to use the queries gives an "impossible pattern" error.


micampe:
I hadn't checked the opened issues, this fixes #811

kiyoon:
Thank you!

PR #830, opened 2025-11-26
feat(zsh): add various textobjects

chrisgrieser:
nvim-treesitter has recently added zsh support: https://github.com/nvim-treesitter/nvim-treesitter/pull/8240

Since zsh and bash are so similar, simply inheriting the bash textobject queries appears to work just fine. At least all cases I tried do work.

### DIFF
--- a/queries/zsh/textobjects.scm
+++ b/queries/zsh/textobjects.scm
@@ -1,1 +1,37 @@
-; inherits: bash
+(function_definition) @function.outer
+
+(function_definition
+  body: (compound_statement
+    .
+    "{"
+    _+ @function.inner
+    "}"))
+
+(case_statement) @conditional.outer
+
+(if_statement
+  (_) @conditional.inner) @conditional.outer
+
+(for_statement
+  (_) @loop.inner) @loop.outer
+
+(while_statement
+  (_) @loop.inner) @loop.outer
+
+(comment) @comment.outer
+
+(regex) @regex.inner
+
+((word) @number.inner
+  (#lua-match? @number.inner "^[0-9]+$"))
+
+(variable_assignment) @assignment.outer
+
+(variable_assignment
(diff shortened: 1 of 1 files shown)


chrisgrieser:
Not sure what the failing test is supposed to mean, since I did not change the README and I thought the large table was auto-generated?

kiyoon:
Thank you!

kiyoon:
@clason I no longer have the permission I can bypass the check and merge. Since the CI is broken, I no longer can merge anything.

clason:
@kiyoon sorry, the tests should not have been required as long as they are failing. (Fixing these should be the highest priority!)

I've removed them, and also gave you bypass permissions (which I forgot when setting up the branch rules for the new default branch).

kiyoon:
@clason Thank you. I remember you once changed the test CI and how is the progress? in master branch it used to have checking all queries compile without an error, and also some selection region checks

clason:
I have ported the full CI infrastructure from `master`, so the same tests run. ("Lint query files" is the check that verifies the queries are valid and compatible with the parsers pinned by nvim-treesitter.)

The "Tests" check is the selection etc. check, and what is currently broken. (I believe it was already broken on `master`, so this seems to be a regression related to testing with nightly/0.12.)

clason:
> Since zsh and bash are so similar, simply inheriting the bash textobject queries appears to work just fine. 

Please don't do that unless the grammars also inherit from each other (which the ones in the C family do). Each of these parsers can make a breaking change at any time, at which point the link will fail.

So if the same queries work, just copy them.

chrisgrieser:
> So if the same queries work, just copy them.

Sure thing, have done so.

clason (approved):

clason:
Note that the zsh parser is under active development, so please keep an eye out for breaking changes. I consider you the "zsh maintainer" from now on ;)

PR #831, opened 2025-11-26
feat(jsonc,git_config): add `@comment.outer` textobject

chrisgrieser:
adds `@comment.outer` for `jsonc` and `git_config`

edit: Not sure what the failing test is supposed to mean, since I did not change the README and I thought the large table was auto-generated?

### DIFF
--- a/queries/git_config/textobjects.scm
+++ b/queries/git_config/textobjects.scm
@@ -1,0 +1,1 @@
+(comment) @comment.outer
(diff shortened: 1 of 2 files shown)


kiyoon:
We're currently working on the test infrastructure because it has been broken. Thank you!

clason:
Note that the `jsonc` parser is unmaintained (stuck at ABI 13 for years) and hosted on Gitlab (which creates flakiness) so will likely be removed in the near future.

PR #833, opened 2025-12-01
fix(repeatable_move): using `;` or `,` in operator-pending mode misses one character. ci: fix broken test

### LINKED ISSUE #699: Change behavior when repeating last move from builtin_f_expr and  builtin_t_expr
When using `repeat_last_move`on operation pending mode, after triggering`builtin_f_expr`
or `builtin_t_expr`, it operates on one character less that it would be expected to.

Taking this snippet as an exemple:
```sh
vim.keymap.set({ 'n', 'x', 'o' }, ';', ts_repeat_move.repeat_last_move)
```
If the user presses `f(`, returns to the first character and presses `d;`, it would result on:
```sh
({ 'n', 'x', 'o' }, ';', ts_repeat_move.repeat_last_move)
```
`t(_d;` would result on:
```sh
t({ 'n', 'x', 'o' }, ';', ts_repeat_move.repeat_last_move)
```

For `f(_d;`, it would be expected to behave, as:
```sh
{ 'n', 'x', 'o' }, ';', ts_repeat_move.repeat_last_move)
```
and for `t(_d;`:
```sh
({ 'n', 'x', 'o' }, ';', ts_repeat_move.repeat_last_move)
```
**Additional context**
This diverges from the way it works in core nvim and is inconsistent with `F`, `T`,
and visual selection behavior.

As far as i can tell, this is only reproducible when repeating moves from `builtin_f_expr` and 
`builtin_t_expr` on operation pending mode, not with `builtin_F_expr` and `builtin_T_expr`.

kiyoon:
Fixes #699 properly.

Also fixes broken tests.

I personally found it very hard to test the CI, because it's hard to test locally due to the fact that all parsers are installed into `~/.local/share/nvim`. Thus, I think it's good to sandbox the data into `.test-deps` in `minimal_init.lua`.

Also I believe that lint-docs should be a separate CI that opens a PR of a new updated docs.


kiyoon:
@clason for some reason I cannot run the CI

clason:
> I personally found it very hard to test the CI, because it's hard to test locally due to the fact that all parsers are installed into ~/.local/share/nvim. Thus, I think it's good to sandbox the data into .test-deps in minimal_init.lua.

True, although that would force people to redownload parsers, which can take a lot of time and bandwidth. Feel free to add a `setup` call to `minimal_init.lua`.

> Also I believe that lint-docs should be a separate CI that opens a PR of a new updated docs.

Absolutely not. Commits should be atomic and not be (routinely) followed by what amounts to "fixup" commits. Also, such workflows are only run on the default branch, so that wouldn't work for us now.

kiyoon:
> True, although that would force people to redownload parsers, which can take a lot of time and bandwidth. Feel free to add a `setup` call to `minimal_init.lua`.

For now the test only run with python, and other checks still use the default directory. What do you mean `setup`, the `nvim-tressiter`'s setup? Is it necessary?

> Absolutely not. Commits should be atomic and not be (routinely) followed by what amounts to "fixup" commits. Also, such workflows are only run on the default branch, so that wouldn't work for us now.

Well, it would force all contributors to run the docs generating script, which will download bunch of parsers in their machine, making it way harder to contribute.  Also it wouldn't be possible without directory sandboxing to force people to do that, and I couldn't actually get it working.

I get a why a commit should be atomic in a way, but it's such a minor thing that doesn't matter that much, imo.

clason:
> For now the test only run with python, and other checks still use the default directory. What do you mean setup, the nvim-tressiter's setup? Is it necessary?

Yes, if you don't want to install the parsers in the default directory.

> Well, it would force all contributors to run the docs generating script, which will download bunch of parsers in their machine, making it way harder to contribute. 

No, that should not happen. The docs script doesn't need the parsers, only the queries (which are always available), same as for nvim-treesitter. If this is currently not the case, that's a bug in the script that needs to be fixed.

**EDIT** Yeah, I see the issue now. That will require some thought. But note that CI spits out the expected change, so you can always just copy&paste manually.

(On a related note, that table should be moved to a separate document; no need to clutter the main README.)

> it's such a minor thing that doesn't matter that much, imo.

Believe me, such bot PRs are more than a nuisance ;) It made working with nvim-treesitter (master) incredibly annoying.

kiyoon:
> Yes, if you don't want to install the parsers in the default directory.

I sandboxed it by setting XDG variables so the default directory is actually good. On top of the parsers, it will just install everything in the directory. 

> **EDIT** Yeah, I see the issue now. That will require some thought. But note that CI spits out the expected change, so you can always just copy&paste manually.

Then let's put it verbosely in the Github Summary of the CI and instruct users to copy ir. 

> (On a related note, that table should be moved to a separate document; no need to clutter the main README.)

Noted. I will change this too

> Believe me, such bot PRs are more than a nuisance ;) It made working with nvim-treesitter (master) incredibly annoying.

Okay. But I will separate the test CI completely because it doesn't make sense that the CI stops just because of readme mismatching. Clearly the test is more important and it doesn't even require all parsers.

clason:
Well, the jobs are combined because they both require parsers installed. 

Let's keep them together but make sure both jobs run even if one fails (it's a trivial change).

clason:
> I sandboxed it by setting XDG variables so the default directory is actually good. On top of the parsers, it will just install everything in the directory.

That works; I was considering NVIM_APPNAME instead but your approach is a better fit here. We should set this in the makefile. Then it applies to the docs workflow as well.

kiyoon:
> Well, the jobs are combined because they both require parsers installed.
> 
> Let's keep them together but make sure both jobs run even if one fails (it's a trivial change).

No, in fact the test only requires the python parser and not all. I think separating it makes it more clear what the dependencies are for the tests in case users want to test on their local machine

clason:
But it might not stay that way; at some point it would be nice to have regression tests for the actual queries.

kiyoon:
1. I made BUILTIN_TEXTOBJECTS.md to separate it from the README.md
2. Now the CI instructs how to patch the doc.
3. The CI doesn't stop even if previous checks fail.

I think this is now ready.

PR #834, opened 2025-12-04
feat(ocaml): include "in" keyword in function.outer

beajeanm:
Make sure the fonction.outer includes the in keyword `in`, e.g.:

let f x = let g y = 1 in f (g 2)
          ^_____________^

### DIFF
--- a/queries/ocaml/textobjects.scm
+++ b/queries/ocaml/textobjects.scm
@@ -1,6 +1,8 @@
-(value_definition
+((value_definition
   (let_binding
-    body: (_) @function.inner)) @function.outer
+    body: (_) @function.inner)) @_fun_start @_fun_end
+  ("in")? @_fun_end
+  (#make-range! "function.outer" @_fun_start @_fun_end))
 
 (method_definition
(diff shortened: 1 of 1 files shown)


kiyoon:
make-range is removed in `main`

beajeanm:
@kiyoon Where can I read about the removal and the alternative to make-range? Thanks.

kiyoon:
The reason `#make-range` was necessary was due to the lack of support for quantified captures in previous neovim (#612). I'm not experienced in ocaml thus I don't understand what you're trying to achieve, but most likely the important parts should be supported by the language parser.

beajeanm:
Hmmm, ok, I'll see if I can figure it out.

> I'm not experienced in ocaml thus I don't understand what you're trying to achieve
 
OCaml has top level and nested declaration, the nested ones are terminated by the `in` keyword. I was trying to include `in` in the function.outer group.
As you can see in the screenshot below "in" is not part of the value_definition node, so I was using make-range to work around that and capture both the cases with "in" and the case without "in".

Screenshot 2025-12-11 at 11 36 35

PR #835, opened 2025-12-06
feat(json)!: move jsonc comment textobject to json

clason:
The `json` parser now supports comments, so a separate `jsonc` parser is
no longer necessary and will be removed from nvim-treesitter (which will
instead set up a `jsonc->json` alias).

(I'm actually suprised there were no `json` textobjects before; something like
`@attribute` for properties sounds useful. But that's a follow-up PR from
someone else.)


kiyoon:
Looks simple enough :)

PR #840, opened 2025-12-19
fix: range containment check (e.g. weird @parameter.inner behavior)

Dook97:
Fixes #832

PR #841, opened 2025-12-20
fixup! fix: range containment check (e.g. weird @parameter.inner beha…

Dook97:
I "fixed" an issue in #840 only to realize, I left it broken in another way. Sorry for that. Anyway, here's *the actual* fix (I hope).

I frankly still don't understand, why `end_col = end_col - 1` is needed in the first place, as it breaks the expected behaviour of `is_in_range()`. I assume it's a quirk specific to the location where the function was being called from, so I moved it there.

### DIFF
--- a/tests/select/lua/selection_mode.lua
+++ b/tests/select/lua/selection_mode.lua
@@ -1,0 +1,7 @@
+function a()
+  print('foo')
+end
+
+function b()
+  print('bar')
+end
--- a/tests/select/lua_spec.lua
+++ b/tests/select/lua_spec.lua
@@ -1,0 +1,12 @@
+local Runner = require('tests.select.common').Runner
+
+local run = Runner:new(it, 'tests/select/lua', {
+  tabstop = 2,
+  shiftwidth = 2,
+  softtabstop = 0,
+  expandtab = true,
+})
+
+describe('command equality Lua:', function()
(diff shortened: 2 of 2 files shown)


kiyoon:
Because treesitter range is end-exclusive. For example there are cases like `(last_row+1, 0)` is the end position to mean it's till the end of the file. If you don't calibrate it well, it will throw out of range error.

I think the safest fix is to copy the functions from neovim upstream for now.

contains()

https://github.com/neovim/neovim/blob/81e2d2ea53479da1b7b147ab17a72d25babf6b8e/runtime/lua/vim/treesitter/_range.lua#L142-L160

is_in_range()

https://github.com/neovim/neovim/blob/81e2d2ea53479da1b7b147ab17a72d25babf6b8e/runtime/lua/vim/treesitter.lua#L250-L259

Dook97:
Is it okay to use private modules like this? If so the following passes the tests.

Dook97:
Added a new test which covers the case which my previous PR broke.

kiyoon:
I'm sorry but this is being replaced with #845

kiyoon:
but your test is useful. I appreciate your work still!

Dook97:
I'm just happy it's getting fixed :) Merry Christmas!

PR #845, opened 2025-12-24
fix: range contains(), is_in_range() to use nvim equivalents

kiyoon:
Changes most functions to accept generic `Range` type instead of only accepting a specific `Range4`, `Range6` type etc. and converting the type at the call site.

Mainly changed `shared.torange4()` -> `_range.unpack4()` to achieve this.

### DIFF
--- a/tests/select/lua_spec.lua
+++ b/tests/select/lua_spec.lua
@@ -12,5 +12,5 @@ describe('Look back if within @function.outer range (Lua):', function()
 end)
 
 describe('Look forward if outside @function.outer range (Lua):', function()
-  run:compare_cmds('lookback.lua', { row = 4, col = 0, cmds = { 'dim', '2j^D' } })
+  run:compare_cmds('lookback.lua', { row = 5, col = 0, cmds = { 'dim', '2j^D' } })
 end)


clason:
This is the correct approach, IMO. (My plan is to be aggressive when dropping support in nvim-treesitter#main, so hopefully this won't be long.)

clason:
But it's not really clear what the upstream plan is here -- whether `vim.Range` will (be able to) completely replace `vim.treesitter._range`, or whether there will be some under-the-hood refactoring of the latter on top of the former (in which case we'd also have to bundle `vim.Range`).

kiyoon:
> But it's not really clear what the upstream plan is here -- whether `vim.Range` will (be able to) completely replace `vim.treesitter._range`, or whether there will be some under-the-hood refactoring of the latter on top of the former (in which case we'd also have to bundle `vim.Range`).

I think as long as we stay updated with the upstream and this code just follows similarly, it should be maintainable and not introduce serious bugs.

clason:
Yes, just mentioning possible difficulties (and making clear that I have no privileged information) here.

Issue #847, opened 2025-12-30
After switch to main selection only works for few languages


mikehaertl:
As `main` is now default for nvim-treesitter I've tried to migrate my config to `main` also for textobjects.

Unfortunately text selection now only works for few languages for me. I've tested with `select_textobject("@function.outer", "textobjects")`:

 * Selection works: lua, javascript
 * Does nothing: java, php, typescript

Treesitter itself seems to work fine for all languages: I can open the tree with `:InspectTree` and navigate it.

### What I've tried

I've tried to debug but I'm not an expert on the treesitter API. I've followed the code to this point:

https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/ecd03f5811eb5c66d2fa420b79121b866feecd82/lua/nvim-treesitter-textobjects/shared.lua#L152-L157

To debug I changed it like so:
```lua
  local ranges = {} ---@type Range6[]
  print("point1")
  parser:for_each_tree(function(tree, lang_tree)
    print("point2")
```
For the failing languages it only prints `point1` but not `point2`. 

### Configuration

```lua
require'nvim-treesitter'.install {
  'angular',
  'bash',
  'c',
  'cmake',
  'comment',
  'cpp',
  'css',
  'csv',
  'dap_repl',
  'diff',
  'dockerfile',
  'fennel',
  'go',
  'html',
  'http',
  'java',
  'javascript',
  'jsdoc',
  'json',
  'json5',
  'jsonc',
  'kotlin',
  'lua',
  'make',
  'markdown',
  'markdown_inline',
  'perl',
  'php',
  'phpdoc',
  'properties',
  'python',
  'query', -- treesitter query language
  'regex',
  'ruby',
  'rust',
  'scheme',
  'sql',
  'toml',
  'typescript',
  'vim',
  'vimdoc',
  'vue',
  'yaml',
}

require("nvim-treesitter-textobjects").setup {
  select = {
    lookahead = true,
    selection_modes = {
      ["@function.outer"] = "V"
    },
  },
}

-- Using whichkey to configure mappings
local whichkey = require('which-key')
local textobjects = require('nvim-treesitter-textobjects.select')
whichkey.add({
  {
    mode = { "o", "x" },
    { "ac", function() textobjects.select_textobject("@comment.outer", "textobjects") end, desc = "a comment" },
    { "af", function() textobjects.select_textobject("@function.outer", "textobjects") end, desc = "a function" },
[...]

PR #848, opened 2026-01-02
docs: mention `vim.treesitter.start()` is needed

kiyoon:
Avoids confusion such as #847

And I think it's pretty important that it mentions its requirements clearly.


clason:
But it is not! That's the whole point of the rewrite. If something doesn't work without, that's a bug. (Probably related to a missing `parse()` call somewhere.)

clason:
Sorry, this _really_ needs to work without calling `start()`.

PR #853, opened 2026-01-06
feat(ocaml): extend textobjects

neojski:
Ocaml function textobjects updates.

I actually tested those when I wrote the change in https://github.com/meain/evil-textobj-tree-sitter/pull/134 but I was kindly asked to upstream here.

### DIFF
--- a/queries/ocaml/textobjects.scm
+++ b/queries/ocaml/textobjects.scm
@@ -25,14 +25,15 @@
 (method_definition
   body: (_) @function.inner) @function.outer
 
-; let x = expr (also matches let f x = expr due to grammar limits)
+; let pattern = body (also matches let f x = expr due to grammar limits)
+; Since we want @assignment.inner to match both pattern and body we have to split it into two:
 (value_definition
   (let_binding
     pattern: (_) @assignment.lhs @assignment.inner)) @assignment.outer
 
 (value_definition
   (let_binding
-    body: (_) @assignment.rhs @assignment.inner)) @assignment.outer
+    body: (_) @assignment.rhs @assignment.inner))
 
 ; module M = struct ... end
 (module_definition


kiyoon:
Thanks for contributing.

Can you describe what it fixes and what the previous function capture was doing?

neojski:
Added some inline comments to make it easier for future devs to understand it as well.

neojski:
I would also appreciate some guidance on how to test it directly in this repo. README explains how to install it but it looks like it mostly focuses on the official package and not the locally developed one.

neojski:
I managed to test it locally:
1. I needed to install `nvim-treesitter-parser` from AUR
2. ran `nvim --clean -u scripts/minimal_init.lua test.ml`

neojski:
So, I think this doesn't actually cover some cases, e.g.
```
List.map ~f:(fun x -> x * x)
```
It won't match `fun x -> x * x` for inner function. I'll add those soon

kiyoon:
A good way to design a query is

1. open `:InspectTree`
2. press `o` to open query editor/viewer

image

neojski:
That's really useful, would you mind if I add that to https://github.com/nvim-treesitter/nvim-treesitter-textobjects?tab=contributing-ov-file?

I pushed the changes to handle the "standalone" fun & function now, so I think it's ready.

neojski:
I also added struct & module support, which worked great locally, but those seem to be rejected :cry: 

I could go with using `class` for `module`, perhaps, and `block` for `struct` but that would be a breaking change for `block`. But I don't think `parenthesized_expression` is useful given that `vi(` just works.

neojski:
I just did.

kiyoon:
> That's really useful, would you mind if I add that to https://github.com/nvim-treesitter/nvim-treesitter-textobjects?tab=contributing-ov-file?

~Sure, please change the CONTRIBUTING.md file here.~

I opened #856

kiyoon:
assignment.inner should also select lhs as a separate query

neojski:
All done, I think?

neojski:
Updated. Now inner should work on both sides of the assignment.

kiyoon:
Thanks, the inner problem is fixed, but the assignment.outer is repeated two times. Basically the two queries capture exactly the same region which is redundant and will affect performance. Please remove in one of the two places.

neojski:
Fixed. Having worked through all the feedback I wish there was some test suite that I could encode this feedback in. I added a number of comments to code inline to help future contributors to ocaml, at least.

kiyoon:
I'm sorry for the inconvenience, and thanks a lot for the patience and contribution!

PR #855, opened 2026-01-07
feat(gleam): add textobjects

locnide:
This pull requests adds textobjects for gleam.

I mainly took inspiration from the rust `textobject.scm`. 

Feel free to review, it is the first treesitter query file I wrote.

PR #715 is similar but seem abandonned.

---

Some thing that I still have to implement before merge: 

- [x] trailing comma for parameter.outer. Currently trailing comma are ignored.
- [x] add `@assign.inner` but I’m not sure what it should select, see https://github.com/nvim-treesitter/nvim-treesitter-textobjects/discussions/852
- [x] migrate my install to treesitter-nvim main… current CI issues mighnt be due to an obsolete parser I used for testing
- [x] rebase and squash

---

Some tags are not implemented in this PR:

- `@attribute` couldn’t find examples in languages I know. Are there supposed to be type/class attributes ?
- `@comment.inner` I would have to use directives
- `@frame`, `@loop` and `@regex` not relevant
- `@scopename` I do not know what it is
---

~~P.S. : I had trouble making `make docs` work https://github.com/nvim-treesitter/nvim-treesitter-textobjects/discussions/854~~


locnide:
The PR should be ready but there is an error in CI I don’t understand and can’t reproduce.

The treesitter `./scripts/install-parsers.lua` fails with 502 for gleam. I don’t understand why 502 (and not 40x) and only for gleam.
(and I can’t restart the job to check if it wasn’t a github issue)

clason:
That's a Github brownout.

PR #860, opened 2026-01-23
feat(git_config): add @assignment, @block and @statement text objects

locnide:
It adds the following textobjects for `git_config`:

- `@assignment.lhs`, `@assignment.inner`, and `@assignment.rhs`
- `@block.inner` and `blok.outer`
- `@statement.outer`


kiyoon:
assignment.inner and outer selects the same range. You need to keep two separate queries for assignment.inner otherwise it will act as a range.

locnide:
> assignment.inner and outer selects the same range. You need to keep two separate queries for assignment.inner otherwise it will act as a range.

Oh, sorry. it should have been fixed.

---

Is it because there is only an anonymous node separating the two `@assignment.inner` captures, or will a single query always produce a single capture for a specific name ?

kiyoon:
single query for single range capture for each name. However I'm not entirely sure if it's treesitter doing it or just this plugin.

PR #865, opened 2026-02-20
fix(supercollider): update queries after breaking change

elgiano:
Updated queries for supercollider after breaking parser update


elgiano:
pinging @clason because of nvim-treesitter PR https://github.com/nvim-treesitter/nvim-treesitter/pull/8521

PR #867, opened 2026-03-08
feat(typst): add textobjects

antinomie8:
This PR adds textobjects for typst.

All textobjects have been added, besides `@attribute.*`, `@class.*`, `@frame.*` and `@comment.inner`.

There is a tiny issue with `@function.inner`: typst lambdas can either contain a block as their value or not, e.g.
```typst
it => it
```
and
```typst
it => {
    // do stuff ...
}
```
are both valid, but `@function.inner` shouldn't match the braces in the second case (at least according to what is done for other languages). However, currently I match both with and without braces in the second case (instead of only without).

It's my first time writing nontrivial tree-sitter queries, I haven't tested everything thoroughly, and I probably don't know all the syntax features of typst, so there might be some errors or omissions.

### DIFF
--- a/queries/typst/textobjects.scm
+++ b/queries/typst/textobjects.scm
@@ -7,11 +7,8 @@
     "}")) @function.outer
 
 (lambda
-  value: (_
-    .
-    "{"?
-    _+ @function.inner
-    "}"?)) @function.outer
+  value: (_) @function.inner
+  (#match? @function.inner "^[^{].*[^}]$")) @function.outer
 
 ; named functions (#let fn(x) = { ... })
 (let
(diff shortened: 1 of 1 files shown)


antinomie8:
> There is a tiny issue with `@function.inner`: typst lambdas can either contain a block as their value or not, e.g.
> 
> ```typst
> it => it
> ```
> 
> and
> 
> ```typst
> it => {
>     // do stuff ...
> }
> ```
> 
> are both valid, but `@function.inner` shouldn't match the braces in the second case (at least according to what is done for other languages). However, currently I match both with and without braces in the second case (instead of only without).

Now fixed using `#match` predicate.

PR #874, opened 2026-03-20
test: replace plenary with minimal fork

clason:
Problem: plenary.nvim as test runner is overkill and no longer
maintained.

Solution: Replace with a minimal fork based on Neovim API and vendored
luassert.

Follow-up to https://github.com/nvim-treesitter/nvim-treesitter/pull/8592

This finally fixes user config polluting the system-under-testing by running the instances with `--clean` (which requires manual handling of `rtp` in the minimal init to ensure queries and parsers are available; follow-up should add a `nvim-treesitter.setup` call for explicit custom parser and query dir).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,13 +81,13 @@ nvim_ts: $(NVIM_TS)
 $(NVIM_TS):
 	git clone --filter=blob:none --single-branch https://github.com/nvim-treesitter/nvim-treesitter -b main $(NVIM_TS)
 
-PLENARY := $(DEPDIR)/plenary.nvim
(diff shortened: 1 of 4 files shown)


clason:
@kiyoon instead of messing with XDG environment variables, wouldn't it be easier to set a custom `$NVIM_APPNAME` (as that is precisely what it's for)?

kiyoon:
that still makes your test directory saved multiple places, and it's harder to clean up.

I followed the approach here which is battle-tested and worked for me for a long time.

https://github.com/folke/lazy.nvim/blob/main/bootstrap.lua

Issue #877, opened 2026-04-06
`select_textobject("@parameter.inner")` doesn't catch Rust's `ordered_field_declaration_list` syntax tree node


adaszko:
**Describe the bug**
`@parameter.inner` and `@parameter.outer` do not capture some `ordered_field_declaration_list` elements in Rust

**To Reproduce**
Place the cursor on `usize` on a Rust program below and call `require "nvim-treesitter-textobjects.select".select_textobject("@parameter.inner", "textobjects")`:

```Rust
pub enum Foo {
    Bar(usize, u32),
}

fn main() {
    println!("Hello, world!");
}
```
The cursor jumps to and selects `"Hello, world!"`.

**Expected behavior**

The entire `usize` word should get selected.

**Output of `:checkhealth nvim-treesitter`**

<details>
<code>

nvim-treesitter:                                                            ✅

Requirements ~
- ✅ OK Neovim was compiled with tree-sitter runtime ABI version 15 (required >=13).
- ✅ OK tree-sitter-cli 0.26.7 (/opt/homebrew/bin/tree-sitter)
- ✅ OK tar 3.5.3 (/usr/bin/tar)
- ✅ OK curl 8.7.1 (/usr/bin/curl)
  curl 8.7.1 (x86_64-apple-darwin25.0) libcurl/8.7.1 (SecureTransport) LibreSSL/3.3.6 zlib/1.2.12 nghttp2/1.68.0
  Release-Date: 2024-03-27
  Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns ldap ldaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
  Features: alt-svc AsynchDNS GSS-API HSTS HTTP2 HTTPS-proxy IPv6 Kerberos Largefile libz MultiSSL NTLM SPNEGO SSL threadsafe UnixSockets

OS Info ~
- release: 25.3.0
- version: Darwin Kernel Version 25.3.0: Wed Jan 28 20:54:46 PST 2026; root:xnu-12377.91.3~2/RELEASE_ARM64_T6000
- sysname: Darwin
- machine: arm64

Install directory for parsers and queries ~
- /Users/adaszko/.local/share/nvim/site
- ✅ OK is writable.
- ✅ OK is in runtimepath.

Installed languages     H L F I J ~
- ecma
- go                    ✓ ✓ ✓ ✓ ✓
[...]

PR #878, opened 2026-04-06
feat(rust): @parameter.{inner,outer} for ordered_field_declaration_list

adaszko:
Fix for https://github.com/nvim-treesitter/nvim-treesitter-textobjects/issues/877

### DIFF
--- a/queries/rust/textobjects.scm
+++ b/queries/rust/textobjects.scm
@@ -261,6 +261,26 @@
   .
   "," @parameter.outer .)
 
+(enum_variant
+  body: (ordered_field_declaration_list
+    "," @parameter.outer
+    .
+    (_) @parameter.inner @parameter.outer))
+
+(enum_variant
+  body: (ordered_field_declaration_list
+    .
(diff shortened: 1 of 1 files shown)


kiyoon:
can you follow the other implementations like it has to have 3 variations per situation, the ordering of each and the last one should cover the last trailing comma.

adaszko:
Good catch!  Amended

queries/rust/textobjects.scm
```diff
@@ -261,6 +261,25 @@
...
+    .
+    (_) @parameter.inner @parameter.outer))
+
+(enum_variant
+  body: (ordered_field_declaration_list
+    .
+    (_) @parameter.inner @parameter.outer
+    .
+    ","? @parameter.outer))
+
```
kiyoon, line 276: can you also add this comment here?

; last element, with trailing comma

adaszko: Done

kiyoon:
Thank you!

adaszko:
My pleasure!

PR #885, opened 2026-05-08
feat(move,swap): wrapping around sibling lists

mazunki:
I was a bit surprised that `@parameter.inner` would go to the next procedure call, and not stop at the end of the parameter list when swapping order.

I've added support for `opts = { wrap = true }` which limits the scope in which `move.goto_{next,previous}{,_end,_start}` and `swap.swap_{next,previous}` operate.

By default, this feature is disabled, and the API remains unchanged: so the feature is entirely opt-in. My opinion is that at the very least, the swapping commands should wrap around its siblings, but I don't intend on breaking expectations with this PR.

I haven't tested this on other things than parameter lists, but I suspect it would work with them just the same.


mazunki:
rebased to address linter issues

PR #887, opened 2026-05-12
add support for xml language

mazunki:
I was undecided between `function` and `class` for tags here. Ideally they should just be called `@tag` I feel like, but alas, this is what HTML uses, which is why I went with that.

Related to https://github.com/nvim-mini/mini.nvim/issues/2417


kiyoon:
- `attribute.inner` includes the quotes. cit on `id="1"` leaves `id=` and you retype the quotes. This repo's HTML query excludes them and produces no capture for an empty value, jsx does the same.
- `comment.inner` equals `comment.outer`. Most languages here trim the delimiters, HTML defines no inner at all. Use #offset! 0 4 0 -3, with a guard for the empty `<!---->` comment, where the zero-width range misbehaves
- The `_+ @function.inner` pattern and the `(EmptyElemTag) @function.outer` pattern are redundant. tree-sitter-xml wraps all children in one content node, so both duplicate the pattern above them.

PR #892, opened 2026-07-03
Permit swap_previous, swap_next to take a list of query strings

fionn:
The `swap_next` function takes a `query_strings` argument. Despite this being plural, the type annotation claims that this should be a `string`. However, we pass this argument to `swap_textobject` which expects type `string | string[]`.

This also makes the `swap` functions better correspond to those of `move`, e.g. `goto_next_start` also accepts `query_strings` of type `string[]`.

The same holds also for `swap_previous`.


fionn:
Thanks!

PR #893, opened 2026-07-15
fix(swap): annotate query_strings as string|string[]

samov0619-prog:
M.swap_next and M.swap_previous are annotated ---@param query_strings string, but they pass the argument straight through to swap_textobject, which is annotated string|string[] and unwraps tables natively (if type(query_strings) == 'string' then query_strings = { query_strings } end). A list is the native path — a plain string is the one that gets wrapped.
The parameter name is already plural. move.lua annotates its goto_* functions correctly as string|string[], so only these two are affected.
Passing a list works at runtime but makes lua_ls emit Cannot assign 'table' to parameter 'string' [param-type-mismatch], forcing users to either add ---@diagnostic disable-next-line or cast with --[[@as string]], which is a lie about the actual type.


clason:
duplicate of #892

